PR #21, opened 2022-03-16
[Fix] Project config

kilemensi:
## Description

This PR fixes project setup:

  - [x] Ensures root doesn't contain unnessary config; each package or app configures everything it needs,
  - [x] Introduces eslint-config-commons-ui as a standalone package to handle eslint configs for normal React package as well as Next.js app,
  - [x] Delegates lint-stage config from root to individual package or app,

## Type of change

## Screenshots

N/A

## Checklist:


.lintstagedrc.json
```diff
@@ -0,0 +1,3 @@
+{
+    "*.{json,md,yml}": "prettier --write"
+}
```
kilemensi, line 3:
```suggestion
}

```

nerdlet (approved):
Looks good @kilemensi  🚀

apps/web/package.json
```diff
@@ -6,17 +6,26 @@
     "dev": "next dev",
     "build": "next build",
     "start": "next start",
-    "lint": "next lint"
+    "lint": "eslint --fix './'",
+    "clean": "rm -rf .turbo && rm -rf node_modules"
```
gertie-sheshe, line 10: When do we need to run this? @kilemensi

apps/web/package.json
```diff
@@ -6,17 +6,26 @@
...
   },
   "dependencies": {
-    "next": "^12.1.0",
+    "@babel/core": "^7.17.7",
+    "@emotion/react": "^11.8.2",
+    "@emotion/styled": "^11.8.1",
+    "@mui/material": "^5.5.1",
+    "next": "12.1.0",
+    "next-transpile-modules": "^9.0.0",
     "react": "^17.0.2",
```
gertie-sheshe, line 19: Why are these here? Aren't they meant to be peer dependencies of mui? How does that work?

apps/web/package.json
```diff
@@ -6,17 +6,26 @@
...
     "react-dom": "^17.0.2",
     "ui": "workspace:*"
   },
   "devDependencies": {
-    "config": "workspace:*",
-    "eslint": "^7.32.0",
-    "next-transpile-modules": "^9.0.0"
+    "@types/node": "^17.0.21",
+    "@types/react": "^17.0.40",
+    "eslint": "^8.11.0",
```
gertie-sheshe, line 26: Why is `eslint` a dependency in this project? Isn't there an eslint config that already has eslint? Won't they clash?

packages/ui/Button.js
```diff
@@ -1,5 +1,8 @@
+import MuiButton from "@mui/material/Button";
```
gertie-sheshe, line 1: This cancels out my work 👀 A heads up would have been appreciated lol @kilemensi

apps/web/package.json
kilemensi: That's part of the review @gertie-sheshe. When do you **_think_** we'd need this (if at all)?

apps/web/package.json
kilemensi: They **_are_** peer dependencies in `MUI` and `ui` and hence why they're here. Exactly, how does it [work](https://github.com/npm/cli/blob/latest/docs/content/configuring-npm/package-json.md#devdependencies)?

apps/web/package.json
kilemensi: They won't clash. Yes, **_why_**?

packages/ui/Button.js
kilemensi: ... This is why I haven't merged this PR @gertie-sheshe. Waiting for you to merge yours.

apps/web/package.json
gertie-sheshe: Part of the review is to ask questions where we do not understand, not to explain 😭😭😭 Wawu @kilemensi

kilemensi: Welcome to the **new school** @gertie-sheshe .... basically whenever you need to reset the repo e.g. you're having issues with turbo cache, call `pnpm clean`, having issues with dependencies and you'd like to run `pnpn install` from scratch, `call pnpm clean`, etc., etc.

apps/web/package.json
kilemensi: Because it's in the [docs](https://github.com/npm/cli/blob/latest/docs/content/configuring-npm/package-json.md#devdependencies) @gertie-sheshe and I'm a believer in the old old adage, _“If you give a woman a fish, you feed her for a day. If you teach a woman how to read the docs ..."_. In short:

**dependencies**: Used in the package/app and bundled with the package/app when distributed or deployed.
**devDependencies**: Used in the package/app but are not bundled with the package/app when distributed or deployed (Well, technically you can make a dev deployment but normally you don't). `eslint-config-commons-ui` has `eslint` under `devDependencies` so packages/apps using `eslint-config-commons-ui` will not see/have access to the `eslint `in `eslint-config-commons-ui`.
**peerDependencies**: This is rarely used but it is useful in packages more than apps. When a package declares `peerDependencies`, it is basically saying I (the package) needs this dependency to run but I'm not bundling it with me. Whichever package/app wants to use me (the package), they'll have to include the dependency in question themselves. There are a number of reasons why a package author will want to do this but the most obvious ones are usually bundle size (imagine you have 10 packages using `eslint` and each one has bundled its own `eslint`) and/or some dependencies don't work well when there more than one copy of them at runtime (**React**, **MUI**, etc.). Again, `eslint-config-commons-ui` has `eslint` under `peerDependencies` meaning every package/app using `eslint-config-commons-ui` has to declare its own `eslint`. This is why we have `eslint` in `app/web/package.json`.

packages/ui/Button.js
kilemensi: PR merged

PR #31, opened 2022-03-31
[Chore] Add Theme

gertie-sheshe:
## Description

Add a custom `createTheme` function that will include our own needed options to the [@mui/material/styles/createTheme](https://mui.com/customization/theming/#createtheme-options-args-theme). For now the option we need is `contentWidths` (previously widths) to be used as page content width for respective` breakpoints.`

Fixes # (issue)

## Type of change

## Checklist:


kilemensi:
Looks like you're on the right track to me @gertie-sheshe ... You're just not using contentWidths anywhere that's why it doesn't get picked up in storybook. For example, by modifying RichTypography storybook as 👇🏽 , everything appeared to be working:

```jsx
const Section = styled("div")(({ theme }) => ({
  display: "flex",
  [theme.breakpoints.up("sm")]: {
    justifyContent: "center",
    maxWidth: theme.contentWidths.values.sm,
  },
  [theme.breakpoints.up("md")]: {
    maxWidth: theme.contentWidths.values.md,
  },
}));

function Template({ content, ...args }) {
  return (
    <Section>
      <RichTypography {...args}>{content}</RichTypography>
    </Section>
  );
}
```

packages/commons-ui-core/styles/createTheme/createTheme.js
```diff
@@ -0,0 +1,30 @@
...
+          sm: 360, // mobile
+          md: 720, // tablet
+          lg: 1280, // desktop
+        },
+      },
+      contentWidths: {
+        values: {
+          md: 608, // 0, 80, 0, 80 margin
+          lg: 1160, // 0, 140, 0, 140 margin
+        },
```
kilemensi, line 24: We need to define these for **all** breakpoints.

kilemensi (commented):
👍🏽

packages/commons-ui-core/.storybook/preview.js
```diff
@@ -0,0 +1,18 @@
+import createTheme from "../styles/createTheme";
+import { ThemeProvider } from "@mui/material/styles";
+import { ThemeProvider as Emotion10ThemeProvider } from "emotion-theming";
+
+const theme = createTheme();
+import "./styles.css";
+
+const withThemeProvider = (Story, context) => {
+  return (
+    <Emotion10ThemeProvider theme={theme}>
```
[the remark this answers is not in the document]
gertie-sheshe: 1. Added the change to `main.js` The `Emotion10ThemeProvider` is just a naming, it references `emotion/core` from [here](https://mui.com/guides/migration-v4/#storybook-emotion-with-v5)
2. So far, I do not think we do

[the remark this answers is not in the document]
kilemensi: May be my question wasn't so clear re `Emotion10ThemeProvider` @gertie-sheshe: I wasn't asking about the name or where it comes from but if we still need to jump through these hoops any more. For example, the [issue](https://github.com/storybookjs/storybook/issues/16099) MUI was talking about [here](https://mui.com/guides/migration-v4/#storybook-emotion-with-v5), has already been fixed in Storybook [here](https://github.com/storybookjs/storybook/issues/16099#issuecomment-964445878). And we're already using `v6.4.x` of Storybook. So have you tested this without the `Emotion10ThemeProvider`, `config.resolve.alias` in Webpack, etc., and it still couldn't pick up our theme?

packages/commons-ui-core/src/RichTypography/index.stories.js
```diff
@@ -14,7 +16,11 @@ export default {
 };
 
 function Template({ content, ...args }) {
-  return <RichTypography {...args}>{content}</RichTypography>;
+  return (
+    <Section>
+      <RichTypography {...args}>{content}</RichTypography>
+    </Section>
+  );
```
KhadijaMahanga, line 23: @gertie-sheshe does it mean all stories in commons would need to be wrapped with `Section` cc @kilemensi

packages/commons-ui-core/styles/createTheme/createTheme.js
```diff
@@ -0,0 +1,25 @@
...
+      contentWidths: {
+        values: {
+          xs: 0,
+          sm: 480,
+          md: 608,
+          lg: 1160,
+          xl: 1440,
+        },
+      },
+    })
```
KhadijaMahanga, line 21: incase `contentWidths` is defined in options, then we should have `{ contentWidths: {} }` followed by `options` in the deepmerge.

packages/commons-ui-core/src/RichTypography/index.stories.js
gertie-sheshe: No @KhadijaMahanga this is a temporary implementation to test that themes works. I will add a comment

packages/commons-ui-core/.storybook/preview.js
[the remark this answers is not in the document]
gertie-sheshe: Oooh yeah I tested that this morning after reading your comment and Storybook crashed @kilemensi let me look into it further

packages/commons-ui-core/styles/createTheme/createTheme.js
gertie-sheshe: Not sure I understood this - the `contentWidths` option is the default given if one is not passed...

packages/config/storybook/main.js
```diff
@@ -1,4 +1,5 @@
 const path = require("path");
+const toPath = (filePath) => path.join(process.cwd(), filePath);
```
kilemensi, line 2: Do we still need these?

kilemensi (approved):
🚀 

---

Still not sure how but we'd somehow need to check that someone hasn't actually [customized](https://mui.com/customization/breakpoints/#custom-breakpoints) `breakpoints` before we add `contentWidths` since `breakpoitns` and `contentWidths` **must** use the exact same keys.

packages/commons-ui-core/styles/createTheme/createTheme.js
KhadijaMahanga: @gertie-sheshe I see the code above it (`if statement`) has changed, so it is now alright. Initially, the if statement above was checking for both `breakpoints && contentWidths` ...meaning if just contentWidths was defined on options without breakpoints it would be overridden by the default. That's why I had suggested the default contentWidths appear as the first parameter of `deepmerge`

KhadijaMahanga (approved):

PR #32, opened 2022-04-01
NavBar

### LINKED ISSUE #28: @commons-ui/core/NavBar
Using [`AppBar` and `Toolbar`](https://mui.com/components/app-bar/), create a `NavBar` component that we'll use as base of our navigation. The component should have the following props:

- position: sticky,
- color: primary,
- backgroundColor: palette.background.default

kilemensi:
## Description

Fixes #28 

## Type of change

## Checklist:

### DIFF
--- a/packages/commons-ui-core/.storybook/main.js
+++ b/packages/commons-ui-core/.storybook/main.js
@@ -1,1 +1,17 @@
-module.exports = require("config/storybook/main");
+const path = require("path");
+const defaultConfig = require("config/storybook/main");
+
+const { webpackFinal } = defaultConfig;
+
+module.exports = {
+  ...defaultConfig,
+  webpackFinal: async (config) => {
+    config = await webpackFinal(config);
+    config.resolve.alias = {
+      ...config.resolve.alias,
+      "@/commons-ui/core": path.resolve(__dirname, "../src"),
+    };
+
+    return config;
+  },
+};
--- a/packages/commons-ui-core/jest.config.js
+++ b/packages/commons-ui-core/jest.config.js
@@ -1,1 +1,12 @@
-module.exports = require("jest-config-commons-ui");
+const defaultConfig = require("jest-config-commons-ui");
+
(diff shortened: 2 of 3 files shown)


packages/commons-ui-core/src/NavBar/NavBar.stories.js
```diff
@@ -0,0 +1,49 @@
+import Box from "@mui/material/Box";
+import React from "react";
+
+import NavBar from "./NavBar";
```
gertie-sheshe, line 4: from `"."` ?

packages/commons-ui-core/src/NavBar/index.js
```diff
@@ -0,0 +1,3 @@
+import NavBar from "./NavBar";
+
+export default NavBar;
```
gertie-sheshe, line 3: So far we have `NavBar` exported twice. I know this is the approach you recommended, just curious if the double exportation is also intended

kilemensi: I only see a single `export` @gertie-sheshe but I kinda see what you mean. eslint doesn't allow use to export default [directly](https://eslint.org/docs/rules/no-restricted-exports#default-exports) similar to how [MUI](https://github.com/mui/material-ui/blob/dccbdbb25e0e335b8c1e0db13ec17ff0bd1ae7af/packages/mui-material/src/AppBar/index.js#L1) does it.

packages/commons-ui-core/src/NavBar/NavBar.stories.js
kilemensi: Yeah, it would work I think. I just want to be explicit here that `NavBar.{stories,test}.js` are for components in `NavBar.js` file since in some components, a single folder can contain multiple related [components](https://github.com/mui/material-ui/tree/master/packages/mui-material/src/ButtonBase) @gertie-sheshe

gertie-sheshe (approved):
🎉

PR #36, opened 2022-04-07
[Setup] Initial cfa website

nerdlet:
## Description

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/apps/cfasite/.eslintrc.js
+++ b/apps/cfasite/.eslintrc.js
@@ -1,0 +1,1 @@
+module.exports = require("eslint-config-commons-ui/next");
--- a/apps/cfasite/.lintstagedrc.js
+++ b/apps/cfasite/.lintstagedrc.js
@@ -1,0 +1,1 @@
+module.exports = require("eslint-config-commons-ui/.lintstagedrc");
--- a/apps/cfasite/README.md
+++ b/apps/cfasite/README.md
@@ -1,0 +1,11 @@
+# CFA WEBSITE
+
+This is the official cfa website
+
+## Getting Started
+
+First, run the development server:
+
+```bash
+pnpm  dev
+```
--- a/apps/cfasite/jest.config.js
+++ b/apps/cfasite/jest.config.js
@@ -1,0 +1,1 @@
+module.exports = require("jest-config-commons-ui/next");
--- a/apps/cfasite/jsconfig.json
+++ b/apps/cfasite/jsconfig.json
@@ -1,0 +1,7 @@
+{
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {}
+  },
+  "exclude": ["node_modules"]
+}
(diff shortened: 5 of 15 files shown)


kilemensi (commented):
👍🏽 ... Good stuff

-------------

1. Since the monorepo already separates `apps` from `packages`, we don't need to add `site` to the name of apps, we can just go with `cfa`. I prefer the full name `codeforafrica` to go with the domain but I'm not _against_ just `cfa`.
2. New app should mean new port so this will be running on port `3002` I believe.

--------------------------

**BONUS**

Setup import alias `@/codeforafrica` or `@/cfa` so that we don't use relative paths in import i.e.` '../../Something';`. You can follow the @comons-ui/core example for [`jsconfig.json`](https://github.com/CodeForAfrica/ui/blob/main/packages/commons-ui-core/jsconfig.json) and [`jest.config.js`](https://github.com/CodeForAfrica/ui/blob/main/packages/commons-ui-core/jest.config.js)

apps/codeforafrica/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "@codeforafrica",
```
kilemensi, line 2:
```suggestion
  "name": "codeforafrica",
```

apps/codeforafrica/playwright.config.js
```diff
@@ -0,0 +1,2 @@
+/* eslint-disable import/no-extraneous-dependencies */
+module.exports = require("playwright-config-commons-ui");
```
kilemensi, line 2: Need to configure webServer to run on port `3002` as well. See [`promisetracker`](https://github.com/CodeForAfrica/ui/blob/beed8171c0a43702b4b12baaf24056d1e1804df7/apps/promisetracker/playwright.config.js#L2-L14).

apps/codeforafrica/src/pages/index.spec.js
```diff
@@ -0,0 +1,7 @@
+const { test, expect } = require("@playwright/test");
+
+test("basic test", async ({ page }) => {
+  await page.goto("http://localhost:3000/");
```
kilemensi, line 4:
```suggestion
  await page.goto("http://localhost:3002/");
```

apps/codeforafrica/src/pages/index.test.js
```diff
@@ -0,0 +1,21 @@
+import { render, screen } from "@testing-library/react";
+import "@testing-library/jest-dom/extend-expect";
```
kilemensi, line 2:
```suggestion
import "@testing-library/jest-dom/extend-expect";
import React from "react";
```

kilemensi (commented):
👍🏽 

---

pnpm-lock.yaml may need updating as well.

kilemensi (approved):
🚀

PR #51, opened 2022-04-19
[Feature] Stay In Touch Component

gertie-sheshe:
## Description

Add StayInTouch Component

Fixes # (issue)

## Type of change

## Screenshots
Screenshot 2022-04-20 at 22 25 17

## Checklist:


kilemensi (commented):
👍🏽 ... Looks good.

----------------

1. Since we don't know how reusable in other apps this `StayInTouch` component will be, it's best to start implementing it in `codeforafrica` app first and move it to `commons-ui` only when we're sure it's reusable across more than one app. This is the approach I've taken in #53 as well.
2. I think we can simplify things ([Thinking in React](https://reactjs.org/docs/thinking-in-react.html) may help).
    i. Title can be its own component
    ii. Then all we're left with is a row of buttons with icons.
    iii. MUI already has IconButton which we can pass `component` to i.e. `<IconButton component={A or Link}` and we're done.

No need to customize `Grid`, `img`, etc.

PR #92, opened 2022-05-19
Automate Image building and deployment to Dokku

esirK:
## Description

Automate Image building and deployment to Dokku

## Type of change

## Checklist:


kilemensi (commented):
👍🏽 

----

If we're going to use a lot of workflows, then we need a better naming scheme for these workflows. I think `codeforafrica-build-deploy.yml` with `name: "codeforafrica.org | Build and Deploy` is a good start-ish.

.github/workflows/codeforafrica_app.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: Build Docker image
+        if: steps.version-check.outputs.changed == 'true'
+        uses: docker/build-push-action@v3
+        with:
+          build-args: |
+              PACKAGE_PATH=${{ env.PACKAGE_PATH }}
+          cache-from: type=local,src=/tmp/.buildx-cache
+          cache-to: type=local,dest=/tmp/.buildx-cache-new
+          context: .
+          push: false
```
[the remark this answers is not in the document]
esirK: yep.

Dockerfile
```diff
@@ -0,0 +1,27 @@
+FROM node:16-alpine as deps
```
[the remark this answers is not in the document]
esirK: I don't think/haven't seen something that breaks due to lack of it for now but I've added it since it's recomended.

.github/workflows/codeforafrica-build-deploy.yml
```diff
@@ -0,0 +1,87 @@
+name: codeforafrica.org | Build and Deploy
+on:
+  push:
+    branches: [main]
+
+env:
+  DOKKU_REMOTE_BRANCH: "master"
+  DOKKU_REMOTE_URL: "ssh://dokku@ui-1.prod.codeforafrica.org/code-for-africa-ui"
+  IMAGE_NAME: "codeforafrica/codeforafrica-ui"
```
kilemensi, line 9: Match image and Dokku app name if possible. If we're going with `codeforafrica-ui`, then both should be that. Likewise if we're going with `code-for-africa-ui` or just `codeforafrica`.

.github/workflows/codeforafrica-build-deploy.yml
```diff
@@ -0,0 +1,87 @@
...
+  PACKAGE_PATH: "codeforafrica"
+  FILE_NAME: "./apps/codeforafrica/package.json"
```
kilemensi, line 11:
```suggestion
  VERSION_FILE_NAME: "./apps/codeforafrica/package.json"
```

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+ARG PNPM_VERSION=7.1.1 \
+    APP \
+    # Next.js collects completely anonymous telemetry data about general usage.
+    # Learn more here: https://nextjs.org/telemetry
+    NEXT_TELEMETRY_DISABLED=1
+
+ENV APP=${APP} \
+    NEXT_TELEMETRY_DISABLED=${NEXT_TELEMETRY_DISABLED}
+
+WORKDIR /workspace/cfa_ui
```
kilemensi, line 14: Why not just
```suggestion
WORKDIR /workspace/
```
?

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+RUN corepack enable
+RUN corepack prepare pnpm@${PNPM_VERSION} --activate
```
kilemensi, line 17: Can't we combine these?
```suggestion
RUN corepack enable && \
    corepack prepare pnpm@${PNPM_VERSION} --activate
```

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+FROM base as dev
+
+COPY pnpm-lock.yaml .
+RUN pnpm fetch
```
kilemensi, line 22: Shouldn't these 2 be part of base since they're reusable across all apps?

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+COPY . .
```
kilemensi, line 24: Works but we can also just copy the files we need for the app we're building i.e. files in `root` folder and `apps/${APP}` folder.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+RUN pnpm --filter "${APP}" install --frozen-lockfile --unsafe-perm
```
kilemensi, line 26:
```suggestion
RUN pnpm --filter "${APP}" install --offline --frozen-lockfile --unsafe-perm
```

i. Since we've already fetched,` --offline` is good strategy here.
ii. Why do we need `--unsafe-perm`?

kilemensi (commented):
👍🏽 

---

For the last stage, can't we just create an image with built files?: https://github.com/CodeForAfrica/TwoopsTrackerUI/blob/cbecc0093f4c9501cb1c649c7ca8a2193129f3e0/Dockerfile#L26-L54

Dockerfile
esirK: I've updated it.

Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+
+FROM base as dev
+
+COPY *.yaml .
+COPY *.json .
+COPY packages ./packages
+COPY apps/${APP} ./apps/${APP}
+
+RUN pnpm --filter "${APP}" install --frozen-lockfile
+RUN pnpm --filter "${APP}" build
```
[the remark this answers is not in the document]
esirK: @kilemensi  I can update this although we have already performed the heavy lifting I think by only including files of the app we are building no?
Also, if we intend to use this approach we will have to remove this approach here 
```
RUN pnpm --filter "${APP}" install --frozen-lockfile
```
and install all dev dependencies.
I feel like `turbo` would help when we have multiple repos and but in this case, we have already removed/ignored all the other repos.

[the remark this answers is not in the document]
kilemensi: 1. apps don't depend on each; they depend on packages. Why would we need to install anything more that what a specific app needs.
2. Include `--offline` on `install` (Otherwise, there is no benefit of [`pnpm fetch`](https://pnpm.io/cli/fetch))
3. `turbo` [benefits](https://turborepo.org/docs#why-turborepo).

kilemensi (approved):
🚀 🚀

PR #107, opened 2022-06-03
Add Dockerfile to be used for ui deployment

esirK:
## Description

Add Dockerfile to be used for ui deployment

## Type of change

## Checklist:


esirK:
@kilemensi 
Instead of using the `contrib/dokku/Dockerfile` approach as we have done before, I thought having the specific dockerfile inside an individual app like a better approach. What do you think?

kilemensi:
> @kilemensi Instead of using the `contrib/dokku/Dockerfile` approach as we have done before (For dokku deployment), I thought having the specific dockerfile inside an individual app like a better approach. What do you think?

My only reservations with this approach is: `Dockerfile` at the root of an app always feels like a file to **build** an image of the app. But this `Dockerfile` is for **pulling** an already built image of the said app.

esirK:
> 

So placing it under `apps/codeforafrica/contrib/dokku/Dockerfile` would be a better approach @kilemensi ?

kilemensi:
> > 
> 
> So placing it under `apps/codeforafrica/contrib/dokku/Dockerfile` ...

I think so @esirK ... since it clearly shows if you aren't using Dokku, you probably don't need it.

esirK:
> I think so @esirK ... since it clearly shows if you aren't using Dokku, you probably don't need it.

All good now I think @kilemensi

kilemensi (approved):
🚀

PR #139, opened 2022-06-23
@/codeforafrica/pages/opportunity

### LINKED ISSUE #133: @/codeforafrica/pages/opportunites/[slug]

kilemensi:
## Description

Resolves #133 .

## Type of change

## Screenshots

Screenshot 2022-06-23 at 09-59-22 Security Evangelist Secure our defences against digital threats Stories Code for Africa

## Checklist:


apps/codeforafrica/src/components/Opportunities/Opportunities.js
```diff
@@ -0,0 +1,100 @@
...
+  const [selectedTag, setSelectedTag] = useState(ALL_TAGS);
+  const [page, setPage] = useState(pageProp);
+  const [pagination, setPagination] = useState(() => {
+    return computePagination(opportunities, page, pageSize);
+  });
+  const filteredOpportunies = useMemo(() => {
+    if (selectedTag === ALL_TAGS) {
+      return opportunities;
+    }
+    return opportunities.filter((p) =>
```
esirK, line 37: This works @kilemensi  but shouldn't we support multiple tags filtering? Right now I can't find Opportunities in Kenya and Remote.

apps/codeforafrica/src/components/Author/Author.js
```diff
@@ -1,61 +1,45 @@
...
 import PropTypes from "prop-types";
 import React from "react";
 
 const Author = React.forwardRef(function Author(props, ref) {
-  const { author, profession, ...other } = props;
+  const { name, profession, ...other } = props;
 
   return (
-    <Box
+    <Stack
```
esirK, line 10: Out of curiosity, why switch to `Stack`?

apps/codeforafrica/src/components/Opportunities/Opportunities.js
kilemensi: Possible @esirK and may be we will:

1. Looking at the design and looking an [Material UI](https://material.io/components/chips#choice-chips), these chips look more like Choice chips than Filter chips.
2. This exact same approach is taken for Projects, Stories, Team members and Opportunites. Once we know if all or some need to support Filtering while others Choosing, we can refactor.
3. There is also a search field that is currently missing (need CMS/API). Choice + search should be able to do Kenya + Remote.

esirK (approved):
LGTM

apps/codeforafrica/src/components/Author/Author.js
kilemensi: Best tool for the job @esirK :
```jsx
  <Box
      sx={{
        display: "flex",
        flexDirection: { xs: "column", md: "row" },
      }}
  >
```

is exactly equal to

```jsx
  <Stack
    direction={{ xs: "column", md: "row" }}
  >
```

esirK:
One more thing @kilemensi  what should the `Positions in:` be about?
Screenshot 2022-06-23 at 15 36 39

esirK:
Also seems like we are missing the top section as shown here
Screenshot 2022-06-23 at 15 39 16

kilemensi:
> One more thing @kilemensi what ...
>

From the designs, it's just the summary/excerpt of the story/opportunity. It can be anything @esirK 

### Story

Screenshot from 2022-06-23 15-38-57

### Opportunity

Screenshot from 2022-06-23 15-39-22

kilemensi:
> Also seems like we are missing the top section as shown here Screenshot 2022-06-23 at 15 39 16



> Also seems like we are missing the top section as shown here Screenshot 2022-06-23 at 15 39 16

Good catch @esirK ... Let me add that minus `by Code for Africa` and `comments`. We have author at the bottom of the article and we don't allow comments.

PR #152, opened 2022-06-30
@/codeforafrica/pages/member

### LINKED ISSUE #151: @/codeforafrica/pages/member
Finalize the /about/members/[slug] [page](https://www.figma.com/file/MbCygAaNhIjgJOHrQZpZ2U/CfA---CfA-site?node-id=7255%3A41546) by making sure:

- There is description,
- There is a Project list, and
- It has the right layout and size.

### Screenshot

Image

kilemensi:
## Description

Resolves #151 

## Type of change

## Screenshots

Screenshot 2022-06-30 at 13-50-45 Jacobo Ottaviani Members About Code for Africa

## Checklist:

### DIFF
--- a/apps/codeforafrica/src/components/AboutMemberPageHeader/AboutMemberPageHeader.snap.js
+++ b/apps/codeforafrica/src/components/AboutMemberPageHeader/AboutMemberPageHeader.snap.js
@@ -3,10 +3,10 @@
 exports[`<AboutMemberPageHeader /> renders unchanged 1`] = `
 <div>
   <div
-    class="MuiBox-root css-wpqv47"
+    class="MuiBox-root css-h16mw8"
   >
     <div
-      class="MuiContainer-root MuiContainer-maxWidthLg MuiContainer-fixed MuiContainer-disableGutters css-107e3nu-MuiContainer-root"
+      class="MuiContainer-root MuiContainer-maxWidthLg MuiContainer-fixed MuiContainer-disableGutters css-h42qkr-MuiContainer-root"
     >
       <div
         class="MuiGrid-root MuiGrid-container css-1fyy4z-MuiGrid-root"
--- a/apps/codeforafrica/src/components/ArticleHeader/ArticleHeader.snap.js
+++ b/apps/codeforafrica/src/components/ArticleHeader/ArticleHeader.snap.js
@@ -17,14 +17,14 @@ exports[`<ArticleHeader /> renders unchanged 1`] = `
         class="css-r7or14-MuiStack-root"
       >
         <button
-          aria-label="Facebook"
+          aria-label="Twitter"
           class="react-share__ShareButton"
           data-mui-internal-clone-element="true"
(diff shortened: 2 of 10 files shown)


apps/codeforafrica/src/components/RelatedProjects/RelatedProjects.js
```diff
@@ -1,40 +1,33 @@
...
-          }}
-        >
-          {title}
-        </RichTypography>
-        <ProjectTileList projects={projects.slice(0, 3)} />
-      </Section>
-    </Box>
+        {title}
+      </RichTypography>
+      <ProjectTileList {...tileListProps} projects={projects.slice(0, 3)} />
```
esirK, line 29: What happens if someone has more than 3 projects?

apps/codeforafrica/src/lib/index.js
```diff
@@ -1591,6 +1601,7 @@ function getTeamMemberPageStaticProps(params) {
       }) === 0
   );
   if (member) {
+    const startIndex = getRandomInt(projects.length - 3);
```
esirK, line 1604: I see; Randomly show 3 projects 👍

apps/codeforafrica/src/pages/about/members/[slug].page.js
```diff
@@ -21,32 +15,60 @@ function Index({ member, sections, ...props }) {
...
       {sections?.map((section) => {
         switch (section.slug) {
           case "related-projects":
-            return <RelatedProjects {...section} key={section.slug} />;
+            return (
+              <React.Fragment key={section.slug}>
+                <SectionDivider
+                  sx={{
+                    maxWidth: {
+                      sm: "648px",
```
Comment (esirK) [inline apps/codeforafrica/src/pages/about/members/[slug].page.js:50]: The values of most of these viewports seem the same; Couldn't we have them in a 'global' location?

esirK (approved):
LGTM

apps/codeforafrica/src/components/RelatedProjects/RelatedProjects.js
kilemensi: `ProjectTileList` uses `Grid` so it can show more than 3 projects. The above code is just to match current designs only. Should/Will be removed once we switch to a functional CMS.

apps/codeforafrica/src/pages/about/members/[slug].page.js
Comment (kilemensi) [reply apps/codeforafrica/src/pages/about/members/[slug].page.js:50]: Ideally [theme](https://github.com/CodeForAfrica/ui/blob/14638e687cf125cff0332ba461939490cc2fb19e/apps/codeforafrica/src/theme/index.js#L32-L40) is where they would have been but some pages like this one uses different sizes... If you check [Figma](https://www.figma.com/file/MbCygAaNhIjgJOHrQZpZ2U/CfA---CfA-site?node-id=7255%3A41478), you can see even the above is just estimation i.e. it changes from one section to the next. I think we'd need to confirm with the designer first as part of the review.

PR #153, opened 2022-06-30
Adds Projects collection and their relation to other pages

esirK:
## Description
- Projects Collection
- Relate projects to index page

## Type of change

##### Partners 
Screenshot 2022-07-01 at 15 53 14

Screenshot 2022-06-30 at 15 07 16

##### Projects
Screenshot 2022-07-01 at 15 53 53

## Checklist:


esirK:
@kilemensi To prevent having to review 50+ files, I think this can be reviewed independently for Project.

apps/codeforafrica/content/projects/africanDrone.md
```diff
@@ -0,0 +1,42 @@
+---
```
kilemensi, line 1: Shouldn't the filename be lower case i.e. `africandrone`?

kilemensi (commented):
👍🏽 ... Looks dope.

----

1. Can't really remember why we used JSON for `config` but we should just change it to a JS file. Having to type `""` on everything is just tiresome.
1. Optional but I think arranging the "data" stuff under a single prefix can help organizing the dashboard. I've used Data but it can be something else.
Screenshot from 2022-07-01 10-17-59
2. I don't think we should have slug as an editable field. We should use use the generated filename as slug (and we can look into something like https://github.com/netlify/netlify-cms/issues/1975#issuecomment-1033773082 for adding read only/hidden permanent ID to relate items)
3. I think Partners at least should also have `href` (required could be false) since I think they're meant to be clickable on the design)... but of course we can add it to Donors as well.

apps/codeforafrica/content/projects/africanDrone.md
esirK: Netlify uses `"identifier_field": "name",` (can be changed to another field) to determine the filename hence the name since the project name is `africanDrone`

kilemensi: No, I'm cool with it using name @esirK ... I'm just surprised that `africanDRONE` would result in `africanDrone` while `WanaData` would result in `wanadata`. I was under the impression the "slugify" function that NetlifyCMS uses would lower-case the names first.

esirK: lemme check that. Might be the file was created before I specified the `identifier_field`

apps/codeforafrica/src/pages/admin/UUIDWidget.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable */
```
kilemensi, line 1: We don't need this file.

apps/codeforafrica/src/pages/admin/index.page.js
```diff
@@ -3,6 +3,8 @@ import Head from "next/head";
 import Script from "next/script";
 import React from "react";
 
+import CustomWidget from "@/codeforafrica/pages/admin/UUIDWidget";
```
kilemensi, line 6:
```suggestion
```

apps/codeforafrica/src/pages/admin/index.page.js
```diff
@@ -17,6 +19,8 @@ function AdminPage(props) {
       </Head>
       <NextSeo {...props} />
       <Script src="https://unpkg.com/netlify-cms@^2.0.0/dist/netlify-cms.js" />
+      <Script src="https://unpkg.com/netlify-cms-widget-uuid-v4@^1.0.12/dist/index.js" />
+      <CustomWidget />
```
kilemensi, line 23: I think lets use normal async script loading for now so that we're sure things are loaded before `load` event is fired.
```jsx
/* eslint-env browser */
import { NextSeo } from "next-seo";
import Head from "next/head";
import React from "react";

function AdminPage(props) {
  return (
    <>
      <Head>
        {/* Note the "type" and "rel" attribute values, which are required. */}
        <link
          href="/api/admin/config.yml"
          type="text/yaml"
          // eslint-disable-next-line react/no-invalid-html-attribute
          rel="cms-config-url"
        />
        <script
          src="https://unpkg.com/netlify-cms@^2.0.0/dist/netlify-cms.js"
          async
        />
        <script
          src="https://unpkg.com/netlify-cms-widget-uuid-v4@^1.0.12/dist/index.js"
          async
        />
        <script
          // eslint-disable-next-line react/no-danger
          dangerouslySetInnerHTML={{
            __html: `
              window.addEventListener("load", () => {
                window.CMS.registerWidget(
                  "uuid",
                  window.uuidWidget.UuidControl
                );
              });
            `,
          }}
        />
      </Head>
      <NextSeo {...props} />
    </>
  );
}

export default AdminPage;

```


... And I'm skipping registering Preview because the UUID is not part of content & hence shouldn't be visible to the users.

apps/codeforafrica/src/pages/api/admin/config.js
```diff
@@ -0,0 +1,294 @@
...
+              display_fields: ["name"],
+              multiple: true,
+            },
+          ],
+        },
+      ],
+    },
+    {
+      name: "partners",
+      label: "Data | Partners",
```
kilemensi, line 64:
```suggestion
      label: "Data | Partners",
      label_singular: "Partner"
```

So that the button show "New Partner" instead of "New Data | Partners"

kilemensi (approved):
🚀 

---

Lets update the content so that UUIDs are stored in markdown files.

apps/codeforafrica/src/pages/admin/UUIDWidget.js
esirK: I thought having all custom widgets(registration) under one file is better than always adding them under the index page every time we need a new widget.

PR #158, opened 2022-07-05
Ft about page

esirK:
## Description

Adds 
- About Us Page
- Teams collection
- 
## Type of change

## Screenshots
Screenshot 2022-07-05 at 12 17 32

Screenshot 2022-07-05 at 12 17 22

Screenshot 2022-07-05 at 12 17 11

Screenshot 2022-07-06 at 12 21 12

## Checklist:


apps/codeforafrica/src/pages/api/admin/config.js
```diff
@@ -281,5 +342,84 @@ module.exports = {
...
+        {
+          label: "Image",
+          name: "image",
+          widget: "image",
+        },
+        {
+          label: "Projects",
+          name: "projects",
+          widget: "relation",
+          label_singular: "Project",
```
kilemensi, line 382: For obvious cases, NetlifyCMS handles label singular well enough.

kilemensi (commented):
👍🏽 

---

In [about](https://www.figma.com/file/MbCygAaNhIjgJOHrQZpZ2U/CfA---CfA-site?node-id=7255%3A38953) page, I think all the 👇🏽 collections should be displayed in full i.e. no need to select individual items:

- Guiding principles,
- Team,
- Partners, and
- Impact

apps/codeforafrica/content/team/jacobo-ottaviani.md
```diff
@@ -0,0 +1,20 @@
...
+  et rhoncus sem, laoreet tortor tristique cursus ultrices. Sed urna nec
+  convallis in tempus lacinia velit viverra neque, molestie malesuada fringilla
+  ut vulputate faucibus ad potenti, nisi enim nam montes nostra curae ridiculus
+  curabitur. Elementum phasellus conubia eget ac dignissim hendrerit ullamcorper
+  magna pretium pulvinar, auctor dui pelle.**"
+image: https://res.cloudinary.com/code-for-africa/image/upload/v1655127335/codeforafrica/images/team/image_11_ch6dnb.jpg
+projects:
+  - 14fdf60f-2928-4776-b208-6641e8413f74
+  - afcfdec9-99d7-4f2c-8bdb-964159c0676d
+  - 07f3259e-2379-4850-83b8-85cacde0fcb8
```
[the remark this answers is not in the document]
esirK: Currently not. 
I think defining team members in a Project is the better approach since a project has multiple members and thus we won't need to duplicate projects for different team members.

esirK:
> 👍🏽
> 
> In [about](https://www.figma.com/file/MbCygAaNhIjgJOHrQZpZ2U/CfA---CfA-site?node-id=7255%3A38953) page, I think all the 👇🏽 collections should be displayed in full i.e. no need to select individual items:
> 
> * Guiding principles,
> * Team,
> * Partners, and
> * Impact

@kilemensi  This sounds good for *Guiding principles* and *Impact*; However, for *Team* and *Partners*,  these are collections on their own and we can only use a *Relation* to add them to the *About* page. Do you suggest we be adding the Team/Partners directly in the About us page?

esirK:
> 

@kilemensi For Guiding Principles, I don't see like they are used anywhere else other than the About page. Wouldn't it be better to just be able to edit them directly from this page?

kilemensi:
> @kilemensi For Guiding Principles, I don't...
>

Yes, I don't think they're visible anywhere outside the about page. Having them as a stand-alone collection i.e. `Data | Guiding Principles` will make it more consistent for content editors since that's how almost everything else is edited **but** it's not a must in this case.

apps/codeforafrica/src/pages/api/admin/config.js
esirK: In this case since the `Label` was updated to `Data | Projects` it isn't able to translate it to just `Project`

kilemensi (approved):
🚀 

----

Looks good and ready to merge but there are a few things we need to consider may be on this PR or separately.

1. I think guiding principles would need to be user sortable to drive message home. Again, I could be misremembering but I think the `List` widget allows items to be sorted drag-n-drop style. So `List` of `Object` could do the trick.
2. Most of the non-markdown widget should actually be `string` instead of `text`. _names_, _href_, _tag lines_, _short titles_, etc. There is rarely need for multiline text input that's not rich text/markdown.

PR #165, opened 2022-07-13
CMS footer page

### LINKED ISSUE #163: @codeforafrica site content
Using [`blog-starter`](https://github.com/vercel/next.js/tree/canary/examples/basic-export) example (or any other that make heavy use of `Markdown`), pull `Markdown` content into the site replacing the current `lib/index.js` placeholder content.

**NOTE**: Ensure the content "schema" is in sync between Netlify CMS widgets and the code that converts Markdown frontmatter to html.

esirK:
## Description

Adds CMS Footer page

Resolves #163 

## Type of change

#### Settings
Screenshot 2022-08-02 at 15 44 51

#### Header Navigation
Screenshot 2022-08-02 at 15 45 01

Screenshot 2022-08-02 at 15 45 12

#### Footer Navigation
Screenshot 2022-08-02 at 15 45 24

Screenshot 2022-08-02 at 15 45 42

Screenshot 2022-08-02 at 15 45 51

Screenshot 2022-08-02 at 15 46 02

Screenshot 2022-08-02 at 15 46 13

## Checklist:


kilemensi (commented):
👍🏽 ... Looks good

----

`Footer` is not a page. Thinking about this further, may be we shouldn't have a `Footer` section. Both `Footer` and `Header` can pull content from other sections such as below (just a suggestion, feel free to improve on it).

1. **Settings** section (WP calls it appearance, Ghost calls it settings)
1.1 **General** (WP calls it Site identity, Ghost calls it General). Here we include `title`, `description`, and `logo`. We may need 2 logos, primary (used in `Header`) and secondary (used in `Footer`).
1.2 **Navigation** (WP calls it Menu, Ghost calls it navigation): Here we include primary navigation (same in `Header` and `Footer`) and secondary navigation (only in `Footer`. I think this is called Page links in the current PR).

Social media and Subscribe embed code can go in multiple places: In `Settings > General` as well, in their own `Data | ---` sections, or we can add a `Data | Footer` section and include these two since this is the only place they appear in (however, `Twitter` needs to also appear in `Header` so we may have to think a bit out that).

esirK:
@kilemensi  All this makes sense. I've updated the code and also the Screenshots.

For this
> 
> Social media and Subscribe embed code can go in multiple places: In `Settings > General` as well, in their own `Data | ---` sections, or we can add a `Data | Footer` section and include these two since this is the only place they appear in (however, `Twitter` needs to also appear in `Header` so we may have to think a bit out that).

I decided to go with an approach where a user can set if an item belongs to the header or the footer using a boolean field.

kilemensi (commented):
👍🏽 ... I like the thinking

-------------

1. The 2 navigations have 2 distinct designs (font-size at least). This means we need to know which is which. Selecting that navigation item should go in footer or header is not enough.
Screenshot from 2022-07-19 15-06-14
2. Navigation should be sortable i.e. moving **Stories** before **About** shouldn't require code change.

Given the above points, it makes sense to me to have 2 distinct navigations: primary and secondary. And they should both be built using `list` widget which allows item sorting.

Another approach could be having completely separate **Header Navigation** and **Footer Navigation**. While this will have duplication since one has to repeat stories, about, etc., in both menus, it will however make it easier to set location specific logos and/or icons. For example Header Navigation will have `black` Twitter logo only but Footer Navigation can have _white_ logos for all social media i.e. Stay in Touch. Using [variable list types](https://www.netlifycms.org/docs/beta-features/#list-widget-variable-types) should allow us to easily create list items of different content type I think.

esirK:
@kilemensi  I've updated the code and screenshots too.
Still thinking if the `Imprint` and `Privacy Policy should be treated as `external-links` or not.

apps/codeforafrica/content/settings/header-navigation.md
```diff
@@ -0,0 +1,26 @@
...
+      content: Stories
+      href: /stories
+  - link:
+      content: Opportunities
+      href: /opportunities
+  - link:
+      content: Contact
+      href: /contact
+logo:
+  src: http://localhost:3002/_next/image?url=%2F_next%2Fstatic%2Fmedia%2FCfAlogoBW.2d9d4139.png&w=256&q=75
```
kilemensi, line 19: 1. Weird link
2. Header logo should be the coloured one.

kilemensi (commented):
👍🏽 ... Cool, lets go with this structure

--------------

1. Should be `Settings > Header` and `Settings > Footer` because we now have more than just navigation.
2. `Settings > Header`: Arrange items as they appear on the site: logo, main navigation, and social media link (single item). For the social media link, one must be able to set different icons for mobile (white) and desktop (black).
3. `Settings > Footer`: Arrange items as they appear on the site: logo, description and stay in touch in column one, main navigation and secondary navigation (not external links) in the 2nd column, and subscription last.

apps/codeforafrica/src/pages/api/admin/config.js
```diff
@@ -930,5 +930,314 @@ module.exports = {
...
+                          ],
+                        },
+                      ],
+                    },
+                  ],
+                },
+                {
+                  label: "Main Navigation Links",
+                  name: "main-navigation-links",
+                  label_singular: "Main Navigation Link",
```
kilemensi, line 1163: 1. Move this outside Social/Stay in touch. It should be a standalone widget under Footer.
2. Can it just link/reference `Main navigation` in Header?
```suggestion
                  label: "Main Navigation",
                  name: "main-navigation",
                  label_singular: "Navigation Link",
```

apps/codeforafrica/src/pages/api/admin/config.js
```diff
@@ -930,5 +930,314 @@ module.exports = {
...
+                          name: "href",
+                          widget: "string",
+                        },
+                      ],
+                    },
+                  ],
+                },
+                {
+                  label: "Subscription Link",
+                  name: "subscription-link",
```
kilemensi, line 1214:
```suggestion
                  label: "Newsletter subscription",
                  name: "newsletter-subscription",
```

apps/codeforafrica/src/pages/api/admin/config.js
```diff
@@ -930,5 +930,314 @@ module.exports = {
...
+                          widget: "string",
+                        },
+                        {
+                          label: "Href",
+                          name: "href",
+                          widget: "string",
+                        },
+                      ],
+                    },
+                  ],
```
kilemensi, line 1234:
```suggestion
                  widget: "code",
                  allow_language_selection: false,
                  default_language: "html"
```

We just need the code (similar to [this](https://github.com/CodeForAfrica/ui/blob/99af7c008a5834ba713d095398ec3eadc6b34210/apps/codeforafrica/src/lib/index.js#L130-L144))

apps/codeforafrica/src/pages/api/admin/config.js
```diff
@@ -930,5 +930,314 @@ module.exports = {
...
+                },
+                {
+                  label: "Secondary Navigation Links",
+                  name: "secondary-navigation-links",
+                  label_singular: "Secondary Navigation Link",
```
kilemensi, line 1189:
```suggestion
                  label: "Secondary Navigation",
                  name: "secondary-navigation",
                  label_singular: "Navigation Link",
```

kilemensi (approved):
🚀

apps/codeforafrica/src/pages/api/admin/config.js
esirK: For `Can it just link/reference Main navigation in Header?` unless we make Navigation an individual `Collection` and use `relation` field.

PR #167, opened 2022-07-14
Ft pull cms content

esirK:
## Description
Pull and display CMS content into the CFA site

- Pull Donors' content
- Pull Hero and Our Team Content
- Pull Badges Content
- Pull Partners Content
- Pull Projects Content

## Type of change

## Screenshots
N/A

## Checklist:


apps/codeforafrica/package.json
```diff
@@ -43,7 +43,9 @@
     "@mui/utils": "^5.8.4",
     "fast-equals": "^4.0.1",
     "frontmatter-markdown-loader": "^3.7.0",
```
kilemensi, line 45: Shouldn't `frontmatter-markdown-loader` be removed?

kilemensi (commented):
:+1: ... Looks good.

Still think about the api (folder or file should be named `api.netlify-cms` or `api.netlify-cms.js` however since it's more common to name packages using `kebab-case` than `snake_case` ìn JS e.g. `material-ui`, `testing-library/react`, `node-mysql`, etc.).As far as CMS content is concerned though:

### Eliminate the need to map description / address

We can eliminate majority of schema mapping by naming the main field in folder collections `body` as recommended in [docs](https://www.netlifycms.org/docs/configuration-options/#fields). Hence renaming `description` or `address` in the following collections to `body` (leaving label to what it is) will mean we don't have to remember that these fields are markdown and `matter` will convert them to HTML automatically as `content`.

>Badges, Donors, Guiding Principles, Offices, Our impact, Partners, Projects, and Team

# Let theme do the heavy lifting

Some fields can be converted to simple string or text if multiline is required since the theme already does the necessary styling (**bold**, etc.) These include:

> `title` in Team, and `subtitle` in About us/Hero

### Remaining required mapping

Only the following fields will hence still require manually mapping:

> `description` in About us/Our mission, `title` in Homepage/Hero and `description` in Homepage/Meet our team.


### Others

I think `count` in **Our impact** should be changed to `value` since it can be more than just a number e.g. **81+**.

esirK:
This sounds good @kilemensi .I think I had missed the use of the `body` field.

kilemensi:
> Ola @kilemensi For
> 
> > ...
> 
> I tested this out and seems like the `content` isn't converted and thus had to use `marked`.
> 

Cool @esirK as long as `matter` separates `content` from `data`, we're all set.

> For
> 
> > ...
> 
> Since we are currently using `string`, it already allows for this.

Is it named/labelled as value as well @esirK?

apps/codeforafrica/src/lib/api.netlify-cms/mappings.js
```diff
@@ -0,0 +1,22 @@
+const Project = {
```
kilemensi, line 1: Not sure about this mapping file structure e.g. why do we need `title`, etc. My current thinking would be to remain close to the Netlify CMS nomenclature i.e. collections. Something like
```js
export const markdownContent = {
  projects: {
    "*": {
      subtitle: true,
    },
  },
  pages: {
    index: {
      "hero.title": true,
    },
  },
}
```

1. `projects` and `pages` are **collections**
2. `*` and `index` are slug for _folder_ and _file_ collection respectively (`*` means applicable to all slugs since all items in a folder collection have the same structure, while `index` means match `index` slug only), and
3. "hero.title" means title in hero object obviously.

We can use JS's optional chaining to convert that string e.g.
```js
var indexPage = { hero: { title: "Civic Tech Heroes" } }; 
function getObjectField(object, path) {
  return path.split(".").reduce((acc, curr) => acc?.[curr], object);
}

console.log(getObjectField(indexPage, "hero.title")); // Civic Tech Heroes
```

apps/codeforafrica/src/lib/api.netlify-cms/getProjects.js
```diff
@@ -0,0 +1,57 @@
...
+export default function getProjects(fields = []) {
+  const slugs = getCollectionSlugs(projectsDir);
+  const projects = [];
+  slugs.forEach((_slug) => {
+    const collection = getCollectionBySlug(projectsDir, _slug, fields);
+    const project = collection.items;
+    if (fields.includes("badges")) {
+      const badges = getBadges(["id", "name", "content", "date"]);
+      project.badges = badges.filter((badge) =>
+        collection.data.badges.id.includes(badge.id)
```
kilemensi, line 19: Isn't this in project too i.e. why not use `project.badges.id` or `collections.items.badges.id`? Mixing `collection.items` and `collection.data` is kind of confusing.

apps/codeforafrica/src/lib/api.netlify-cms/getProjects.js
```diff
@@ -0,0 +1,57 @@
...
+        "id",
+        "slug",
+        "name",
+        "content",
+        "href",
+        "logo",
+      ]);
+      project.partners = {
+        title: "Partners",
+        list: partners.filter((partner) =>
```
kilemensi, line 33: Should we check first if we do indeed have partners i.e. `partners?.filter`? Otherwise partners could be `undefined`.

apps/codeforafrica/package.json
```diff
@@ -42,8 +42,9 @@
     "@mui/material": "^5.8.5",
     "@mui/utils": "^5.8.4",
     "fast-equals": "^4.0.1",
-    "frontmatter-markdown-loader": "^3.7.0",
+    "gray-matter": "^4.0.3",
     "js-yaml": "^4.1.0",
+    "marked": "^4.0.18",
```
kilemensi, line 47: Why not use `remark`? [Netlify CMS](https://www.netlifycms.org/docs/beta-features/#remark-plugins) and Next.js examples seem to prefer it.

apps/codeforafrica/src/lib/api.netlify-cms/utils.js
```diff
@@ -0,0 +1,44 @@
...
+  const fullPath = join(collectionDir, `${realSlug}.md`);
+  const fileContents = fs.readFileSync(fullPath, "utf8");
+  const { data, content } = matter(fileContents);
+  const items = {};
+
+  fields.forEach((field) => {
+    if (field === "content") {
+      items.description = marked(content);
+    } else {
+      items[field] = data[field] || "";
```
kilemensi, line 24:
```suggestion
      items[field] = data[field] || null;
```

Or is there a reason to prefer `""`?

kilemensi (commented):
👍🏽 ... getting close.

kelvinkipruto (approved):
LGTM

apps/codeforafrica/src/lib/api.netlify-cms/utils.js
```diff
@@ -0,0 +1,44 @@
...
+            });
+          });
+        } else {
+          data[field] = marked(data[field]);
+          items[field] = data[field];
+        }
+      });
+    }
+  });
+  return { items, data }; // return data just incase the caller needs to access its contents
```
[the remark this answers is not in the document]
esirK: yes, that's correct.

apps/codeforafrica/package.json
esirK: For this, I think it's just a preference. But `marked` seems to be more actively maintained and it's standalone not requiring other plugins e.g `remark-html`

apps/codeforafrica/src/lib/api.netlify-cms/sections/getMeetOurTeam.js
```diff
@@ -0,0 +1,15 @@
+import { join } from "path";
```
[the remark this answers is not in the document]
esirK: @kilemensi  are you referring to how the `index.js` file knows about different `sections`? e.g by 
```
import getMeetOurTeam from "./api.netlify-cms/sections/getMeetOurTeam";
```

apps/codeforafrica/src/lib/api.netlify-cms/getProjects.js
```diff
@@ -0,0 +1,58 @@
...
+export default function getProjects(fields = []) {
+  const slugs = getCollectionSlugs(projectsDir);
+  const projects = [];
+  slugs.forEach((_slug) => {
+    const collection = getCollectionBySlug(projectsDir, _slug, fields);
+    const project = collection.items;
+    if (fields.includes("badges")) {
+      const badges = getBadges(["id", "name", "content", "date"]);
+      project.badges = badges?.filter((badge) =>
+        project.badges.id.includes(badge.id)
```
kilemensi, line 21: Why do we have `badges.id` and not just badges? (similar to just `donors` or `projects`).

apps/codeforafrica/src/lib/api.netlify-cms/getProjects.js
```diff
@@ -0,0 +1,58 @@
...
+    }
+    if (fields.includes("donors")) {
+      const donors = getDonors();
+      project.donors = {
+        title: "Donors",
+        list: donors?.filter((donor) => project.donors.includes(donor.id)),
+      };
+    }
+    if (fields.includes("links")) {
+      project.links = collection.data.links.map(
```
kilemensi, line 48:
```suggestion
      project.links = collection.links.map(
```

apps/codeforafrica/src/lib/api.netlify-cms/sections/getMeetOurTeam.js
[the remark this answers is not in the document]
kilemensi: Yes @esirK . From API _user_ point of view, is there any reason why the 2 below are different?

```js
import getMeetOurTeam from "./api.netlify-cms/sections/getMeetOurTeam";
```
```js
import getProjects from "./api.netlify-cms/getProjects";
```

apps/codeforafrica/src/lib/api.netlify-cms/utils.js
```diff
@@ -0,0 +1,29 @@
+import fs from "fs";
+import { join } from "path";
+
+import matter from "gray-matter";
+import { marked } from "marked";
+
+export const getCollectionSlugs = (collectionDir) => {
```
kilemensi, line 7: Since we use `function getCollectionSlugs`, lets keep using that. Or is there a way to mix it up?

apps/codeforafrica/src/lib/api.netlify-cms/utils.js
```diff
@@ -0,0 +1,29 @@
...
+
+  fields.forEach((field) => {
+    if (field === "content") {
+      items.description = marked(content);
+    } else if (field === "slug") {
+      items.slug = realSlug;
+    } else {
+      items[field] = data[field] || null;
+    }
+  });
```
kilemensi, line 26: 1. We still assign `items.description = marked(content);`, are we sure **_every_** collection is like that? For example **imprint** page, `content` is `content`.
2. `reduce` is perfect here:
```suggestion
  const items = fields.reduce((acc, curr) => {
    if (curr === "content") {
      acc["content"] = marked(content);
    } else if (curr === "slug") {
      acc["slug"] = realSlug;
    } else {
      acc[field] = data[field] || null;
    }
    return acc;
  }, {});
```

apps/codeforafrica/src/lib/api.netlify-cms/getCollectionData.js
```diff
@@ -0,0 +1,12 @@
...
+
+export default function getCollectionData(collectionDir, fields = []) {
+  const slugs = getCollectionSlugs(collectionDir);
+  const collections = slugs.reduce((acc, slug) => {
+    const collection = getCollectionBySlug(collectionDir, slug, fields);
+    return fields.length
+      ? acc.concat(collection.items)
+      : acc.concat(collection.data);
+  }, []);
+  return collections;
```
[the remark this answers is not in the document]
esirK: >Not sure I fully follow: if we don't have any fields, why do we return data? Shouldn't we just return {} or undefined

This was to allow users to have a default behavior where if `fields` aren't specified then we give them everything which in this case is data.
So; Performing a `getDonors()` will return all `Donors` data without filtering specific fields. If we only want some fields then we can do a `getDonors(['id', 'name'])`. This way the API user won't be surprised by not getting any data if they don't know about passing in the fields.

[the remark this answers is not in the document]
kilemensi: That makes sense but it's the opposite of how the API is designed. In [`getCollectionBySlug`](https://github.com/CodeForAfrica/ui/pull/167/files#diff-ca52c6797a8c76b83bf228b65ab57a656cca06ccc1185d44464a60dc5f099cd4R11), by default `fields = []` i.e. API user expects nothing to be returned.

apps/codeforafrica/src/lib/api.netlify-cms/sections/getMeetOurTeam.js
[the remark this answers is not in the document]
kilemensi: It's not a JS thing, it's just section feels like an implementation detail here. Guiding principles, Partners, etc., are all sections in there respective pages but we don't include section in their import paths.

If I read the code correctly, the only real reason we have section here is because `MeetOurTeam` and `Hero` are part of file-based collections while the rest are folder-based collections. If this is true, then this **is** an implementation detail that users shouldn't really have to deal with.

apps/codeforafrica/src/lib/api.netlify-cms/getCollectionData.js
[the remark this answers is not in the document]
esirK: the reason for having `fields = []` is to remove the need to do a `fields` check before doing the `reduce`. Without this, I think the code will look like 
```
function getCollectionBySlug(collectionDir, slug, fields) {
 if (fields){
    items = fields.reduce(
````

which I think also makes sense.

File: apps/codeforafrica/content/projects/pesacheck.md [hunk 14]
```diff
@@ -2,41 +2,41 @@
...
-
-  <p>PesaCheck fact-checks in two international languages (English and French), as well as major African languages such as Kiswahili and Amharic. Our network helps track political promises by politicians (through our Wajibisha/PromiseTracker toolkit), helps unpack budget and census data (through our PesaYetu and TaxClock platforms), and builds machine learning/artificial intelligence tools (such as DebunkBot) to help automate verification.</p>
-category: Projects
+tag: Initiatives
 icon:
   src: https://res.cloudinary.com/code-for-africa/image/upload/v1652431239/codeforafrica/icons/Type_PesaCheck_rmswvg.svg
 thumbnail:
   src: https://res.cloudinary.com/code-for-africa/image/upload/v1652705959/codeforafrica/images/Property_1_PesaCheck_iahlrh.jpg
-href: /projects/pesa-check
+href: /projects/pesacheck
```
kilemensi, line 15: I guess in most (all?) of these items, we won't need internal `href` anymores since we can generate them from `slug`.

apps/codeforafrica/src/pages/api/admin/config.js
```diff
@@ -677,19 +721,31 @@ module.exports = {
...
-              name: "href",
-              widget: "string",
+              label: "Link",
+              name: "link",
+              widget: "object",
+              fields: [
+                {
+                  label: "slug",
+                  name: "slug",
+                  widget: "string",
```
kilemensi, line 736: Does any item need a manually entered `slug` any more?

apps/codeforafrica/src/lib/api.netlify-cms/utils.js
```diff
@@ -0,0 +1,28 @@
...
+    if (curr === "content") {
+      acc.content = marked(content);
+    } else if (curr === "slug") {
+      acc.slug = realSlug;
+    } else {
+      acc[curr] = data[curr] || null;
+    }
+    return acc;
+  }, {});
+  return { items, data }; // return data which can be used as default incase of no fields
```
kilemensi, line 27: I'm still not 💯 on this: Say no fields is requested, we return `data`. Cool... the problem however is `data` doesn't include `content` or `realSlug` (could be using slug with `.md`). So, in the end, when you don't specify fields, you really still don't get everything.

kilemensi (approved):
🚀 🚀

apps/codeforafrica/src/pages/api/admin/config.js
esirK: For these two I think they have to be set but I'll confirm

Screenshot 2022-07-21 at 11 53 32

apps/codeforafrica/src/lib/api.netlify-cms/utils.js
esirK: For me, I still think there should be a default behavior which as you had said earlier is returning all the data we have read in the markdown file. We then let the consumer decide on what to do with the data. So on this, I would suggest just including `content` into the returned data which is all 'data' we got from the MD.

apps/codeforafrica/src/pages/api/admin/config.js
esirK: [here](https://github.com/CodeForAfrica/ui/blob/main/apps/codeforafrica/src/components/ProjectDetails/ProjectDescriptionButton.js#L14) @kilemensi

apps/codeforafrica/src/lib/api.netlify-cms/utils.js
kilemensi: Cool, then lets return _all_ data, including `content` and `slug`.

apps/codeforafrica/src/pages/api/admin/config.js
kilemensi: Cool. Looking at that code, the purpose of `slug` was to pick appropriate icon since icons are/were hard-coded. Now that we have CMS, shouldn't the icon be set here as part of the link?

PR #168, opened 2022-07-18
ft pulled ghost cms data

### LINKED ISSUE #164: @codeforafrica publication content

kelvinkipruto:
## Description

Pulled Ghost CMS data for stories and opportunity pages

Resolves #164 

## Type of change

## Screenshots
N/A
## Checklist:


apps/codeforafrica/src/pages/opportunities/[slug].page.js
```diff
@@ -3,6 +3,7 @@ import React from "react";
 import ArticlePage from "@/codeforafrica/components/ArticlePage";
 import Page from "@/codeforafrica/components/Page";
 import { getPageStaticProps } from "@/codeforafrica/lib";
+import { getAllPostsWithSlug } from "@/codeforafrica/lib/api";
```
Comment (kilemensi) [inline apps/codeforafrica/src/pages/opportunities/[slug].page.js:6]: Instead of including CMS-specific code in pages, we can defined `getPageStaticPath` in lib/index.js (similar to getPageStaticProps). This way pages don't need to be know which CMS back-end a certain piece of content comes from.

kilemensi (commented):
👍🏽 ... Looks good.

-----

Since we're using **2** CMSes, I think lets rename `api.js` to `api.ghost.js`.

kilemensi:
RE: `camelCase` vs `snake_case` @kelvinkipruto, it looks like it's a lot of props that would need to be excluded from lint. While I haven't used it myself, can't something like https://github.com/sindresorhus/camelcase-keys help us here or do you think it's not worthy it?

cc @esirK

apps/codeforafrica/src/lib/api.js
```diff
@@ -0,0 +1,69 @@
+import GhostContentAPI from "@tryghost/content-api";
+
+// const GHOST_API_URL = process.env.GHOST_CMS_URL;
+const GHOST_API_URL = "https://cms.codeforafrica.org";
```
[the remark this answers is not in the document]
kelvinkipruto: ooh, yes. I was having some issue when using the environment variables, and forgot to change it back. I will update it.

apps/codeforafrica/src/lib/api.js
```diff
@@ -0,0 +1,69 @@
...
+const GHOST_API_KEY = process.env.GHOST_CMS_API_KEY;
+
+const api = new GhostContentAPI({
+  url: GHOST_API_URL,
+  key: GHOST_API_KEY,
+  version: "v3.0",
+});
+
+export async function getAllPosts(limit = 10, page = 1) {
```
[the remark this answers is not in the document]
kelvinkipruto: I agree. This is a much better approach.

apps/codeforafrica/src/lib/api.ghost.js
```diff
@@ -0,0 +1,71 @@
...
+  const posts = api.posts.read({
+    slug,
+    fields:
+      "id,title,slug,published_at,feature_image,excerpt,custom_excerpt,excerpt,html",
+    include: "authors",
+  });
+  return posts;
+}
+
+export async function getAllTags(options) {
```
kelvinkipruto, line 35: @kilemensi Any idea of how to filter the tags based on another tag? For example, we could want to only get tags for stories alone.

apps/codeforafrica/src/components/ArticlePage/ArticlePage.js
```diff
@@ -80,18 +79,14 @@ function ArticlePage({
           px: { xs: 2.5, sm: 0 },
         }}
       >
-        <Author {...author} />
+        {/* 
+        //TODO: confirm if this is to be shown on opportunities page
+         */}
+
+        <Author {...(primary_author || authors[0])} />
```
kelvinkipruto, line 86: @kilemensi Author should only be shown only on a story page, and not on the opportunity page right?

kilemensi: I kinda of agree @kelvinkipruto but [designs](https://www.figma.com/file/MbCygAaNhIjgJOHrQZpZ2U/CfA---CfA-site?node-id=7255%3A39753) says we should show author in both.

apps/codeforafrica/src/lib/api.ghost.js
kilemensi: No I don't have any @kelvinkipruto ... Lets just pull all posts by `primary_tag` and then concat tags from each post either manually or using [`helpers`](https://ghost.org/docs/content-api/javascript/#helpers).

apps/codeforafrica/src/components/ArticlePage/ArticlePage.js
kilemensi: 1. There is a feedback review period after development where we can raise these issues. Since we do have the value, lets show it for now.
2. I believe there is `bio` field for authors. We can use that for now, no?

kelvinkipruto: Ok. I agree. Lets use `bio` for now.

apps/codeforafrica/.env.example
```diff
@@ -0,0 +1,6 @@
+NEXT_PUBLIC_APP_LOGO_URL=https://res.cloudinary.com/code-for-africa/image/upload/v1656078512/codeforafrica/logos/CfA_logo_BW_twlmo6.png
+NEXT_PUBLIC_APP_NAME="Code for Africa"
+NEXT_PUBLIC_APP_URL="http://localhost:3000"
+NEXT_PUBLIC_IMAGE_DOMAINS="res.cloudinary.com,cms.codeforafrica.org"
+GHOST_URL=
+GHOST_API_KEY=
```
kilemensi, line 6: We have `env.template` that serves as an example.

apps/codeforafrica/next.config.js
```diff
@@ -12,9 +12,9 @@ const outputFileTracingRoot = PROJECT_ROOT
   ? path.resolve(__dirname, PROJECT_ROOT)
   : undefined;
 
-const ghostUrl =
-  process.env.GHOST_ADMIN_URL?.trim() || process.env.GHOST_URL?.trim();
-const ghostAdminUrl = new URL("/ghost", ghostUrl).toString();
+// const ghostUrl =
+//   process.env.GHOST_ADMIN_URL?.trim() || process.env.GHOST_URL?.trim();
+// const ghostAdminUrl = new URL("/ghost", ghostUrl).toString();
```
kilemensi, line 17: No commented out code in repo. `git log -p` can show needed history.

apps/codeforafrica/package.json
```diff
@@ -41,10 +41,13 @@
     "@googlemaps/react-wrapper": "^1.1.35",
     "@mui/material": "^5.8.5",
     "@mui/utils": "^5.8.4",
+    "@tryghost/content-api": "^1.11.0",
+    "camelcase-keys": "^8.0.1",
```
kilemensi, line 45: If we're not using a package, remove it.

apps/codeforafrica/src/components/ArticleCardList/ArticleCardList.js
```diff
@@ -23,31 +23,31 @@ const ArticleCardList = React.forwardRef(function ArticleCardList(props, ref) {
...
             sx={{
               width: "100%",
               height: "100%",
             }}
           >
             <CardActionArea
-              component={article.href ? Link : undefined}
-              href={article.href}
+              component={article.slug ? Link : undefined}
+              href={`/stories/${article.slug}`}
```
kilemensi, line 35: Can we add this in the lib? `ArticleXXX` components are designed to be reusable for both opportunities and stories.

apps/codeforafrica/src/components/ArticleCardList/ArticleCardList.js
```diff
@@ -23,31 +23,31 @@ const ArticleCardList = React.forwardRef(function ArticleCardList(props, ref) {
...
+                src={article.featureImage}
               />
               <ArticleCardContent>
                 <Typography variant="subtitle1">{article.title}</Typography>
                 <Typography
                   sx={{ color: "#9F9494", display: "block", mt: 2 }}
                   variant="caption"
                 >
-                  {new Date(article.date).toLocaleDateString("en", {
+                  {new Date(article.publishedAt).toLocaleDateString("en", {
```
kilemensi, line 50: Lets also do formatting in lib.

apps/codeforafrica/src/components/ArticlePage/ArticlePage.js
```diff
@@ -31,13 +30,13 @@ function ArticlePage({
...
-        src={image?.src}
+        src={featureImage}
       />
       <ArticleHeader
         title={title}
-        date={date}
+        date={publishedAt}
         tags={tags}
-        summary={summary}
+        summary={excerpt || customExcerpt}
```
kilemensi, line 39: Lets also do this in lib... I think `customExcerpt` means hand-written excerpt, right? if yes, then it should take precedence over general `excerpt`.

apps/codeforafrica/src/components/ArticlePage/ArticlePage.js
```diff
@@ -80,18 +79,10 @@ function ArticlePage({
           px: { xs: 2.5, sm: 0 },
         }}
       >
-        <Author {...author} />
+        <Author {...(primaryAuthor || authors[0])} />
```
kilemensi, line 82: Every post should have a primary author, no?

apps/codeforafrica/src/components/ArticlePage/ArticlePage.test.js
```diff
@@ -10,6 +10,8 @@ const render = createRender({ theme });
 
 const defaultProps = {
   coverImage: { src: "/images/stories-1.png" },
+  authors: [],
+  featureImage: "/images/stories-1.png",
 };
```
kilemensi, line 15: Lets just remove default props.

apps/codeforafrica/src/components/OpportunityCardList/OpportunityCardList.js
```diff
@@ -16,7 +16,7 @@ const OpportunityCardList = React.forwardRef(function OpportunityCardList(
     <Stack spacing={{ xs: 5, md: 10 }} {...other} ref={ref}>
       {opportunities.map((opportunity) => (
         <OpportunityCard
-          key={opportunity.href || opportunity.title}
+          key={opportunity.slug || opportunity.title}
```
kilemensi, line 19: I think `id` is better key since it's stable (not related to content).

apps/codeforafrica/src/lib/api.ghost.js
```diff
@@ -0,0 +1,88 @@
+import GhostContentAPI from "@tryghost/content-api";
+
+import convertToCamelCase from "@/codeforafrica/utils/camelcaseKeys";
+
+const { GHOST_URL } = process.env;
+// const { GHOST_URL } = "https://longform.codeforafrica.org";
+const { GHOST_API_KEY } = process.env;
```
kilemensi, line 7: 1. Don't destructure environment variables. See a note about this in the [docs](https://nextjs.org/docs/basic-features/environment-variables#loading-environment-variables).
2. Include `GHOST_API_VERSION` as part of envs as well (defaulting to `"v5.0"`);

apps/codeforafrica/src/lib/api.ghost.js
```diff
@@ -0,0 +1,88 @@
...
+
+export async function getAllPosts(options) {
+  const { limit = 10, page = 1, ...other } = options || {};
+
+  const api = new GhostContentAPI({
+    url: GHOST_URL,
+    key: GHOST_API_KEY,
+    version: "v3.0",
+  });
```
kilemensi, line 16: Move this into a function initializeAPI so that you don't have to duplicate the code.

apps/codeforafrica/src/lib/api.ghost.js
```diff
@@ -0,0 +1,88 @@
...
+
+  const posts = await api.posts.read({
+    slug,
+    fields:
+      "id,title,slug,published_at,feature_image,excerpt,custom_excerpt,excerpt,html",
```
kilemensi, line 38: excerpt is repeated

apps/codeforafrica/src/lib/index.js
```diff
@@ -655,7 +661,56 @@ function getProjectTags(options = { includeAll: true }) {
...
+}
+
+export async function getStories(options) {
+  const {
+    tag: originalTag,
+    page,
+    "page-size": pageSize = 10,
+    q,
+  } = options || {};
+  const tag = originalTag || "stories";
```
kilemensi, line 697: I think we're mixing **normal** tags (`Kenya`, `Misinformation`, etc.) with **primary** tags (`Stories` and `Opportunities`). The tag here should be one of the _normal_ tags (or All) since the method name specifically says these are stories.

apps/codeforafrica/src/lib/index.js
```diff
@@ -1334,3 +1333,15 @@ export async function getPageStaticProps(params) {
       return { notFound: true };
   }
 }
+
+export async function getGhostCMSStaticPaths(primaryTag, options) {
```
kilemensi, line 1337:
```suggestion
export async function getPageStaticPaths(primaryTag, options) {
```
There shouldn't be any need to include `GhostCMS` in the name (similar to `getPageStaticProps`

kilemensi (commented):
👍🏽 . Good start.

apps/codeforafrica/next.config.js
kelvinkipruto: Yes, I commented this out when testing why CI was failing and forgot to change it back.

apps/codeforafrica/src/components/ArticlePage/ArticlePage.js
kelvinkipruto: @kilemensi How do we do this in the lib? I'm not sure how to approach it.

kilemensi: The same way/place you do the camelCase conversion @kelvinkipruto . e.g. you can add something link :point_down: [here](https://github.com/CodeForAfrica/ui/blob/72d551ed4301c58f045a960023413f9073a83987/apps/codeforafrica/src/lib/api.ghost.js#L18-L25).
```js
  const posts = await api.posts.browse({
    limit,
    page,
    fields: "id,title,slug,published_at,feature_image,excerpt",
    include: "authors,tags",
    ...other,
  });
  return posts.map((post) => {
    const { customExcerpt, excerpt: originalExcerpt, ...other } = convertToCamelCase(post);
    const excerpt = customExcerpt || originalExcerpt || null;
    // Any other manipulation can also be done here
    return { ...other, excerpt }; 
  });
```

kelvinkipruto: Ok, so instead of returning both customExcerpt and excerpt, we just return one of them?

kilemensi: Exactamundo! We only need one.

apps/codeforafrica/src/components/ArticleCardList/ArticleCardList.js
kelvinkipruto: @kilemensi resolving this, and I think it could be much better to do it in the `lib/index.js` because we need to either append either stories or opportunites to the slug, allowing us to differentiate between the two.

kilemensi: Sure @kelvinkipruto , in `lib` or in `api.ghost` since we we know primary tags.

apps/codeforafrica/src/components/FeaturedArticle/FeaturedArticle.js
```diff
@@ -32,8 +33,8 @@ const FeaturedArticle = React.forwardRef(function FeaturedArticle(props, ref) {
       ref={ref}
     >
       <CardActionArea
-        component={href ? Link : undefined}
-        href={href}
+        component={slug ? Link : undefined}
+        href={slug}
```
kilemensi, line 37: `slug` is a slug and `href` is a href... These should be 2 different props.

apps/codeforafrica/src/components/OpportunityCardList/OpportunityCardList.js
```diff
@@ -16,7 +16,7 @@ const OpportunityCardList = React.forwardRef(function OpportunityCardList(
     <Stack spacing={{ xs: 5, md: 10 }} {...other} ref={ref}>
       {opportunities.map((opportunity) => (
         <OpportunityCard
-          key={opportunity.href || opportunity.title}
+          key={opportunity.id || opportunity.title}
```
kilemensi, line 19:
```suggestion
          key={opportunity.id}
```
Do you envision a situation where an opportunity will now have an id?

apps/codeforafrica/src/lib/api.ghost.js
```diff
@@ -0,0 +1,101 @@
+import GhostContentAPI from "@tryghost/content-api";
+
+import convertToCamelCase from "@/codeforafrica/utils/camelcaseKeys";
+
+function ghostAPI() {
+  return new GhostContentAPI({
+    url: process.env.GHOST_URL,
+    key: process.env.GHOST_API_KEY,
+    version: process.env.GHOST_API_VERSION || "v5",
```
kilemensi, line 9:
```suggestion
    version: process.env.GHOST_API_VERSION || "v5.0",
```
I think Ghost Content API [version](https://ghost.org/docs/faq/api-versioning/) is always in `v{major}.{minor}` format.

apps/codeforafrica/src/lib/api.ghost.js
```diff
@@ -0,0 +1,101 @@
...
+  });
+}
+
+function transformGhostPost(post) {
```
kilemensi, line 13:
```suggestion
function transformPost(post) {
```
We know it's Ghost from file name.

apps/codeforafrica/src/lib/api.ghost.js
```diff
@@ -0,0 +1,101 @@
...
+
+  const excerpt = customExcerpt || originalExcerpt || null;
+  const publishedAt = new Date(publishedAtRaw).toLocaleDateString("en", {
+    year: "numeric",
+    month: "short",
+    day: "numeric",
+  });
+  return { ...other, excerpt, publishedAt };
+
+  // return convertToCamelCase(post);
```
kilemensi, line 29:
```suggestion
```

apps/codeforafrica/src/lib/api.ghost.js
```diff
@@ -0,0 +1,101 @@
...
+    ...options,
+    fields:
+      "id,title,slug,published_at,feature_image,excerpt,custom_excerpt,html",
+    filter: `primary_tag:${tag}`,
+    include: "tags",
+  };
+
+  const posts = await getAllPosts(optionsWithPrimaryTagFilter);
+  return posts;
+}
```
kilemensi, line 89: Lets think a bit about these 2 method. Yes, we want to `getPostsByPrimaryTag` i.e. `stories` or `opportunites` but we also want to go a step further e.g. get **stories** with tag `kenya`, get **opportunities** with tag `remote`.

It feels like the 2nd method should be `getPostsByPrimaryTagAndTag` or something like that.

apps/codeforafrica/src/lib/api.ghost.js
```diff
@@ -0,0 +1,101 @@
...
+  const optionsWithPrimaryTagFilter = {
+    ...options,
+    fields: "slug",
+    limit: "all",
+    filter: `primary_tag:${tag}`,
+  };
+  // const posts = await api.posts.browse(params);
+  const posts = await getAllPosts(optionsWithPrimaryTagFilter);
+  return posts;
+}
```
kilemensi, line 101: From the name, I expect to be filtering on slug but looks like we're filtering on tags, and primary tags at that. What's the difference between this and `getPostsByPrimaryTag`?

apps/codeforafrica/src/lib/index.js
```diff
@@ -655,7 +661,72 @@ function getProjectTags(options = { includeAll: true }) {
...
+    found = await getPostsByPrimaryTag("stories");
+  } else {
+    found = await getPostsByTag(tag, options);
+  }
+
+  // append /stories/ to the slug && transform published date to ISO format
+  found = found.map((a) => {
+    const { slug: originalSlug, ...other } = a;
+    const slug = `/stories/${originalSlug}`;
+    return { slug, ...other };
```
kilemensi, line 714: 1. `slug` should remain `slug`, create a new key `href`.
2. this can be done as part of `transformPost` by using `primaryTag` value.

apps/codeforafrica/src/lib/index.js
```diff
@@ -776,18 +834,22 @@ export function getOpportunities(options) {
...
+  // append /opportunities/ to the slug && transform published date to ISO format
+  allOpportunities.map((a) => {
+    const { slug: originalSlug, ...other } = a;
+    const slug = `/opportunities/${originalSlug}`;
+    return { slug, ...other };
+  });
+
+  const allTags = allOpportunities
+    .map((article) => article.tags)
+    .flat(Infinity);
```
kilemensi, line 849: What is `.flat(Infinity)`? Doesn't `flatMap` work?
```suggestion
  const allTags = allOpportunities
    .flatMap((article) => article.tags);
```

apps/codeforafrica/src/pages/stories/[slug].page.js
```diff
@@ -35,12 +35,10 @@ function Index({ article, sections, ...props }) {
...
 export async function getStaticPaths() {
-  const paths = [...Array(13).keys()].map((_, i) => ({
-    params: { slug: `article-${i + 1}` },
-  }));
+  const staticPaths = await getPageStaticPaths("stories");
   return {
-    paths,
-    fallback: false,
+    paths: staticPaths.length > 0 ? staticPaths : [],
+    fallback: true,
```
Comment (kilemensi) [inline apps/codeforafrica/src/pages/stories/[slug].page.js:41]: Shouldn't fallback here also be `blocking`?

apps/codeforafrica/src/utils/camelcaseKeys.js
```diff
@@ -0,0 +1,9 @@
+import { isArray, isObject, transform, camelCase } from "lodash";
+
+export default function convertToCamelCase(obj) {
+  return transform(obj, (acc, value, key, target) => {
+    const camelKey = isArray(target) ? key : camelCase(key);
+
+    acc[camelKey] = isObject(value) ? convertToCamelCase(value) : value;
+  });
+}
```
kilemensi, line 9: I posted a fix for ESM on Slack. We shouldn't need to import lodash just for this.
>Just had a look at the camelcase-keys / Jest issue Kelvin and the issue seem to be camelcase-keys is built as ESM. You just need to tell Jess to transform it by adding something like :point_down: to `jest.config.js`:
>
> ```js
>   transformIgnorePatterns: ["<rootDir>/node_modules/(?!camelcase-keys)"],
> ```

apps/codeforafrica/src/components/ArticlePage/ArticlePage.js
```diff
@@ -80,18 +77,10 @@ function ArticlePage({
           px: { xs: 2.5, sm: 0 },
         }}
       >
-        <Author {...author} />
+        <Author {...authors[0]} />
```
kilemensi, line 80: Why are we just using `primaryAuthor`?

kilemensi (commented):
kelvinkipruto

apps/codeforafrica/src/lib/api.ghost.js
kelvinkipruto: According to [this](https://ghost.org/docs/content-api/javascript/#:~:text=version%20%2D%20should%20be%20set%20to%20%E2%80%98v3%E2%80%99) , we should set it to v3. Perhaps v3.0?

kilemensi: In the **Working Example** section in the same link you shared, they're using `v5.0`. Not sure why they're saying it should be `v3` here.

Either way, as long as it works and it's in `v{major}.{minor}`, I'm happy.

apps/codeforafrica/src/utils/camelcaseKeys.js
```diff
@@ -0,0 +1,11 @@
...
+
+export default function convertToCamelCase(obj) {
+  if (Array.isArray(obj)) {
+    return obj.map((item) => convertToCamelCase(item, { deep: true }));
+  }
+  if (typeof obj === "object" && obj !== null) {
+    return camelcaseKeys(obj, { deep: true });
+  }
+  return obj;
+}
```
kilemensi, line 11: We don't need this function. `camelcase-keys` should be able to handle objects and arrays natively.

apps/codeforafrica/src/lib/api.ghost/posts.js
```diff
@@ -0,0 +1,97 @@
...
+
+const cacheDir = path.join(process.env.PWD, "public/data");
+
+function ghostAPI() {
+  return new GhostContentAPI({
+    url: process.env.GHOST_URL,
+    key: process.env.GHOST_API_KEY,
+    version: process.env.GHOST_API_VERSION || "v5.0",
+  });
+}
```
kilemensi, line 16:
```suggestion
import { promises as fs } from "fs";
import path from "path";

import camelcaseKeys from "camelcase-keys";
import GhostContentAPI from "@tryghost/content-api";

import equalsIgnoreCase from "@/codeforafrica/utils/equalsIgnoreCase";

const cacheDir = path.join(process.env.PWD, "public/data");

function initializeContentAPI() {
  return new GhostContentAPI({
    url: process.env.GHOST_URL,
    key: process.env.GHOST_API_KEY,
    version: process.env.GHOST_API_VERSION || "v5.0",
  });
}
```

1. since we're doing async for fetch, lets do async for reading files as well.
2. We're already in api.ghost so we shouldn't repeat Ghost if we can.

apps/codeforafrica/src/lib/api.ghost/posts.js
```diff
@@ -0,0 +1,97 @@
...
+
+function transformPost(post) {
+  const {
+    customExcerpt,
+    slug,
+    primaryTag,
+    publishedAt: publishedAtRaw,
+    excerpt: originalExcerpt,
+    ...other
+  } = convertToCamelCase(post);
```
kilemensi, line 26:
```suggestion
const {
    customExcerpt,
    excerpt: originalExcerpt,
    primaryTag,
    publishedAt: publishedAtRaw,
    slug,
    tags: originalTags,
    ...other
  } = camelcaseKeys(post, { deep: true });
```
1. Prefer listing props alphabetically,
2. It's better to also process all tags here once and for all.

apps/codeforafrica/src/lib/api.ghost/posts.js
```diff
@@ -0,0 +1,97 @@
...
+    day: "numeric",
+  });
+  return {
+    ...other,
+    excerpt,
+    publishedAt,
+    href,
+    slug,
+    primaryTag,
+  };
```
kilemensi, line 42:
```suggestion
  const excerpt = customExcerpt || originalExcerpt || null;
  const href = `/${primaryTag.slug}/${slug}`;
  const publishedAt = new Date(publishedAtRaw).toLocaleDateString("en", {
    year: "numeric",
    month: "short",
    day: "numeric",
  });
  const tags = originalTags
    .filter((t) => equalsIgnoreCase(t.name, "opportunities") || equalsIgnoreCase(t.name, "stories"))
    .map((t) => t.name);
  return {
    ...other,
    excerpt,
    href,
    primaryTag,
    publishedAt,
    slug,
    tags,
  };
```
Opportunites and Stories should only be in primary tag.

apps/codeforafrica/src/lib/api.ghost/posts.js
```diff
@@ -0,0 +1,97 @@
...
+}
+
+async function cachePosts(posts) {
+  const cacheFile = path.join(cacheDir, "posts.json");
+  const data = {
+    date: new Date().toISOString(),
+    posts,
+  };
+  fs.writeFileSync(cacheFile, JSON.stringify(data));
```
kilemensi, line 51:
```suggestion
  await fs.writeFile(cacheFile, JSON.stringify(data));
```

apps/codeforafrica/src/lib/api.ghost/posts.js
```diff
@@ -0,0 +1,97 @@
...
+}
+
+async function getCachedPosts() {
+  try {
+    const cacheFile = path.join(cacheDir, "posts.json");
+    const data = fs.readFileSync(cacheFile);
```
kilemensi, line 57:
```suggestion
    const data = await fs.readFile(cacheFile, "utf8");
```

apps/codeforafrica/src/lib/api.ghost/posts.js
```diff
@@ -0,0 +1,97 @@
...
+  const posts = await api.posts.browse({
+    include: "authors,tags",
+  });
+
+  const allPosts = posts.map(transformPost);
+
+  // Cache the posts
+  await cachePosts(allPosts);
+
+  return allPosts;
```
kilemensi, line 90:
```suggestion
  // If not, fetch from Ghost
  const api = initializeContentAPI();
  const posts = await api.posts.browse({
    include: "authors,tags",
  });
  const allPosts = posts.map(transformPost);
  // Cache the posts
  await cachePosts(allPosts);

  return allPosts;
```
1. Why do you leave so many empty lines?
2. Shouldn't `limit: "all"` be included? in `browse`?

apps/codeforafrica/src/lib/api.ghost/posts.js
```diff
@@ -0,0 +1,97 @@
...
+}
+
+export async function getPost(slug) {
+  const allPost = await getAllPosts();
+  const post = allPost.find((p) => p.slug === slug);
+  return post;
```
kilemensi, line 96:
```suggestion
  return allPost.find((p) => p.slug === slug);
```

apps/codeforafrica/src/pages/stories/[slug].page.js
```diff
@@ -35,16 +35,14 @@ function Index({ article, sections, ...props }) {
...
   return {
-    paths,
-    fallback: false,
+    paths: staticPaths.length > 0 ? staticPaths : [],
+    fallback: "blocking",
   };
 }
 
-export async function getStaticProps({ params: { slug } }) {
+export function getStaticProps({ params: { slug } }) {
```
Comment (kilemensi) [inline apps/codeforafrica/src/pages/stories/[slug].page.js:45]:
```suggestion
export async function getStaticProps({ params: { slug } }) {
```

apps/codeforafrica/src/lib/api.ghost/index.js
```diff
@@ -0,0 +1,47 @@
+import { getAllPosts, getPost } from "@/codeforafrica/lib/api.ghost/posts";
```
kilemensi, line 1:
```suggestion
import { getAllPosts, getPost } from "@/codeforafrica/lib/api.ghost/posts";
 import equalsIgnoreCase from "@/codeforafrica/utils/equalsIgnoreCase";
```

apps/codeforafrica/src/lib/api.ghost/index.js
```diff
@@ -0,0 +1,47 @@
...
+  );
+  return opportunities;
+}
+
+export async function getAllOpportunitiesTags() {
+  const opportunities = await getAllOpportunities();
+  const tags = opportunities.flatMap((post) => post.tags);
+  const uniqueTags = [...new Set(tags.map((tag) => tag.name))];
+  uniqueTags.unshift("All");
+  return uniqueTags;
```
kilemensi, line 16:
```suggestion
  return ["All", ...new Set(tags)];
```

apps/codeforafrica/src/lib/api.ghost/index.js
```diff
@@ -0,0 +1,47 @@
...
+  );
+  return stories;
+}
+
+export async function getAllStoriesTags() {
+  const stories = await getAllStories();
+  const tags = stories.flatMap((post) => post.tags);
+  const uniqueTags = [...new Set(tags.map((tag) => tag.name))];
+  uniqueTags.unshift("All");
+  return uniqueTags;
```
kilemensi, line 37:
```suggestion
  return ["All", ...new Set(tags)];
```

apps/codeforafrica/src/lib/api.ghost/index.js
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export async function getStory(slug) {
+  let story = await getPost(slug);
+  // remove tags from post
+  // NOTE: For some reason, the single page breaks if the tags are included
+  story = { ...story, tags: [] };
+
+  return story;
```
kilemensi, line 46:
```suggestion
  return getPost(slug);
```

The error was because tags were being returned as objects but the single page post expects tags to be strings hence the need to process tags in `transformPost`.

kilemensi (approved):
🚀 🚀 ... We Have Liftoff, We Have Liftoff

----

Can merge after addressing the few issues below.

apps/codeforafrica/src/lib/api.ghost/index.js
kelvinkipruto: Perhaps we no longer need the getStoriesTags and getOpportunitesTags since we can just filter out the tags from the list of either stories or opportunities.

apps/codeforafrica/src/lib/api.ghost/posts.js
kelvinkipruto: @kilemensi Doesn't using this remove all the other tags, and only remain with the primary tag? Should we instead use  `!equalsIgnoreCase` so as to include other tags other than the primary tag?

kilemensi: 💯 @kelvinkipruto . It should be
```js
    .filter((t) => !(equalsIgnoreCase(t.name, "opportunities") || equalsIgnoreCase(t.name, "stories")))
```

apps/codeforafrica/src/lib/api.ghost/index.js
kilemensi: We need to filter **and** make sure they're unique. I still think a standalone function is much better than duplicating this functionality.

apps/codeforafrica/src/lib/api.ghost/posts.js
```diff
@@ -0,0 +1,109 @@
...
+    day: "numeric",
+  });
+  const tags = originalTags
+    .filter(
+      (t) =>
+        !(
+          equalsIgnoreCase(t.name, "opportunities") ||
+          equalsIgnoreCase(t.name, "stories")
+        )
+    )
```
kilemensi, line 44: Having thought about this further, I think we can simplify it to just:
```suggestion
      (t) => !equalsIgnoreCase(t.name, primaryTag.name)
    )
```

`primaryTag.name` will always be `Opportunities` or `Stories`, right?

apps/codeforafrica/src/lib/api.ghost/posts.js
```diff
@@ -0,0 +1,109 @@
...
+
+  const posts = await api.posts.browse({
+    include: "authors,tags",
+    limit: "all",
+  });
+
+  const allPosts = posts.map(transformPost);
+
+  // Cache the posts
+  await cachePosts(allPosts);
```
kilemensi, line 101: Why so many empty lines between code?
```suggestion
  // If not, fetch from Ghost
  const api = initializeContentAPI();
  const posts = await api.posts.browse({
    include: "authors,tags",
    limit: "all",
  });
  const allPosts = posts.map(transformPost);
  // Cache the posts
  await cachePosts(allPosts);
```

apps/codeforafrica/src/lib/api.ghost/index.js
```diff
@@ -0,0 +1,35 @@
+import { getAllPosts, getPost } from "@/codeforafrica/lib/api.ghost/posts";
+
+export async function getAllOpportunities() {
+  const allPosts = await getAllPosts();
+  return allPosts.filter(
+    (post) => post.primaryTag.name.toLowerCase() === "opportunities"
```
kilemensi, line 6: Use `equalsIgnoreCase` instead of generating new string with `toLowerCase`/`toUpperCase`.
```suggestion
    (post) => equalsIgnoreCase(primaryTag.name, "opportunities")
```

apps/codeforafrica/src/lib/api.ghost/index.js
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export async function getOpportnity(slug) {
+  return getPost(slug);
+}
+
+export async function getAllStories() {
+  const allPosts = await getAllPosts();
+  return allPosts.filter(
+    (post) => post.primaryTag?.name.toLowerCase() === "stories"
```
kilemensi, line 23: `equalsIgnoreCase`

kilemensi (approved):
🚀

kelvinkipruto:
@kilemensi There is an issue when displaying a single post when deploying on Vercel, example [here](https://codeforafrica-p67qec95b-codeforafrica.vercel.app/stories/battle-for-gender-equality-in-african-media-continues). The response error is `
INVALID_IMAGE_OPTIMIZE_REQUEST`. From what I can find out online, it could be related to the NextJS version. Any idea how to resolve it, or we have to bump NextJS version?

kilemensi:
> @kilemensi There is an issue when displaying a single post when deploying on Vercel, example [here](https://codeforafrica-p67qec95b-codeforafrica.vercel.app/stories/battle-for-gender-equality-in-african-media-continues). The response error is ` INVALID_IMAGE_OPTIMIZE_REQUEST`. From what I can find out online, it could be related to the NextJS version. Any idea how to resolve it, or we have to bump NextJS version?

Not sure about that @kelvinkipruto. All image optimisations should be disabled by default: https://github.com/CodeForAfrica/ui/blob/99af7c008a5834ba713d095398ec3eadc6b34210/apps/codeforafrica/.env#L5

kilemensi:
> > @kilemensi There is an is ...
> 
> Not sure about that @kelvinkipruto. All image optimisations should be disabled by default:
> 
> https://github.com/CodeForAfrica/ui/blob/99af7c008a5834ba713d095398ec3eadc6b34210/apps/codeforafrica/.env#L5

Did you find out what's causing this @kelvinkipruto ?

PR #174, opened 2022-07-21
@/codeforafrica client side search

kilemensi:
## Description

This PR implements client side search using [Fuse.js](https://fusejs.io/). Search is available on the following pages:

 - [x] **Our work**,
 - [x] **Stories**,
 - [x] **About**, and
 - [x] **Opportunities**

## Type of change

## Screenshots

Peek 2022-07-21 15-46

## Checklist:


apps/codeforafrica/src/components/Articles/Articles.js
```diff
@@ -1,64 +1,87 @@
...
+
+  const handleChangePage = (_, value) => {
     setPage(value);
   };
-  const [pagination, setPagination] = useState(() => {
-    return computePagination(getFilteredArticles(), page, pageSize);
-  });
+
+  const handleChangeQ = (_, value) => {
+    setQ(value || undefined);
```
esirK, line 37: do we need the `|| undefined`? won't `value` be undefined if not supplied?

esirK (approved):
LGTM

kelvinkipruto (commented):
What if rather than filtering all item based on the search text, display the title and maybe context of the searched text under the search bar, i.e display a few words next or after the text searched.

kelvinkipruto (approved):
LGTM

apps/codeforafrica/src/components/Articles/Articles.js
kilemensi: Coming from a text field, we get empty string (`""`) instead of `undefined`.

kilemensi:
> What if rather than filtering all item based on the search text, display the title and maybe context of the searched text under the search bar, i.e display a few words next or after the text searched.

Good suggestions @kelvinkipruto but for MVP we try to stick close to the design so that we can give comprehensive feedback and make any/necessary improvements afterwards.

PR #184, opened 2022-07-27
Enable GitHub auth on Netlify CMS

esirK:
## Description

Adds Github Authentication on our Netlify CMS

## Type of change

## Screenshots
Screenshot 2022-07-27 at 11 38 11

## Checklist:


apps/codeforafrica/src/pages/api/admin/config.js
```diff
@@ -1,7 +1,9 @@
 module.exports = {
   backend: {
-    name: "git-gateway",
+    name: "github",
     branch: "main",
+    repo: "CodeForAfrica/ui",
```
kilemensi, line 5: Can't we set these up in the [`config.yml.page`](https://github.com/CodeForAfrica/ui/blob/b96a2604e2f8393f5246522c69580272e12bc1fb/apps/codeforafrica/src/pages/api/admin/config.yml.page.js#L9-L19) for production use only i.e. default should still be local development

apps/codeforafrica/src/pages/api/oauth/auth.js
```diff
@@ -0,0 +1,37 @@
...
+      )}/api/oauth/callback`,
+      scope: "user, repo",
+    },
+  },
+};
+
+const client = new AuthorizationCode({
+  client: {
+    id: process.env.OAUTH_CLIENT_ID,
+    secret: process.env.OAUTH_CLIENT_SECRET,
```
kilemensi, line 21: These should be added in the `env.template` for documentation.

kilemensi (approved):
🚀

apps/codeforafrica/src/pages/api/admin/config.js
esirK: With this setup, the default is still `local` just that we don't have to use the `git-gateway` backend

kilemensi: Cool, cool... I think the setup was copied from the [docs](https://www.netlifycms.org/docs/beta-features/#working-with-a-local-git-repository) directly. e.g. not sure if we need `proxy_url` either.

PR #186, opened 2022-07-29
Fix  collections folder location

esirK:
Fixes # (issue)
Collections folder location

## Type of change

## Checklist:

### DIFF
--- a/apps/codeforafrica/.env
+++ b/apps/codeforafrica/.env
@@ -4,9 +4,9 @@ NEXT_PUBLIC_APP_URL="https://cfa.dev.codeforafrica.org"
 NEXT_PUBLIC_IMAGE_DOMAINS="longform.codeforafrica.org,res.cloudinary.com"
 NEXT_PUBLIC_IMAGE_UNOPTIMIZED="true"
 NEXT_PUBLIC_VERCEL_URL=${VERCEL_URL}
+NEXT_PUBLIC_APP_DIRECTORY="apps/codeforafrica/"
 GHOST_URL="https://longform.codeforafrica.org"
 GITHUB_BACKEND_REPO="codeforafrica/ui"
 GITHUB_BACKEND_BRANCH="main"
 GITHUB_AUTH_ENDPOINT="api/oauth/auth"
 GIT_REPO_DIRECTORY=../../
-
--- a/apps/codeforafrica/.env
+++ b/apps/codeforafrica/.env
@@ -8,3 +8,5 @@ GHOST_URL="https://longform.codeforafrica.org"
 GITHUB_BACKEND_REPO="codeforafrica/ui"
 GITHUB_BACKEND_BRANCH="main"
 GITHUB_AUTH_ENDPOINT="api/oauth/auth"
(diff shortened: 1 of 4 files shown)


kilemensi (commented):
Doesn't this change require any modifications in terms of how the CMS is run locally? I'm getting empty content for all `Data | XXXX` items.
Screenshot from 2022-08-01 09-57-13

esirK:
> Doesn't this change require modifying how the CMS is run locally? I'm getting empty content for all `Data | XXXX` items. Screenshot from 2022-08-01 09-57-13

This has been fixed @kilemensi

apps/codeforafrica/.env
```diff
@@ -8,3 +8,5 @@ GHOST_URL="https://longform.codeforafrica.org"
 GITHUB_BACKEND_REPO="codeforafrica/ui"
 GITHUB_BACKEND_BRANCH="main"
 GITHUB_AUTH_ENDPOINT="api/oauth/auth"
+GIT_REPO_DIRECTORY=../../
```
kilemensi, line 11: Isn't this just equivalent to running the proxy server from root directory instead of app directory?

apps/codeforafrica/src/pages/api/admin/config.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
     repo: "CodeForAfrica/ui",
     proxy_url: "http://localhost:8081/api/v1", // Set proxy to work on local repo
   },
-  media_folder: "public/images",
+  media_folder: "apps/codeforafrica/public/images",
```
kilemensi, line 8: We can just declare `const APP_DIRECTORY` or env var `NEXT_PUBLIC_APP_DIRECTORY` (or both, i.e. declare `APP_DIRECTORY` initialized from `NEXT_PUBLIC_APP_DIRECTORY` env var) and reuse it rather than repeating `apps/codeforafrica`.

kilemensi (approved):
🚀

apps/codeforafrica/.env
esirK: I think it depends on how you run it for me I do
`pnpm --filter "./apps/codeforafrica" cms`
How do you run it from your side?

kilemensi: That's equivalent to `cd apps/codeforafrica && pnpm cms`. What I'm saying (or thinking out loud) is why not just `pnpm cms` at root? Does the proxy server care from which app you'll be connecting from?

esirK: Unless we make the install the server at the root folder. Right now only `codeforafrica` has the `cms`

esirK: alternatively, we can just remove it and move the `NEXT_PUBLIC_APP_DIRECTORY` to `.env.local` @kilemensi

kilemensi: Cool, then lets leave it in `codeforafrica` for now @esirK. We'll move install/move the proxy server to root when a 2nd app needs NetlifyCMS.

PR #198, opened 2022-08-04
Fix server error due to missing environment variable.

esirK:
## Description
Use `process.cwd()` instead of relying on the environment variable which might be unset

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/apps/codeforafrica/src/lib/api.ghost/posts.js
+++ b/apps/codeforafrica/src/lib/api.ghost/posts.js
@@ -1,12 +1,12 @@
 import { promises as fs } from "fs";
-import path from "path";
+import { join } from "path";
 
 import GhostContentAPI from "@tryghost/content-api";
 import camelcaseKeys from "camelcase-keys";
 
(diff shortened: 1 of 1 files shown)


esirK:
@kelvinkipruto  @kilemensi 
Seems like `process.env.PWD` is returning `undefined` in the production server which is causing the App to crash.

kilemensi:
> @kelvinkipruto @kilemensi Seems like `process.env.PWD` is returning `undefined` in the production server which is causing the App to crash.

Yes @esirK, I think it was a typo. I don't think there is such thing as `process.env.PWD`. It's always supposed to be [`process.cwd()`](https://github.com/CodeForAfrica/ui/blob/1d3023066efc13d3230a8d0741910a70f2413b27/apps/promisetracker/src/lib/cache/index.js#L10). @kelvinkipruto correct me if I'm wrong though.

kilemensi (approved):

PR #205, opened 2022-08-11
Project setup update

kilemensi:
## Description

 - [x] Upgrade all deps (except for React 18, waiting on Next.js / MUI to fully catch up),
 - [x] Use ARG only during builder step in Dockerfile, and
 - [x] Ensure turbo knows all env vars used so that it decide whether cache or full rebuild is required.

## Type of change

- [x] Chore

## Screenshots

N/A

## Checklist:


Dockerfile
```diff
@@ -43,10 +42,6 @@ COPY apps/${APP} ./apps/${APP}
 # Use virtual store: https://pnpm.io/cli/fetch#usage-scenario
 RUN pnpm install --recursive --offline --frozen-lockfile
 
-ENV NEXT_TELEMETRY_DISABLED=${NEXT_TELEMETRY_DISABLED} \
-    PROJECT_ROOT=${PROJECT_ROOT} \
-    GHOST_API_KEY=${GHOST_API_KEY}
```
esirK, line 48: I think `GHOST_API_KEY` is still needed during runtime and thus should be included in the `ENV`

kilemensi: I think `ARG` is enough in `builder`... `ENV` are required if you'll docker run the stage which we don't.

... but try building it locally and let me know. Background: https://www.saltycrane.com/blog/2021/04/buildtime-vs-runtime-environment-variables-nextjs-docker/

esirK: I think we had issues last time when we didn't have it as an `ENV` but let me try building it locally and see.

esirK: @kilemensi  for `ENV are required if you'll docker run the stage which we don't.`
Doesn't dokku do a `docker run` behind the scenes?
Also, I've built the Image and supplied  `GHOST_API_KEY` but if I run the image without the `GHOST_API_KEY` ENV, it fails.

kilemensi: When you (or **Dokku**) run the docker image, it actually runs the `runner` stage not the `builder` stage. All pages created during `builder` stage should be accessible even  during `runner` stage. You won't be able to fetch any updates however until you provide the `runner` stage with appropriate ENV vars.

**Dokku** supplies the `runner` stage with ENV vars from `dokku config:set`. Locally you'd need to supply the ENV vars either via `docker run --env-file` or `docker run --env`

This way the `secrets` are **NEVER** stored in the image.

esirK (approved):
Lgtm

kelvinkipruto (approved):

thepsalmist:
This looks good to me

thepsalmist (approved):

PR #216, opened 2022-08-16
Fix /about pages

kilemensi:
## Description

- [x] Ensure `/about/members` uses background.main background,
- [x] Ensure `TeamMemberCardList` stretches cards to have equal height,
- [x] Add `deactivated` prop on team members to allow hiding of a team member without having to delete them. This is because we don't allow any selection on on /about page,
- [x] Ensure `/about/partners` loads partners ([currently](https://cfa.dev.codeforafrica.org/about/partners) they're missing),
- [x] Ensure **Get Involved** loads content from CMS, and
- [x] Ensure all components in `/about` page have their content editable and loadable from CMS.

## Type of change

## Screenshots

Screenshot 2022-08-16 at 16-39-54 About Code for Africa

## Checklist:

### DIFF
--- a/apps/codeforafrica/src/components/GuidingPrinciplesCardList/GuidingPrinciplesCardList.js
+++ b/apps/codeforafrica/src/components/GuidingPrinciplesCardList/GuidingPrinciplesCardList.js
@@ -6,9 +6,9 @@ import GuidingPrinciplesCard from "../GuidingPrinciplesCard";
 
 const GuidingPrinciplesCardList = React.forwardRef(
   function GuidingPrinciplesCardList(props, ref) {
-    const { principles, title, ...other } = props;
+    const { list, title, ...other } = props;
 
-    if (!principles?.length) {
+    if (!list?.length) {
       return null;
     }
     return (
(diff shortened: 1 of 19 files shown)


apps/codeforafrica/src/components/GuidingPrinciplesCardList/GuidingPrinciplesCardList.js
```diff
@@ -6,9 +6,9 @@ import GuidingPrinciplesCard from "../GuidingPrinciplesCard";
 
 const GuidingPrinciplesCardList = React.forwardRef(
   function GuidingPrinciplesCardList(props, ref) {
-    const { principles, title, ...other } = props;
+    const { list, title, ...other } = props;
```
esirK, line 9: @kilemensi  isn't `principles` more descriptive? Also, the component already ends with `List`

kilemensi: It is but we're not consistent e.g. `OurImpact`, `OurPartners`, etc., seem to be using `title` and `list` while `Articles` seem to be using `articles`, etc. I've just adopted the `title` and `list` here since they'll be suitable in all situations and one doesn't need to remember the prop name.

... kinda like how we switched every html body to just content rather than keep mixing `content` and `description`.

esirK: I see I see. Makes sense.

esirK (approved):
LGTM

Just the version bump now.

Issue #229, opened 2022-08-22
Automatic package versioning


kilemensi:
Our current release process seem to be manual and random. As part of release discussion, we lean towards two approaches:

 - Release a new version on `every` merge to main, and
 - Release whenever there is something **significant** to release.

Having read upon this a bit further, our true crime is the manual and random part. Whether its on every merge on when there is something **significant** to release, doesn't really matter much in the end.

Having read on [Changesets](https://github.com/changesets/changesets) again (did a bit of it at the time of setting up this repo), it seems like it can help us automate and bring predictability to our releases. I'd urge everyone of us to at least read the [intro](https://github.com/changesets/changesets/blob/main/docs/intro-to-using-changesets.md) and [automating changesets](https://github.com/changesets/changesets/blob/main/docs/automating-changesets.md) if they don't have time to read through the full docs.

In a nutshell, we'll add a custom script to bump version of Dockerfile after changesets version has bumped package.json. Similar to [`changesets/action`](https://github.com/changesets/action/blob/main/scripts/bump.js) updates changelog. Changesets will then create a PR with these changes.

Let me know what you guys think @esirK, @kelvinkipruto, @thepsalmist

kilemensi:
Resolved by #232

PR #232, opened 2022-08-24
Chore use changeset

esirK:
## Description
Use changeset for releases
- Update both app package.json and the appropriate Dockerfile

Finishes https://github.com/CodeForAfrica/ui/issues/229

Example Changeset PR https://github.com/CodeForAfrica/ui/pull/231
## Type of change

## Checklist:


bump.js
```diff
@@ -0,0 +1,21 @@
+const fs = require("fs");
+const path = require("path");
+const { exec } = require("@actions/exec");
+
+(async () => {
+  await exec("pnpm changeset", ["status", "--output=./.changeset/status.json"]);
+  await exec("pnpm changeset", ["version"]);
```
kilemensi, line 7: Why do we run changeset `twice`?

bump.js
```diff
@@ -0,0 +1,21 @@
...
+  const changesetStatus = require("./.changeset/status.json");
+  changesetStatus?.releases.map((release) => {
+    const { name: appName } = release;
+    const newVersion = `${require(`./apps/${appName}/package.json`).version}`;
```
kilemensi, line 11: Don't we need a way of checking that we're dealing with an app and not a a package here? I'm assuming changeset will bump both packages and apps if it deems it necessary.

package.json
```diff
@@ -19,7 +19,8 @@
     "test": "turbo run jest playwright --parallel",
     "clean": "turbo run clean && rm -rf node_modules",
     "format-check": "prettier --check --ignore-path './.gitignore' \"**/*.+(json|md|yml)\" './'",
-    "format": "prettier --write --ignore-path './.gitignore' \"**/*.+(json|md|yml)\" './'"
+    "format": "prettier --write --ignore-path './.gitignore' \"**/*.+(json|md|yml)\" './'",
+    "bump": "node bump.js"
```
kilemensi, line 23: We should run `prettier` after bumping since we'll be creating/updating files.

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+        os: [ubuntu-latest]
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+
+      - uses: pnpm/action-setup@v2.0.1
+        with:
+          version: 7.3.0
```
kilemensi, line 24: Lets harmonize all these pnpm versions across these workflow files and Dockerfile to a single version. May be we should declare an org/repo env var for it?

kilemensi (commented):
👍🏽 Good stuff

-----

1. All scripts should live in `scripts` folder.
2. Lets use the `@changesets/changelog-github changelog` format.

bump.js
esirK: The first changeset just generates the `status.json` that tells us which apps **will be** versioned. The second changeset is what does the actual versioning.
Or did you mean why we can't combine the 2?

bump.js
esirK: This should happen locally when you are generating the changeset 
Screenshot 2022-08-24 at 14 45 16

bump.js
kilemensi: Combine the two if possible but should be bump be called **when** there is a changeset already?

esirK: What I think happens is 
Locally, when you think the changes for an app are ready to be released, you run `changeset` which allows you to select the `app` that should be released. This will create a `.changeset` file like [.changeset/light-lions-hug.md](https://github.com/CodeForAfrica/ui/blob/30f577bd11826eb5138a7d2a38e47862b3a34c35/.changeset/light-lions-hug.md)  where you fully describe that changes that this release brings.
You then review the file and if everything looks good you create a PR that gets merged into main.
On push to main, the `release` workflow gets executed which should create a Changeset release PR that gets automatically updated anytime code is merged into main. We are then responsible for determining when the PR should be merged and thus causing the release to be deployed.
So to answer the question, bump is called when there is a changeset already inside the `.changeset` but there can be multiple changesets(multiple apps being released) in the folder and that's what running `status` changeset will help with.

bump.js
esirK: I see.
Any ideas/pattern that we currently use  for packages?

bump.js
kilemensi: I see what you mean & it makes sense but:

If `changeset version` actually bumps the version in package.json **_and_** that's all we need, why not skip the whole status check? If for a given app there was no version change, there won't be any change to the corresponding `Dockerfile` i.e. we'll create the `Dockerfile` with the same version.

esirK: Actually this was the first approach I had used but the issue is we won't have a way of knowing which app was actually changed unless we start doing git checks inside the workflow

kilemensi: Why do we need to know this? The [`app-build.yml` will check](https://github.com/CodeForAfrica/ui/blob/b4e665ad6db1174c4446834762ec46f648bb39a3/.github/workflows/codeforafrica-build-deploy.yml#L11) if it's `package.json` has changed or not. Or are we dropping that approach?

My understanding is we need this script for one purpose only: to ensure for **all** apps, `contrib/dokku/Dockerfile` version matches the corresponding `package.json` version. The rest will be taken care of in the corresponding `app-build.yml`

esirK: With this, it means we will have multiple release workflows for each app; Which I think is alright.

.github/workflows/release.yml
esirK: I've updated the versions to `7.4.0`. For now, I don't think Github actions support org/repo env (Or I can't seem to find it) Only GitHub secrets which I don't think a version should be.

esirK:
> 2. Lets use the `@changesets/changelog-github changelog` format.

not sure I get what you mean here @kilemensi

kilemensi:
> 2\. hangelog-github changelog

https://github.com/changesets/changesets/blob/main/docs/config-file-options.md#changelog-false-or-a-path @esirK

kilemensi:
One question @esirK : Doesn't a single `pnpm changeset version` bump all selected apps/packages? My thinking is say I've fixed a bug in `@commons-ui/core` and I'd like to release a new version of the `@commons-ui/core` package itself and every app that depends on it.

Who does the release workflow work in this case?

esirK:
> One question @esirK : Doesn't a single `pnpm changeset version` bump **all** selected apps/packages? My thinking is say I've fixed a bug in `@commons-ui/core` and I'd like to release a new version of the `@commons-ui/core` package itself and every app that depends on it.
> 
> How does the release workflow work in this case?

The Immediate way I'm thinking of is you should select all apps depending on the package and for all of them do a minor release.

Screenshot 2022-08-25 at 12 06 56

Screenshot 2022-08-25 at 12 07 06

Screenshot 2022-08-25 at 12 07 14

Screenshot 2022-08-25 at 12 07 51


I'm not sure if this is the best approach but lemme do some digging.

kilemensi:
> 
> I'm not sure if this is the best approach but lemme do some digging.

I think this is the correct way @esirK as far as the selection of what to release goes. What I'm asking is how do the `app-release.yml` workflows handle this? If one app, say PromiseTracker calls `changeset version` in `bump`, I'm sure _changesets_ will update **all** versions. What will CodeForAfrica then see when it calls `bump`?

esirK:
> > I'm not sure if this is the best approach but lemme do some digging.
> 
> I think this is the correct way @esirK as far as the selection of what to release goes. What I'm asking is how do the `app-release.yml` workflows handle this? If one app, say PromiseTracker calls `changeset version` in `bump`, I'm sure _changesets_ will update **all** versions. What will CodeForAfrica then see when it calls `bump`?

I see; Having different releases might not work afterall. 
Lemme think about it but I think we might have to have 1 `release` that has multiple version checks for all packages/apps in the repo.

thepsalmist (approved):
Looks good to me @esirK

esirK:
I think it's good now @kilemensi

apps/promisetracker/contrib/dokku/Dockerfile
```diff
@@ -1 +1 @@
-FROM codeforafrica/promise-tracker-ui:0.2.0-alpha.6
+FROM codeforafrica/promise-tracker-ui:0.2.0-alpha.5-next
```
kilemensi, line 1: Why are we going backwards here?

package.json
```diff
@@ -28,5 +29,9 @@
     "prettier": "^2.7.1",
     "turbo": "^1.4.3"
   },
-  "packageManager": "pnpm@7.4.0"
+  "packageManager": "pnpm@7.4.0",
+  "dependencies": {
+    "@actions/exec": "^1.1.1",
+    "@changesets/changelog-github": "^0.4.6"
+  }
```
kilemensi, line 36: Aren't these `devDependencies`?

scripts/bump.js
```diff
@@ -0,0 +1,19 @@
+const fs = require("fs");
+const path = require("path");
+const { exec } = require("@actions/exec");
+
+(async () => {
+  await exec("pnpm changeset", ["version"]);
+  const apps = fs.readdirSync("apps");
```
kilemensi, line 7: The only thing I _may_ add here is to filter directories only (in case there is `apps/README.md` or anything other flat file.

scripts/bump.js
```diff
@@ -0,0 +1,19 @@
...
+  apps.map((app) => {
+    const newVersion = `${require(`../apps/${app}/package.json`).version}`;
+    const dockerFilePath = path.join(`apps/${app}/contrib/dokku`, "Dockerfile");
```
kilemensi, line 10:
```suggestion
    const dockerFilePath = path.join("apps", app, "contrib/dokku/Dockerfile");
```


Why use `path.join` if we're using `${}`? Lets pick one.

kilemensi (approved):
🚀 

---

Need to merge `main` here.

package.json
esirK: yeah yeah

apps/promisetracker/contrib/dokku/Dockerfile
esirK: I think `package.json` should be the main source of the version so that's what was picked up.

kilemensi: Then lets be careful; moving backwards will cause a clash in the future when we get to .6 again since the image with that version number will already be in docker hub.

PS: I think it will best to change versioning of PromiseTracker in this repo vs the one in the standalone repo just in case we need to make a quick fix in the standalone repo... May be we should go for `0.3.0-alpha.1` for this one since we know we **_won't_** continue feature development in the standalone repo.

PR #235, opened 2022-08-25
Pull legal pages content from CMS

kilemensi:
## Description

Remove hard-coded content for legal pages (/imprint & /privacy), replacing it with CMS content.

## Type of change

## Screenshots

N/A

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/apps/codeforafrica/src/lib/api.netlify-cms/getBody.js
+++ b/apps/codeforafrica/src/lib/api.netlify-cms/getBody.js
@@ -2,7 +2,7 @@ import { getCollectionBySlug } from "./utils";
 
 function getBody(page) {
   const { content } = getCollectionBySlug("content/pages", page, [
-    ["content"],
+    "content",
   ]).items;
 
   return { content };
--- a/apps/codeforafrica/src/lib/api.netlify-cms/getBody.js
+++ b/apps/codeforafrica/src/lib/api.netlify-cms/getBody.js
@@ -1,0 +1,11 @@
+import { getCollectionBySlug } from "./utils";
+
+function getBody(page) {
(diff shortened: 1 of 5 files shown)


apps/codeforafrica/src/lib/api.netlify-cms/getBody.js
```diff
@@ -0,0 +1,11 @@
+import { getCollectionBySlug } from "./utils";
+
+function getBody(page) {
+  const { content } = getCollectionBySlug("content/pages", page, [
+    ["content"],
```
esirK, line 5: looks good but why are you passing 2 arrays to fields?

esirK (approved):
🚀
Shouldn't the `getBody` also be inside the `sections` folder?

kilemensi:
> rocket Shouldn't the `getBody` also be inside the `sections` folder?

Don't start me again with that sections folder business @esirK . I still don't know why as an API user I need to import from `api.netlify.cms/sections` instead of just a`pi.netlify.cms`. I feel we're just exposing implementation detail rather than something useful to the user.

Breath, breath, breath.

esirK:
> Don't start me again with that sections folder business @esirK . I still don't know why as an API user I need to import from `api.netlify.cms/sections` instead of just `api.netlify.cms`. I feel we're just exposing an **implementation detail** rather than something useful to the user.
> 
> Breath, breath, breath.

I hear you @kilemensi 
But I think that's why we added the `index.js` to prevent exposing the **implementation detail**? But let's leave the 🤼 for another day.

PR #241, opened 2022-08-31
Fix RichTypography

kilemensi:
## Description

We currently have `RichTypography` features spread all over the site. This PR moves everything back to the `RichTypography` in packages.

 - [x] Moves `color` handling back to `@common-ui/core/RichTypography` via `LinkProps.color` prop,
 - [x] Moves `onclick` attachment to `@common-ui/core/RichTypography` if `LinkProps.onClick` prop is provided, and
 - [x] Introduces `@commons-ui/next/RichTypography` which integrates `RichTypography` with `next/router` for opening relative URLs but retaining the ability to open external links on new tab.

## Type of change

## Screenshots

S

## Checklist:


kelvinkipruto (approved):
Looks Good

packages/commons-ui-next/src/utils/isExternalUrl.js
```diff
@@ -0,0 +1,10 @@
+/* eslint-disable import/prefer-default-export */
+
+function isExternalUrl(url) {
+  return (
+    typeof url === "string" &&
+    (url.indexOf("http") === 0 || url.indexOf("mailto:") === 0)
```
esirK, line 6: Just finding something to comment on. Maybe we should support even `tel`?

esirK (approved):
LGTM

packages/commons-ui-core/src/RichTypography/RichTypography.js
```diff
@@ -1,39 +1,61 @@
...
+        for (let i = 0; i < anchors.length; i += 1) {
+          anchors[i].onclick = LinkProps.onClick;
+        }
+      }
+    }
+  }, [LinkProps?.onClick]);
+  let children;
+  let dangerouslySetInnerHTML;
+  if (typeof childrenProp === "string") {
+    dangerouslySetInnerHTML = {
```
esirK, line 37: With this, we assume that we trust anyone adding content to the CMS right? Tested this by adding something like 
`` which gets executed in the F.E

kilemensi: Yes and no.

Since it's behind GitHub login, we can say we trust the people adding content. But really, we shouldn't.

But, since you're the one who chose `markedjs` over `Remark`, you should have added `DOMPurify` per their [docs](https://marked.js.org/#usage).

packages/commons-ui-next/src/utils/isExternalUrl.js
kilemensi: I think we only added mailto to support contact us via email which we don't even use (contact form being a Mailchimp embed).

packages/commons-ui-core/src/RichTypography/RichTypography.js
esirK: yeah yeah. Should it be added to the current PR? 😂

kilemensi: I take it we'll be getting a PR adding DOMPurify from you @esirK ?

PR #244, opened 2022-09-02
Chore add dompurify

esirK:
## Description

Adds dompurify for sanitizing HTML output as [recommended](https://marked.js.org/#usage) by marked 

## Type of change

## Screenshots

How `` gets sanitized

Screenshot 2022-09-02 at 12 12 57

## Checklist:


kilemensi (commented):
👍🏽 

-----

Looks good but just thinking out loud for a better approach: Shouldn't we introduce `lib/marked` (since it's server-side only) that exposes same functional interface to `marked` but we call `DOMPurify` on returned results? This way, it'll be impossible to forget to sanitize HTML output

```js
import { marked as realMarked } from './Tokenizer.js';
import { JSDOM } from 'jsdom';
import DOMPurify from 'dompurify';

const window = new JSDOM('').window;
const purify = DOMPurify(window);

function marked(src, opt, callback) {
  return purify.sanitize(realMarked(src, opt, callback));
}

marked.parseInline = function(src, opt) {
 return purify.sanitize(realMarked.parseInline(src, opt));
}

export default marked;
```

The only change we make in the rest of the code is just changing `marked` import to `lib/marked`.

esirK:
> 👍🏽
> 
> Looks good but just thinking out loud for a better approach: Shouldn't we introduce `lib/marked` (since it's server-side only) that exposes same functional interface to `marked` but we call `DOMPurify` on returned results? This way, it'll be impossible to forget to sanitize HTML output
> 
> ```js
> import { marked as realMarked } from './Tokenizer.js';
> import { JSDOM } from 'jsdom';
> import DOMPurify from 'dompurify';
> 
> const window = new JSDOM('').window;
> const purify = DOMPurify(window);
> 
> function marked(src, opt, callback) {
>   return purify.sanitize(realMarked(src, opt, callback));
> }
> 
> marked.parseInline = function(src, opt) {
>  return purify.sanitize(realMarked.parseInline(src, opt));
> }
> 
> export default marked;
> ```
> 
> The only change we make to the rest of the code is just changing `marked` import to `lib/marked`.

Yeah yeah. I think this is a better approach.

apps/codeforafrica/jest.setup.js
```diff
@@ -1 +1,9 @@
 module.exports = require("@commons-ui/testing-library/jest.setup");
+
```
kilemensi, line 2: Having gone through the code and the problem, I kinda of agree with this [comment](https://github.com/jsdom/jsdom/issues/2524#issuecomment-902027138): The issue seem to be we're creating JSDOM inside Jest's jsdom environment and we should try avoid it. My quick suggestion that will make this setup unnecessary is in `lib/marked` file comment.

apps/codeforafrica/src/lib/marked.js
```diff
@@ -0,0 +1,16 @@
+import DOMPurify from "dompurify";
+import { JSDOM } from "jsdom";
+import { marked as realMarked } from "marked";
+
+const { window } = new JSDOM("");
+const purify = DOMPurify(window);
```
kilemensi, line 6:
```suggestion
/* eslint-env browser */

import DOMPurify from "dompurify";
import { marked as realMarked } from "marked";

let global;
if (typeof window !== "undefined") {
  global = window;
} else {
  // lib/marked will only be used server-side hence require is fine
  // eslint-disable-next-line global-require
  const { JSDOM } = require("jsdom");
  global = new JSDOM("").window;
}
const purify = DOMPurify(global)
```

We can switch that `require` to `import` but support for it is still experimental in [Webpack 5](https://stackoverflow.com/questions/68339243/how-can-i-use-top-level-await-in-typescript-next-js).

apps/codeforafrica/src/lib/marked.js
```diff
@@ -0,0 +1,16 @@
...
+
+function marked(src, opt, callback) {
+  return purify.sanitize(realMarked(src, opt, callback));
+}
+
+marked.parseInline = function (src, opt) {
```
kilemensi, line 12:
```suggestion
marked.parseInline = function parseInline(src, opt) {
```

kilemensi (commented):
👍🏽

apps/codeforafrica/src/lib/marked.js
```diff
@@ -0,0 +1,24 @@
...
+if (typeof window !== "undefined") {
+  global = window;
+} else {
+  // lib/marked will only be used server-side hence require is fine
+  // eslint-disable-next-line global-require
+  const { JSDOM } = require("jsdom");
+  global = new JSDOM("").window;
+}
+
+const purify = DOMPurify(global);
```
kilemensi, line 15:
```suggestion
const purify = DOMPurify(global);

```

kilemensi (approved):
🚀

apps/codeforafrica/src/lib/marked.js
esirK: Ola @kilemensi  just curious, other than`import` being one of the new ways to perform imports, what other benefits will we get from using the experimental support?

Also, I'm not sure if the `import` is supported even on the experimental (Still can't use it locally despite trying all the experimental defined [here](https://webpack.js.org/configuration/experiments/#experiments))

kilemensi: > what other benefits will we get from using the experimental support?
>

For us, nothing I can think of and [looking around](https://stackoverflow.com/questions/46677752/the-difference-between-requirex-and-import-x), I don't see much... but being a standard is a big deal and if possible we should try to stick to it @esirK.

Enabling it in `next.config.js` should work fine when running the app... Haven't tried looking at how to enable it under Jess given Jess doesn't use `next.config.js`

esirK: @kilemensi  I've tried different options for this including suggestions e.g [this](https://jestjs.io/docs/ecmascript-modules), and suggestions [here](https://stackoverflow.com/questions/70725063/jest-wont-accept-top-level-awaits-with-nodejs16-typescript) but I don't seem to be making any progress. Any other pointers I can look at?

kilemensi: Since we're already using Next.js 12 @esirK , I think our best bet is to switch `jest-config-commons-ui/next.js` package to use Next.js's built-in [configuration](https://nextjs.org/docs/testing#setting-up-jest-with-the-rust-compiler). It should load `next.config.js` even under Jest.

PR #246, opened 2022-09-02
ft/Removed hard coded content

### LINKED ISSUE #242: Review lib/index.js for any remaining hard-coded strings
Now we have full integration with CMSes, it's time to review and remove all remaining hard-coded strings in code. Specifically, we should comb through `lib/index.js`.

kelvinkipruto:
## Description

This PR intends to remove hard-coded content and move it to CMS

Fixes #242 

## Type of change

## Checklist:


apps/codeforafrica/src/lib/index.js
```diff
@@ -689,17 +691,12 @@ function getContactPageStaticProps() {
...
         },
         {
           slug: "join-our-slack",
-          title: "We are on Slack!",
-          subtitle: "Join us",
-          action: {
-            label: "Join our Slack",
-            href: "https://docs.google.com/forms/d/e/1FAIpQLSdkfLU2yi2S1_7D27Z0I1TumkWy5brlam809Od9cc6CnXGA-A/viewform",
-          },
+          ...joinUs,
```
kilemensi, line 699: Looking at this, join us and contact form are two separate sections/components. Why are we combining them in the CMS?

kelvinkipruto: Initially, they were bundled together in the Contact form, so I just simplified it. Let me separate them.

apps/codeforafrica/src/lib/api.netlify-cms/getGetInTouch.js
```diff
@@ -1,7 +1,9 @@
...
 
 export default function getGetInTouch() {
-  return getCollectionBySlug("content/pages", "about", ["get-in-touch"]).items[
-    "get-in-touch"
-  ];
+  const getInTouch = getCollectionBySlug("content/pages", "about", [
+    "get-in-touch",
+  ]).items["get-in-touch"];
+
+  return { ...getInTouch, slug: "get-in-touch" };
```
kilemensi, line 8: No need to do it this way. `getCollectionBySlug` already [supports](https://github.com/CodeForAfrica/ui/blob/ed09ce7b7ab86baa7a12b8b427d794cb6d188ccc/apps/codeforafrica/src/lib/api.netlify-cms/utils.js#L25-L26) `slug` as one of the fields

apps/codeforafrica/src/pages/api/admin/config.js
```diff
@@ -603,6 +603,17 @@ module.exports = {
...
+              ...seoFields,
+            },
+          ],
+        },
+        {
+          label: "Contact Us | Join Us",
+          name: "contact-us-join-us",
+          file: `${APP_DIRECTORY}content/pages/contact-us-join-us.md`,
+          widget: "object",
+          fields: [
```
kilemensi, line 616: Why a new page? Can't we add this `join-us` field to the _existing_ `contact-us` page?

kilemensi (commented):
I still don't _grok_ your approach @kelvinkipruto ... but, my current thinking is:

1. `slug` is just another field that can be returned from `getCollectionBySlug`. No need to do anything special to handle it.
2. If we need to add a new component, we'll add it to existing page(s).

If there is a situation that necessitates straying away from the above path, let us know.

apps/codeforafrica/src/pages/api/admin/config.js
kelvinkipruto: You [suggested](https://github.com/CodeForAfrica/ui/pull/246#discussion_r961635744) we should separate them since they are in different components.
> Looking at this, join us and contact form are two separate sections/components. Why are we combining them in the CMS?

kelvinkipruto:
@kilemensi As mentioned in the call, I don't think it will be possible to remove all hard coded content. The above solution works for situations where the props does not require a key like `getHero` [here](https://github.com/CodeForAfrica/ui/blob/234308d8667806576aec72b1dc26fa71e6b487bb/apps/codeforafrica/src/lib/index.js#L169-L171) or `getOurImpact` [here](https://github.com/CodeForAfrica/ui/blob/234308d8667806576aec72b1dc26fa71e6b487bb/apps/codeforafrica/src/lib/index.js#L185-L187). When props property keys are required like  `partners: getOurPartners()`  or [here](https://github.com/CodeForAfrica/ui/blob/234308d8667806576aec72b1dc26fa71e6b487bb/apps/codeforafrica/src/lib/index.js#L334-L350) for example, we'll need to hard-code the slugs and titles.

apps/codeforafrica/src/pages/api/admin/config.js
kilemensi: Yes @kelvinkipruto they **should be** separated because they're different _components_. You're separating them as if they're different _pages_.

Components in a page (Next.js) are just fields in a collection (NetlifyCMS).

kilemensi:
> @kilemensi As mentioned in the call, I don't th...

I don't see what you see @kelvinkipruto... but let me pull the code & check.

apps/codeforafrica/src/pages/api/admin/config.js
kelvinkipruto: Ok. Now I get it  👨🏿‍🔧

kelvinkipruto:
@kilemensi Wow  🤯🤩. Thanks. Let me go through it and see how I should have done it.

kilemensi (approved):
🚀 

---

Create and include changeset @kelvinkipruto

kilemensi:
Where is the changeset @kelvinkipruto ?

>🚀 
>
>---
>
>Create and include changeset @kelvinkipruto

kelvinkipruto:
ooh, no, Let me try fix it.

PR #268, opened 2022-11-10
@charterafrica/input

### LINKED ISSUE #257: @/charterafrica/Input
Implement `Input` component per designs.

Image

Please refer to [`@/codeforafrica/Input`](https://github.com/CodeForAfrica/ui/tree/main/apps/codeforafrica/src/components/Input) component as reference. If reusable, move `@/codeforafrica/Input` to `@commons-ui/core/Input` and reuse it in both apps.

kelvinkipruto:
## Description

This PR implements input as per the designs

Fixes #257 

## Type of change

## Screenshots
input

## Checklist:


apps/codeforafrica/src/components/SearchInput/SearchInput.js
```diff
@@ -1,9 +1,9 @@
+import Input from "@/commons-ui/core/Input";
```
kilemensi, line 1: Should be `@commons-ui/core` and not `@/commons-ui/core`

kilemensi (commented):
👍🏽 

---

Are we actually reusing anything in the Input component? Otherwise, there is no benefit of moving Input to @commons-ui/core.

apps/charterafrica/src/components/SearchInput/SearchInput.js
```diff
@@ -0,0 +1,61 @@
...
+        backgroundColor: neutral[800],
+        borderColor: neutral[400],
+        border: "1px solid",
+        borderRadius: "4px",
+        input: {
+          "&::placeholder": {
+            color: neutral[400],
+          },
+        },
+      }}
```
kilemensi, line 53: Similar to how we customized `Button` by customizing [`theme`](https://github.com/CodeForAfrica/ui/blob/5f3583f4067ef776d30230cf2d1861ea80cfe8c6/apps/charterafrica/src/theme/index.js#L157-L197), can't we do all `Textfield` [customization](https://www.figma.com/file/gbnaO1FYPq16V2gAFbLOkl/CfA---EPD-Charter-Africa---Sep-2022?node-id=5138%3A3775&t=GfFlfQF9yJpnE8Rt-0) in the theme as well?

kilemensi (commented):
👍🏽

apps/charterafrica/src/theme/index.js
```diff
@@ -195,6 +195,45 @@ deepmerge(
...
+        root: {
+          ...theme.typography.h6Small,
+          border: "1px solid",
+          borderRadius: "4px",
+          borderColor: neutral[400],
+        },
+        sizeSmall: {
+          height: "36px",
+        },
+        sizeMedium: {
```
kelvinkipruto, line 209: @kilemensi Not sure if this should be here as they do not appear on the [input api](https://mui.com/material-ui/api/input/#css), only `sizeSmall` is present. If not what would be the approach to style medium and large inputs in the theme?

apps/charterafrica/src/components/SearchInput/SearchInput.js
kelvinkipruto: @kilemensi I have given this a try, and got some of it working. What do you think about it? Is it the right approach of do I have to tweak something?

apps/charterafrica/src/theme/index.js
kilemensi: _Heita_ @kelvinkipruto 

1. I think we should focus on [`MuiOutlinedInput`](https://mui.com/material-ui/api/outlined-input/) first since that's the `TextField` variant that closely resembles the designed `Input` component. Only switch `MuiInputBase` for props/styles directly inherited from it.
2. From the API [docs](https://mui.com/material-ui/api/text-field/#props), `TextField` supports `medium` and `small` sizes. This means `large` is the **_only_** one that doesn't have direct support. Nevertheless, can you try setting size to `large` (since they say the `size` prop supports any `string`) and see what class name does MUI generate? 🤞🏽 it will be `sizeLarge`.

apps/charterafrica/src/components/SearchInput/SearchInput.js
kilemensi: Yep, I think we're on the right track @kelvinkipruto . See my comment [here](https://github.com/CodeForAfrica/ui/pull/268#discussion_r1031408726).

kelvinkipruto:
@kilemensi This is ready for review

kilemensi:
> @kilemensi This is ready for review

Seen it. Waiting for 🔌 to test it locally @kelvinkipruto

kilemensi (approved):
🚀

PR #274, opened 2022-11-18
CI/CD

thepsalmist:
## Description

CI/CD workflows for repo.

1. **DEV**: Every time a push is done to `main` (there is a path filter to make sure the changes were indeed made in an app). Currently only `@/charterafrica` is supported.
2. **PROD**: The previous workflows where we only deploy if there is a version bump. `@/charterafrica` and `@/codeforafrica` supported.

## Type of change

- [x] Chore

## Screenshots

N/A

## Checklist:


Dockerfile
```diff
@@ -68,7 +68,7 @@ RUN adduser --system --uid 1001 nextjs
 COPY --from=builder --chown=nextjs:nodejs /workspace/apps/${APP}/.next/standalone ./
 
 # Because of PROJECT_ROOT in builder, standalone output of ${APP} will be in
-# in /workspace/apps/${APP} after the above COPY.
+# in /workspace/apps/${APP} after the above COPY.docker
```
kilemensi, line 71: ?

apps/charterafrica/.env
```diff
@@ -1,7 +0,0 @@
-NEXT_PUBLIC_APP_LOGO_URL=https://charter.dev.codeforafrica.org/logos/main.png
```
kilemensi, line 1: [In Next.js world](https://nextjs.org/docs/basic-features/environment-variables), `.env` file should be committed to git containing public vars/non-secrets. They encourage the use of `.env.local` file for secrets and this is the file that should **never** be committed to git.

kilemensi (commented):
👍🏽

kilemensi:
Let me know how it looks @thepsalmist

.github/workflows/charterafrica-build-deploy.yml
```diff
@@ -7,13 +7,16 @@ env:
   DOKKU_REMOTE_BRANCH: "master"
   DOKKU_REMOTE_URL: "ssh://dokku@ui-1.prod.codeforafrica.org/charterafrica"
```
thepsalmist, line 8: @kilemensi I believe we can set the hostname to be consistent with what we have on Cloudflare, so this would be `charterafrica.prod`

kilemensi: This workflow is to deploy to **dev** @thepsalmist. Still thinking about **prod** deployments

thepsalmist: Cool @kilemensi  LGTM

kilemensi: Any ideas on how we should handle deployments to **dev** vs **prod** @thepsalmist ?

kilemensi: Cool @thepsalmist ... For the now, can you check this setup for CI/CD

1. **DEV**: Every time a push is done to main (there is a path filter to make sure the changes were indeed made in an app).
2. **PROD**: The previous workflows where we only deploy if there is a version bump.

thepsalmist: Makes sense @kilemensi, although if i remember we introduced changesets to manage the version bump

kilemensi: Yes, that hasn’t changed @thepsalmist: Youl’ll still use changesets to bump version in package.json and release workflow will bump the one in contrib/dokku and then a prod deployment will be made once the bump PR is merged to main.

kilemensi (approved):
🚀 

---


If you don't find any issues @thepsalmist, lets merge this.

PR #295, opened 2023-01-17
Feat/charterafrica cms democracy helpdesk 284

### LINKED ISSUE #284: @/charterafrica/cms/DemocracyHelpdesk Global
Using any of the [Fields](https://payloadcms.com/docs/fields/overview), create DemocracyHelpdesk support in the **globals** section of the CMS i.e. same date will be shared between multiple pages. Replace the hard-coded helpdesk content on homepage with the CMS data.

kilemensi:
## Description

Fixes #284 .

## Type of change

## Checklist:

### DIFF
--- a/apps/charterafrica/payload.config.js
+++ b/apps/charterafrica/payload.config.js
@@ -6,6 +6,7 @@ import { buildConfig } from "payload/config";
 import Media from "./src/payload/collections/Media";
 import Pages from "./src/payload/collections/Pages";
 import FocalCountries from "./src/payload/globals/FocalCountries";
+import Helpdesk from "./src/payload/globals/Helpdesk";
 import Navigation from "./src/payload/globals/Navigation";
 import Settings from "./src/payload/globals/Settings";
 import { defaultLocale, locales } from "./src/payload/utils/locales";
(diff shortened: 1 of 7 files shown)


apps/charterafrica/payload.config.js
```diff
@@ -29,7 +30,7 @@ const adapter = s3Adapter({
 export default buildConfig({
   serverURL: appURL,
   collections: [Media, Pages],
-  globals: [FocalCountries, Navigation, Settings],
+  globals: [FocalCountries, Helpdesk, Navigation, Settings],
```
kelvinkipruto, line 33: In my opinion, I think this could be a `block` instead of a `global`

kelvinkipruto (approved):

apps/charterafrica/payload.config.js
kilemensi: And how would you have the exact content on multiple pages @kelvinkipruto? The same component is in at least 2 pages: homepage and community page.

kelvinkipruto: That is ok if the content should be the same on multiple pages. I though each page could have its own content. @kilemensi

PR #314, opened 2023-02-01
Fix CI/CD Pipeline

kilemensi:
## Description

Some needed env variables were not defined at build time resulting in some steps being totally skipped e.g. Next.js build was completely not running i.e. .next folder was completely missing in the resulting image.

This PR ensures all build-time required env vars are declared as ARG and given default values where possible.

## Type of change

- [x] Chore

## Screenshots

N/A

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas

PR #315, opened 2023-02-01
Add missing MONGO_URL env var

kilemensi:
## Description

This PR adds `MONGO_URL` env var missed by #314.

## Type of change

- [x] Chore

## Screenshots

N/A

## Checklist:

### DIFF
--- a/.github/workflows/charterafrica-deploy-dev.yml
+++ b/.github/workflows/charterafrica-deploy-dev.yml
@@ -50,6 +50,7 @@ jobs:
         uses: docker/build-push-action@v3
         with:
           build-args: |
+            MONGO_URL=${{ secrets.CHARTERAFRICA_MONGO_URL }}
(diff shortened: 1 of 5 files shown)


thepsalmist (commented):
I think for dev the issue could be the default deploy branch as per [this](https://github.com/marketplace/actions/dokku) The git deployment defaults to master, we're currently on main.

otherwise GTM @kilemensi

kilemensi:
> I think for dev the i...
> 
> otherwise GTM @kilemensi

We're no longer pushing source code to Dokku @thepsalmist. We're asking Dokku to pull an image: https://github.com/CodeForAfrica/ui/blob/ea356a4ae2564968de8d6bef075c348946a5f024/.github/workflows/charterafrica-deploy-dev.yml#L75

This completely [initializes](https://dokku.com/docs/deployment/methods/git/#initializing-an-app-repository-from-a-docker-image) the app so branches shouldn't matter.

... but, if this PR also fails, then yes, we'll need to look into other possible causes.

PR #326, opened 2023-02-07
@/charterafrica/pages/ErrorPage #319

### LINKED ISSUE #319: @/charterafrica/pages/ErrorPage
## Description

Implement the error pages per [designs](https://www.figma.com/file/gbnaO1FYPq16V2gAFbLOkl/CfA---EPD-Charter-Africa---Sep-2022?node-id=9313%3A29587&t=gHql57tAK6HPaLGe-0). Things to keep in mind:

  1. Content of these pages should be editable in the CMS to allow localisation.
  2. 404 [needs to be static](https://nextjs.org/docs/advanced-features/custom-error-page#404-page). Current config seem to have issue with static building (could be because we run Payload and Next.js in the same server) & hence we may need a temporary solution e.g. hard-code 404 content to the page.
  3. BackgrondBox component can be reused here by customising `bgcolor` and `backgroundImage`.

koechkevin:
## Description

Fixes #319 

## Type of change

## Screenshots
image

## Checklist:


apps/charterafrica/src/components/ErrorPage/ErrorPage.js
```diff
@@ -0,0 +1,92 @@
...
+    title = "",
+    createdAt,
+    updatedAt,
+    action,
+    ...other
+  } = props;
+
+  const link =
+    action &&
+    mapLinkTypeToHref({
```
kilemensi, line 26: Can't think of a single reason why we would need this **_on_** front-end.

kilemensi (commented):
👍🏽 

-----

Quick review; I see we're using `getStaticProps` in error pages but as I mentioned in the [issue](https://github.com/CodeForAfrica/ui/issues/319), this doesn't work. Have you tried running `pnpm build`?

apps/charterafrica/src/components/ErrorPage/ErrorPage.js
koechkevin: Could we be forced to use REST API instead of payload import?🤔.

Because footer and header needs data from payload.

kilemensi: We may have to but I'm hoping there is a fix for our current situation @koechkevin ... local APIs are just too nice.

kilemensi:
My two cents on the approach @koechkevin: Not sure making Errors a collection is a good idea or at least I don't see any benefit.

Error pages are just pages like any other page. Error details can be a Block that is added to a page. This way, 404 on front-end would then just fetch a page `slug=404` from CMS and that will be that.

kilemensi:
Making progress @koechkevin ?

apps/charterafrica/.env.template
```diff
@@ -20,6 +20,8 @@ NEXT_PUBLIC_APP_URL
 NEXT_PUBLIC_DEFAULT_LOCALE
 NEXT_PUBLIC_IMAGE_DOMAINS
 NEXT_PUBLIC_IMAGE_UNOPTIMIZED
+NEXT_PUBLIC_APP_AUTH_EMAIL
+NEXT_PUBLIC_APP_AUTH_PASSWORD
```
kilemensi, line 24: Do we need any auth? I believe all collections allow anyone to read e.g. for Pages: https://github.com/CodeForAfrica/ui/blob/1462db9951eb2fd1dc164dcc83753a212b21facc/apps/charterafrica/src/payload/collections/Pages.js#L19-L21

apps/charterafrica/package.json
```diff
@@ -61,6 +61,7 @@
     "nodemon": "^2.0.20",
     "payload": "^1.6.4",
     "prop-types": "^15.8.1",
+    "qs": "^6.11.0",
```
kilemensi, line 64: Do we need more than what [URLSearchParams](https://nodejs.org/docs/latest-v16.x/api/url.html#class-urlsearchparams) provides?

apps/charterafrica/src/components/ErrorPage/ErrorPage.js
```diff
@@ -0,0 +1,84 @@
...
+    title = "",
+    createdAt,
+    updatedAt,
+    blockType,
+    action,
+    ...other
+  } = props;
+  return (
+    <BackgroundBox
+      {...other}
```
kilemensi, line 25:
```suggestion
```

Avoid this. For example, if you've received props with other params like id, etc., they'll all end up added to DOM.

apps/charterafrica/src/components/ErrorPage/ErrorPage.js
```diff
@@ -0,0 +1,84 @@
...
+import { Link } from "@commons-ui/next";
+import { Button } from "@mui/material";
+import PropTypes from "prop-types";
+import React from "react";
+
+import { secondary, neutral } from "@/charterafrica/colors";
+import BackgroundBox from "@/charterafrica/components/BackgroundBox";
+import RichText from "@/charterafrica/components/RichText";
+
+const ErrorPage = React.forwardRef(function Error(props, ref) {
```
kilemensi, line 12: This doesn't look like a Page to be... It should probably just be named `Error` i.e. just another component.

apps/charterafrica/src/components/ErrorPage/ErrorPage.js
```diff
@@ -0,0 +1,84 @@
...
+        height: "100%",
+        width: "100%",
+        position: { md: "absolute", xs: "relative" },
+        backgroundSize: "50% 100%",
+        ...props?.sx,
+      }}
+      ref={ref}
+    >
+      <Section sx={{ p: 10 }}>
+        <RichTypography color={neutral[900]} textAlign="center" variant="h1">
```
kilemensi, line 42:
```suggestion
        <RichTypography color="neutral.dark" textAlign="center" variant="h1">
```

apps/charterafrica/src/components/ErrorPage/ErrorPage.js
```diff
@@ -0,0 +1,84 @@
...
+          {statusCode}
+        </RichTypography>
+        <RichTypography
+          sx={{ mt: 2.5, textTransform: "uppercase" }}
```
kilemensi, line 46:
```suggestion
          mt={2.5}
          textTransform="uppercase"
```

Typography component support these props directly.

koechkevin:
@kilemensi

apps/charterafrica/src/components/ErrorPage/ErrorPage.js
```diff
@@ -0,0 +1,84 @@
...
+          variant="p3"
+          elements={description}
+        />
+        <RichTypography component="div" textAlign="center">
+          <Link underline="none" href={action.href || ""}>
+            <Button sx={{ mt: 5 }} color="secondary" variant="contained">
+              {action?.title}
+            </Button>
+          </Link>
+        </RichTypography>
```
kilemensi, line 66: Just ```Button component={Link} sx={{ typography: xxxx }}``` should achieve what you want here.

apps/charterafrica/src/components/ErrorPage/ErrorPage.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+import ErrorPage from "./ErrorPage";
+
+import theme from "@/charterafrica/theme";
+
+// eslint-disable-next-line testing-library/render-result-naming-convention
+const render = createRender({ theme });
+
+const defaultProps = {
+  title: "Explainers",
```
kilemensi, line 12: ?

apps/charterafrica/src/components/Layout/Layout.js
```diff
@@ -8,6 +8,7 @@ import NavBar from "@/charterafrica/components/NavBar";
 
 const Main = styled("main")({
   flex: 1,
+  position: "relative",
```
kilemensi, line 11: Still not sure why we need to switch this.

apps/charterafrica/src/lib/data/fetchApi.js
```diff
@@ -0,0 +1,68 @@
+import qs from "qs";
```
kilemensi, line 1: Do we need any of this **_now_**?

apps/charterafrica/src/lib/data/fetchApi.js
```diff
@@ -0,0 +1,68 @@
...
+      this.headers = headers;
+      return;
+    }
+    this.authenticate();
+  }
+
+  async authenticate() {
+    const data = { email: this.email, password: this.password };
+    const res = await this.post(
+      `${process.env.NEXT_PUBLIC_APP_URL}/api/users/login`,
```
kilemensi, line 25: While they'll point to the same thing, any call meant to go to the Payload CMS should use `PAYLOAD_PUBLIC_APP_URL`.

apps/charterafrica/src/lib/data/fetchData.js
```diff
@@ -0,0 +1,70 @@
...
+export const fetchSettings = async (args) => {
+  return fetchGlobals("settings", args);
+};
+
+export const fetchErrorPages = async (args) => {
+  return fetchCollection("errors", args);
+};
+
+export const fetchGlobalProps = async (args) => {
+  const settings = await fetchSettings(args);
```
kilemensi, line 54:
```suggestion
  const settings = await fetchGlobals("settings, args);
```

I don't think we need any of the fetchXXXSpecificXXX. The above is simple enough for globals.

apps/charterafrica/src/pages/404.page.js
```diff
@@ -0,0 +1,33 @@
...
+  const { footer, navbar } = await fetchGlobalProps(args);
+
+  const { docs } = await fetchPage("404");
+  const [{ blocks: pages, slug }] = docs;
+  const setData = pages.find(({ statusCode }) => statusCode === 404);
+  const notFoundBlock = {
+    slug,
+    action: {},
+    ...setData,
+  };
```
kilemensi, line 29: 1. Check if page exist, through an error if it doesn't
2. Fetch globals
3. Profit.

apps/charterafrica/src/payload/blocks/Errors.js
```diff
@@ -0,0 +1,111 @@
...
+      doc: {
+        ...siblingData.doc,
+        relationTo: "pages",
+        breadcrumbs: siblingData.doc.breadcrumbs || [],
+      },
+      linkType: "internal",
+    },
+    req,
+  });
+};
```
kilemensi, line 17:
```suggestion
```

You don't need this.

kilemensi (commented):
👍🏽

apps/charterafrica/.env.template
koechkevin: On the API, I get this error without any authentication

image

kilemensi: I ran the PR locally without providing my email/password and it worked. May be it was that hard-coded json file. Let me try again @koechkevin

kilemensi: Just tried again and I can pull the content without auth info @koechkevin

apps/charterafrica/src/lib/payload/staticData.js
```diff
@@ -0,0 +1,73 @@
+const notFoundData = {
```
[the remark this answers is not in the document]
koechkevin: Yes we do not.

apps/charterafrica/src/components/Layout/Layout.js
koechkevin: Because I need to use absolute positioning somewhere in the child.

kilemensi: ... that's kind of the question @koechkevin . We only need to set one background image, do we really need absolute positioning? Isn't `background-position` enough?

apps/charterafrica/src/lib/data/fetchApi.js
koechkevin: We need a util function that converts nested objects searchParams.

`qs` does that.

Or we have to write a util function to handle that.

koechkevin: @kilemensi

kilemensi: Not sure I follow @koechkevin: How do nested search params look like in a URL?

koechkevin: We need to send a query with params object that looks like this 
```js
{
  locale: 'en',
  'fallback-locale': 'en',
  locales: [ 'en', 'fr', 'pt' ],
  where: { slug: { equals: '404' } }
}
```
We then expect a url like `?locale=en&fallback-locale=en&locales%5B0%5D=en&locales%5B1%5D=fr&locales%5B2%5D=pt&where%5Bslug%5D%5Bequals%5D=404`

but URLSearchParams ends up with something like `'?locale=en&fallback-locale=en&locales=en%2Cfr%2Cpt&where=%5Bobject+Object%5D'`

kilemensi: Damn, that's more like GraphQL-ing search params @koechkevin! `...&slug=404` should have been enough for REST API.

... but if that's the structure Payload expects, there is nothing we can do now.

🚀

apps/charterafrica/.env.template
koechkevin: Could it be something I am missing?  I keep getting this.
image

koechkevin: similar t when I try to access https://charterafrica.prod.codeforafrica.org/api/globals/footer

Unless I register and login.

kilemensi: Could be... Why are you connecting to prod though? Have you tried dev or local/Docker payload @koechkevin ?

kilemensi: Does it work on local/Docker @koechkevin ?

koechkevin: Not connecting to prod.

Just that the responses are similar.

koechkevin: Ah!

Just found out the problem

apps/charterafrica/src/payload/blocks/Errors.js
```diff
@@ -0,0 +1,47 @@
+import linkGroup from "../fields/linkGroup";
+import richText from "../fields/richText";
+
+const Errors = {
+  slug: "errors",
```
kilemensi, line 5:
```suggestion
const Error = {
  slug: "error",{
    singular: {
      en: "Error",
      fr: "Erreur",
      pt: "Erro",
    },
    plural: {
      en: "Errors",
      fr: "Les Erreurs",
      pt: "Erros",
    },
  },
```

kilemensi (commented):
👍🏽 @koechkevin 

-----

Good stuff... Almost there.

Restructure the `lib/data` into `lib/data/local` and `lib/data/rest`. As their names imply, they'll use their namesake APIs to fetch data from Payload. Put all common code i.e. code that process data after fetching, into `lib/data/common`. This way, the only thing different the two libs will be how data is fetched.


`lib/data/local` should export `getPageServerSideProps` while `lib/data/rest` can export `getPageStaticProps`. `lib/data/index.js` will then export both `getPageServerSideProps` for current pages to continue using without breaking, and `getPageStaticProps` to be used by error pages since they need to be created at build time (and any future page that needs to be static by design or performance).

apps/charterafrica/src/pages/_error.page.js
```diff
@@ -0,0 +1,38 @@
...
+      case "500":
+        return <Error key={block.slug} {...block} />;
+      default:
+        return null;
+    }
+  });
+}
+// TODO find a way to deal with other status codes
+// Status code is only accessible on getStaticProps but does not work with payload
+export async function getStaticProps(args) {
```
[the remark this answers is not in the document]
koechkevin: `getInitialProps` is missing crucial arguments such as `locale`  @kilemensi

apps/charterafrica/src/pages/_document.page.js
```diff
@@ -114,7 +114,7 @@ MyDocument.getInitialProps = async (ctx) => {
     originalRenderPage({
       enhanceApp: (App) =>
         function EnhanceApp(props) {
-          return <App emotionCache={cache} {...props} />;
+          return <App emotionCache={cache} {...props} res={ctx.res} />;
```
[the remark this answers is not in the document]
koechkevin: I should remove this.

apps/charterafrica/src/pages/_error.page.js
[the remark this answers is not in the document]
kilemensi: What I'm trying to bring to your attention @koechkevin is the fact that `_error` needs to be rendered whenever an error occurs; client-side, server-side, at runtime. This means `getStaticProps` **should not** be used i.e. if we've set `statusCode` to **500** and generate html file for that at build time (that's what `getStaticProps` does), what will happen when **501** error occurs?

As long as you've read the [caveats](https://nextjs.org/docs/advanced-features/custom-error-page#caveats), if `getServerSideProps` work, great lets use that. Otherwise, we'll have to use `getInitialProps`.... with a bit of planning ahead, we can forward `locale` information to `getInitialProps` through the [custom](https://github.com/vercel/next.js/discussions/20242#discussioncomment-232060) [app](https://stackoverflow.com/a/65438136).

kilemensi:
@kelvinkipruto please review.

kelvinkipruto (commented):
Awesome stuff @koechkevin LGTM

kelvinkipruto (approved):

kilemensi (approved):
🚀

kilemensi:
Lets get this merge @koechkevin

PR #328, opened 2023-02-09
Add payload config transpilation

kilemensi:
## Description

Somewhat under the radar, Payload removed all automatic transpiling that they were doing to the configs under the hood with the [v1.6.1](https://github.com/payloadcms/payload/releases/tag/v1.6.1) minor release. While this wouldn't have been a breaking change if we were using TypeScript, it's unfortunately a major issue for us since we're still on JavaScript.

As recommended in the [Payload PR](https://github.com/payloadcms/payload/pull/1847) that introduced this change, this PR adds minimal TypeScript to our monorepo to support transpilation of Payload config both in dev and when building for prod.

## Type of change

- [x] Chore

## Screenshots

N/A

## Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my own code
- [X] I have commented my code, particularly in hard-to-understand areas


kelvinkipruto (approved):
Looks Good

kelvinkipruto:
@kilemensi Having added `Typescript`, does that mean we can create some `blocks` and `fields` using typescript?

kilemensi:
> @kilemensi Having added `Typescript`, does that mean we can create some `blocks` and `fields` using typescript?

Yes @kelvinkipruto ... if there is a need to.

kilemensi:
Does this PR make sense @koechkevin? It fundamentally changes how we interact with Payload so I'd like everyone to be on board with it.

PR #350, opened 2023-02-27
Fix charterafrica build process

kilemensi:
## Description

We had a single build command that would build 3 separate items: **payload**, **server.ts** (transpiling to server.js basically), and **next.js**. Two of these items, **payload** and **next.js** clashed due to the `PAYLOAD_PUBLIC_APP_URL`.

If `PAYLOAD_PUBLIC_APP_URL` value is set to anything other than `localhost` (which it should be for _PROD_), Next.js will try to fetch data from this remote instance instead of the local instance running on custom server. If this value is not set however, Payload will try to redirect to `localhost` on PROD for `/admin` which would of course fail.

This PR separates the build process for Next.js (first) and Payload (last). During Next.js build, `PAYLOAD_PUBLIC_APP_URL` will point to localhost leading to successful Next.js build. After that, `PAYLOAD_PUBLIC_APP_URL` is set to a public URL leading to a successful build of Payload.

**ASIDE**: This PR also reverts #349 . While the bug in #349 is still there, starting with a clean DB should allow us to quickly re-create the pages we need. Trying to continue with #349 will take a little bit longer than the time we have right now.

## Type of change

- [x] Chore

## Screenshots

N/A

## Checklist:


thepsalmist (commented):
We can bump this to `0.0.25`, we had built `0.0.24` to validate failures from CIs `0.0.23`

kilemensi:
> We can bump this to `0.0.25`, we had built `0.0.24` to validate failures from CIs `0.0.23`

Cool @thepsalmist but lets also push changes to main so that main **always** point to latest build version.

thepsalmist (approved):
This LGTM @kilemensi

Issue #323, opened 2023-02-06
@/charterafrica CORS error


thepsalmist:
**Describe the bug**
Access to fetch at 'https://charterafrica.prod.codeforafrica.org/api' from origin 'https://charter.africa'  being blocked by CORS

PR #356, opened 2023-02-28
chaterafrica define CSRF config

thepsalmist:
## Description

This build off #323 , allowing CORS enables requests from multiple domains as defined in `PAYLOAD_ALLOWED_DOMAINS`. However, payload's `login`, `logout` and `refresh` use HTTP cookies for auth, and doesn't allow authentication for any other domain (other than that defined during build time). These have to be defined in the `csrf` config.

## Type of change

## Checklist:


kilemensi (commented):
👍🏽 

----

I understand your reasoning (especially how we use the domains so far) but I think these two settings should be configured via separate env vars.

Imagine we need to pull data from this site (say, news) to show on Code for Africa blog. We'd need to add `codeforafrica.org` domain to list of `CORS` domains but we 100% do not want to authenticate with that domain (`CSRF`).

apps/charterafrica/payload.config.ts
```diff
@@ -33,7 +33,11 @@ dotenv.config({ path: path.resolve(__dirname, "./.env.local") });
 
 const appURL = process.env.PAYLOAD_PUBLIC_APP_URL;
 
-const allowedDomains = process?.env?.PAYLOAD_ALLOWED_DOMAINS?.split(",")
+const allowedCORSDomains = process?.env?.PAYLOAD_CORS_DOMAINS?.split(",")
+  ?.map((d) => d.trim())
+  ?.filter(Boolean);
+
+const allowedCSRFDomains = process?.env?.PAYLOAD_CSRF_DOMAINS?.split(",")
```
kilemensi, line 40:
```suggestion
const cors = process?.env?.PAYLOAD_CORS?.split(",")
  ?.map((d) => d.trim())
  ?.filter(Boolean);

const csrf = process?.env?.PAYLOAD_CSRF?.split(",")
```

I take it these are just URLs and not necessary domains?

apps/charterafrica/payload.config.ts
```diff
@@ -92,8 +96,8 @@ export default buildConfig({
       },
     }),
   },
-  cors: allowedDomains,
-  csrf: allowedDomains,
+  cors: allowedCORSDomains,
+  csrf: allowedCSRFDomains,
```
kilemensi, line 100:
```suggestion
  cors,
  csrf,
```

kilemensi (approved):
🚀

apps/charterafrica/payload.config.ts
thepsalmist: Yes, these are URLS

kelvinkipruto (approved):

PR #359, opened 2023-03-01
Handle undefined CSRF and CORS

kelvinkipruto:
This PR fixes server crashing when `PAYLOAD_CSRF` is not provided in environment variables


kilemensi (approved):
🚀

PR #343, opened 2023-02-22
Ft/event-card

kelvinkipruto:
## Description
This PR implements `Event` Card as per the design

Fixes #321

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

## Screenshots
screencapture-localhost-3000-opportunities-fellowships-2023-03-01-09_56_28

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation

PR #363, opened 2023-03-01
Feature/336 posts events

koechkevin:
## Description

Fixes # (issue)

## Type of change

## Checklist:


kilemensi:
Any progress on this @koechkevin?

apps/charterafrica/src/lib/data/common/index.js
```diff
@@ -200,43 +173,7 @@ async function processPageFellowships(page) {
...
-      },
-      registerLink: {
-        href: `/register/events/${i}`,
-      },
-      registerText: "Register ",
-      status: ["upcoming", "past"][Math.floor(Math.random() * 2)],
-      featured: i === 0,
-    })),
+    items: events,
     config: {
```
koechkevin, line 177: @kilemensi how should we handle this config on CMS?.

kilemensi: Global @koechkevin. From the [docs](https://payloadcms.com/docs/configuration/globals):

> Globals are perfect for things like header nav, site-wide banner alerts, **`app-wide localized strings`**, and other "global" data that your site or app might rely on.
>

koechkevin:
> Is this up for review @koechkevin or are you still working on it?

It is now ready @kilemensi

apps/charterafrica/src/components/FeaturedPostCard/FeaturedPostCard.js
```diff
@@ -93,6 +93,7 @@ const FeaturedPostCard = React.forwardRef(function FeaturedPostCard(
             <RichText
               sx={{
                 maxHeight: 24 * 6,
+                overflow: "hidden",
```
kilemensi, line 96: `lineClamp` should handle this. What's the issue?

apps/charterafrica/src/components/OpportunityCard/OpportunityCard.js
```diff
@@ -86,7 +86,10 @@ const OpportunityCard = React.forwardRef(function OpportunityCard(props, ref) {
             variant="p1"
             elements={excerpt}
             sx={(theme) => ({
-              height: `calc(${theme.typography.p1.fontSize}px * 3)`,
+              maxHeight: `calc(${
+                theme.typography.p1.fontSize * theme.typography.p1.lineHeight
+              }px * 3)`,
+              overflow: "hidden",
```
kilemensi, line 92: 1. lineClamp should handle overflow,
2. switch to maxHeight means if element has 1 line, it will take height 1 line, if 2, it will take height of 2, etc. Doesn't this mess up the alignment? We want all card to align perfectly.

apps/charterafrica/src/lib/data/common/index.js
```diff
@@ -90,159 +90,81 @@ async function processPageExplainers(page, api) {
   return page;
 }
 
-async function processPageFellowships(page) {
+export async function processPageFellowships(page, api, { locale }) {
```
kilemensi, line 93: Why are we `export`-ing? Is it used anywhere outside this file?

apps/charterafrica/src/lib/data/common/index.js
```diff
@@ -90,159 +90,81 @@ async function processPageExplainers(page, api) {
...
+  const fellowships = fellowshipDocs.map((item) => ({
+    ...item,
+    description: item.excerpt,
+    image: item.coverImage,
+    status: item.category ?? null,
+    date: new Date(item.deadline).toLocaleString(locale, {
+      year: "numeric",
+      month: "long",
+      day: "numeric",
+    }),
```
kilemensi, line 120: I think lets create a reusable function for this.

apps/charterafrica/src/lib/data/common/index.js
```diff
@@ -90,159 +90,81 @@ async function processPageExplainers(page, api) {
...
+    blocks.find(
+      (block) =>
+        block.slug === "featured-post" &&
+        block.featuredPost?.relationTo === "events"
+    )?.featuredPost?.value ?? null;
+
+  const fellowships = fellowshipDocs.map((item) => ({
+    ...item,
+    description: item.excerpt,
+    image: item.coverImage,
```
kilemensi, line 114: Lets pass the required props only.

apps/charterafrica/src/lib/data/common/index.js
```diff
@@ -90,159 +90,81 @@ async function processPageExplainers(page, api) {
...
   });
+
+  const configs = await api.findGlobal("page-config");
+
+  const featuredArticle =
+    blocks.find(
+      (block) =>
+        block.slug === "featured-post" &&
+        block.featuredPost?.relationTo === "events"
+    )?.featuredPost?.value ?? null;
```
kilemensi, line 109: Move this to where you're dealing with events.

apps/charterafrica/src/payload/blocks/KnowledgePageHeader.js
```diff
@@ -0,0 +1,18 @@
+const KnowledgePageHeader = {
```
kilemensi, line 1: What is this for? News and Research title comes from news/research page title.

apps/charterafrica/src/payload/collections/Events.js
```diff
@@ -0,0 +1,98 @@
...
+      label: {
+        en: "Title",
+        fr: "Titre",
+        pt: "Título",
+      },
+      type: "text",
+      localized: true,
+      required: true,
+    },
+    slug({ fieldToUse: "name" }),
```
kilemensi, line 30: I don't see a `name` field

apps/charterafrica/src/payload/collections/Events.js
```diff
@@ -0,0 +1,98 @@
...
+      admin: {
+        date: {
+          pickerAppearance: "dayAndTime",
+        },
+        position: "sidebar",
+      },
+    },
+    linkGroup({
+      overrides: {
+        label: {
```
kilemensi, line 88:
```suggestion
    linkGroup({
      overrides: {
        name: "register",
        label: {
```

I think a name such as `register` is better than a generic `link`.

apps/charterafrica/src/payload/collections/Fellowships.js
```diff
@@ -0,0 +1,97 @@
...
+          value: "technologies",
+        },
+        {
+          label: {
+            en: "Other",
+            pt: "Outro",
+            fr: "Autre",
+          },
+          value: "other",
+        },
```
kilemensi, line 69: Cool, but lets also confirm if these are the only 2 categories we should support.

apps/charterafrica/src/payload/collections/Fellowships.js
```diff
@@ -0,0 +1,97 @@
...
+        en: "Excerpt",
+        fr: "Extrait",
+        pt: "Excerto",
+      },
+      localized: true,
+      admin: {
+        elements: ["leaves"],
+      },
+    }),
+    content(),
```
kilemensi, line 84: If we're making content optional in one (e.g. Event), we should make it optional in all. Or is there a reason why it's required is some but not others?

apps/charterafrica/src/payload/collections/Grants.js
```diff
@@ -0,0 +1,104 @@
...
+      localized: true,
+      admin: {
+        elements: ["leaves"],
+      },
+    }),
+    content(),
+    {
+      name: "deadline",
+      type: "date",
+      required: true,
```
kilemensi, line 95: Just thinking out loud, if a grant is upcoming, does it need a deadline? Shouldn't deadlines only be required if status is open/closed?

apps/charterafrica/src/payload/collections/Tags.js
```diff
@@ -17,7 +17,30 @@ const Tags = {
       localized: true,
       required: true,
     },
-    slug(),
+    slug({ fieldToUse: "name" }),
+    {
+      name: "collectionType",
```
kilemensi, line 22: Do we need this? Doesn't payload allow offer APIs to query tags from news collection separate from research? I mean there is already a relationship right?

My point being: a tag `Africa` should be reusable in news, research, grants, etc., without having to duplicate it.

apps/charterafrica/src/payload/globals/PageConfig.js
```diff
@@ -0,0 +1,259 @@
+const PageConfig = {
```
kilemensi, line 1: Cool, but I think:

1. We should call this translation strings or something. They're site wide and not page.
2. May be we should include a help text to show where it's being used e.g. showAll -> used to toggle number of visible cards on opportunities page, etc.
3. For these, may be we should also include default values since we don't want the site to break when someone has entered events but hasn't gotten around to entering translations yet.
4. May be we should define a js/json file with all the keys an reuse them in payload as well as data functions. Otherwise spelling mistakes will creep in.

kilemensi (commented):
👍🏽 ... good stuff.

apps/charterafrica/src/components/FeaturedPostCard/FeaturedPostCard.js
koechkevin: I am facing this.

image

payload 
```js
[
{
"children": [
{
"text": "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum.",
"children": null
}
]
},
{
"children": [
{
"children": null,
"text": "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum."
}
]
},
{
"children": [
{
"children": null,
"text": ""
}
]
},
{
"children": [
{
"children": null,
"text": "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum."
}
]
},
{
"children": [
{
"children": null,
"text": ""
}
]
},
{
"children": [
{
"children": null,
"text": ""
}
]
},
{
"children": [
{
"children": null,
"text": ""
}
]
}
]
```

kilemensi: Couple of things @koechkevin :

1. I believe lineClamp happens where there is ....
2. Fact that here is another line starting after that, it tells me it's probably another paragraph or something.

Excerpt should just be one paragraph so may be we should be passing that?

kilemensi:
One thing I forgot in my review @koechkevin: Fellowships and Grants need `apply` link (similar to `register` link on Events).

apps/charterafrica/src/payload/collections/Tags.js
koechkevin: When creating a  slug, we may need to specify which collection(s) does this slug apply to.

This is necessary for filtering purposes.

Perhaps maybe have this field as  `hasMany`. @kilemensi

kilemensi: I still don't understand the problem @koechkevin so I can't really say if `hasMany` will help or not.

Can you give a scenario where we need to specify the collection(s) the slug applies to?

kilemensi: Yes but this is where it breaks down @koechkevin : if in News there is Kenya tag and now there is Grantee from Kenya and content team wants to label the Grantee using Kenya tag.

Isn't this approach going to create a new Tag altogether?

My thinking was tags can be shared in the CMS but on the front end, we'll have to filter tags by collection i.e. give me all tags that have been used in News. I'm assuming the relationship field would allow us to do that.

Isn't this the case?

koechkevin: Got it.

But I think that would be a bit complex.

kilemensi: That's why we're having this discussion @koechkevin ... Why do you think it will be complex?

koechkevin: The structure of a collection does not include the collections in which it has been linked to.

To fetch on the frontend would therefore mean fetching the collection e.g news itself then pick all the tags uniquely which would be an expensive operation @kilemensi

kilemensi: 🤔 ... so if you do say `collection?slug=news&depth=1 (or 2)`, wouldn't it include `tags` as well?

But, my main issue is I don't want us to duplicate tags just per collection. The proof is in the pudding, so may be let me pull and play with it locally.

apps/charterafrica/src/payload/collections/Tags.js
```diff
@@ -17,7 +17,30 @@ const Tags = {
       localized: true,
       required: true,
     },
-    slug(),
```
kilemensi, line 20: We messed up here: `slug` should be `tags`. We can't just change it in code because there is already data in production using `tag` slug.

We need to come up with a plan on how to migrate this field. cc @thepsalmist

apps/charterafrica/src/payload/collections/Tags.js
kilemensi: Having gone through this, I think lets stick to the old design i.e. no filter. Yes, it might be expensive to make extra call but I believe tags are just markers. We shouldn't allow duplication across collections.

May be we should even add a `unique: true` to the name field?

apps/charterafrica/src/lib/data/common/index.js
```diff
@@ -90,159 +92,86 @@ async function processPageExplainers(page, api) {
...
-      showOnMobile: ["technologies"],
-      statusGroupTitleSuffix: "",
-    },
+}
+
+async function processPageFellowships(page, api, { locale }) {
+  const { blocks } = page;
+  const { docs: fellowshipDocs } = await api.getCollection("fellowships", {
+    locale,
   });
```
kilemensi, line 157: Include sort in all these getCollection calls

kilemensi (commented):
👍🏽 

---

Lets push on collections (and may be we should include Grantees here as well?) part first so that content team can begin entering data.

We can leave the translation part on it's own PR for now since it needs a bit more work e.g. we shouldn't be duplicating items per collection, a single item, showAll for example, should be reusable across the site to reduce burden on content editors.

apps/charterafrica/src/lib/data/common/index.js
```diff
@@ -90,159 +92,86 @@ async function processPageExplainers(page, api) {
...
+    image: item.coverImage ?? null,
+    category: item.topic,
+    registerLink: item.register ?? null,
+    registerText: item?.register?.label ?? null,
+    status:
+      new Date(item.date).getTime() < new Date().getTime()
+        ? "past"
+        : "upcoming",
+    date: formatDate(item.date, { locale }),
+    featured: featuredArticle && item.id === featuredArticle?.id,
```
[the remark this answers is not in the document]
koechkevin: In this [PR](https://github.com/CodeForAfrica/ui/pull/343) featured event was implemented a bit differently prompting this approach.

apps/charterafrica/src/lib/data/common/index.js
```diff
@@ -270,25 +176,19 @@ function processPost(post, page, api, context) {
...
+    date: formatDate(post.publishedOn, { locale, includeTime: true }),
     link: {
       href,
     },
   };
 }
 
 async function processPageArticlePost(page, api, context) {
-  const { params } = context;
+  const { params, locale } = context;
```
kilemensi, line 187: Nice!

kilemensi (approved):
🚀

PR #364, opened 2023-03-01
basic post filter component

### LINKED ISSUE #347: @/charterafrica/components/PostFilter
## Description

Implement `PostFilter` component per design. Initially it will be used for News only (even though it should be quite easy to re-use it for all collections with similar structure to News).

## Screenshot

Search

kelvinkipruto:
This PR implements the articles-filter component.

Fixes #347 

## Type of change

## Checklist:

### DIFF
--- a/apps/charterafrica/src/components/ArticlesFilter/ArticlesFilter.js
+++ b/apps/charterafrica/src/components/ArticlesFilter/ArticlesFilter.js
@@ -1,0 +1,112 @@
+import { Section } from "@commons-ui/core";
+import { Box, Grid, Typography, Select, MenuItem, Chip } from "@mui/material";
+import React, { useState } from "react";
+
+import { neutral } from "@/charterafrica/colors";
+import SearchInput from "@/charterafrica/components/SearchInput";
+
+const ArticlesFilter = React.forwardRef((props, ref) => {
+  const { tags, categories } = props;
+  const [selectedCategory, setSelectedCategory] = useState(categories[0]);
+  const [selectedTags, setSelectedTags] = useState(tags[0]);
+
+  return (
+    <Box bgcolor="#fff" ref={ref}>
+      <Section
+        sx={{
+          py: 2.5,
+        }}
+      >
+        <Grid container gap={5}>
+          <Grid item>
+            <Typography variant="h5" color={neutral[800]}>
+              News
+            </Typography>
+          </Grid>
+          <Grid item>
+            <SearchInput
+              placeholder="Search News"
+              sx={{
+                backgroundColor: "#fff",
+                height: "36px",
+                width: "200px",
+              }}
+            />
+          </Grid>
+          <Grid item>
+            <Select
+              inputProps={{ "aria-label": "Without label" }}
+              onChange={(e) => setSelectedCategory(e.target.value)}
+              sx={{
+                height: "36px",
+                minWidth: "200px",
+              }}
+              defaultValue={selectedCategory}
+            >
+              {categories.map((category) => {
+                return (
+                  <MenuItem value={category} key={category}>
+                    {category}
+                  </MenuItem>
+                );
+              })}
+            </Select>
+          </Grid>
+          <Grid item container xs={12} sm={6} md={4} gap={1}>
+            {tags.map((tag) => {
+              return (
+                <Chip
+                  label={tag}
+                  onClick={() => setSelectedTags(tag)}
+                  sx={{
+                    backgroundColor:
+                      selectedTags === tag ? neutral[700] : neutral[50],
+                    color: selectedTags === tag ? neutral[50] : neutral[900],
+                  }}
+                  key={tag}
+                />
+              );
+            })}
+            {/* <Chip
+              label="All"
+              sx={{
+                backgroundColor: neutral[700],
+                color: neutral[50],
+              }}
+            /> */}
+            {/* <Chip
+              label="Africa"
+              sx={{
+                backgroundColor: neutral[50],
+                color: neutral[900],
+              }}
+            /> */}
+            {/* <Chip
+              label="Kenya"
+              sx={{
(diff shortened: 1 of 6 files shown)


apps/charterafrica/src/components/ArticlesFilter/ArticlesFilter.js
```diff
@@ -0,0 +1,101 @@
...
+    setSelectedTags(t);
+  };
+
+  useEffect(() => {
+    router.push({
+      pathname: router.asPath.split("?")[0],
+      query: queryParams,
+    });
+
+    // We don't want to listen to router changes here since we're the ones
```
koechkevin, line 32: Alternatively, you could declare pathname variable outside the react hook then add it into the dependancy array.

i.e 
```js
const pathname  = router.asPath.split("?")[0]
```
but either way this is a bad practice. It could cause recursion.

apps/charterafrica/src/components/ArticlesFilter/ArticlesFilter.js
```diff
@@ -0,0 +1,101 @@
...
+            <Select
+              inputProps={{ "aria-label": "Without label" }}
+              onChange={(e) => setSelectedSorting(e.target.value)}
+              sx={{
+                height: "36px",
+                minWidth: "200px",
+              }}
+              defaultValue={sort ?? ""}
+            >
+              {sorting.map((c) => {
```
koechkevin, line 71: use a more descriptive name here

koechkevin:
@kelvinkipruto good work!

I noticed the following.

1. Changing tags updates query params as expected but reloading the page resets the query params.
2.  Question, can't we have 2 tags selected at a time? Looks like in this PR on 1 tag filter can be selected at a time.

kilemensi:
Thanks for the update @kelvinkipruto .

1. SWR already [caches](https://swr.vercel.app/docs/advanced/cache) returned values based on key i.e. the URL data was fetched.
2. If we want `tags` for a given collection only (just `news` for now), why not make full use of Next.js API capabilities by introducing `/api/tags/news` or `/api/tags?collection=news` or etc.
    1. We can use payload's local APIs in `pages/api` hence no network latency,
    2. Whatever is returned, SWR will cache it and hence no need extra layer of state management on client-side.

kilemensi:
Any progress @kelvinkipruto ?

kelvinkipruto:
@kilemensi I've got it working without using `api`. Two issues though I'm trying to fix. On reload, it clears all query params, and how to handle a lot of tags. Should we use a limited number or allow scrolling?

kilemensi:
> @kilemensi I've got it working without using `api`. Two issues though I'm trying to fix. On reload, it clears all query params, and how to handle a lot of tags. Should we use a limited number or allow scrolling?

@kelvinkipruto 

1. Scrolling is no no... I've never seen scrolling tags.
2. On get it working without API, how then do we ensure the work to compute all tags is save once and cached? We don't want every request to recompute the same thing.

kilemensi:
Any update on the approach @kelvinkipruto ?

kelvinkipruto:
### New approach to filtering Post
We need to get all news `tags`. Based on our current implementation, PayloadCMS doesn't allow us to query tags that belong to a collection, so the only approach would be to get all `news` and filter out the `tags`.

***We can achieve this by using `NextJs Api`, querying all news articles, filtering out all tags and caching them.***

We also cannot display all tags, so we need to score the tags and display the top, say 5 based on a criteria.

1. How do we score them? Scoring them based on the most commonly used seems like the best option.
2. What happens to the rest of the tags? 

> ***Let's say a user visits the news page without any params:***

We can go with the above approach and display 5 most common tags.
>Say they filter by tags:

1. Do we update the available tags with new ones from the filtered news that are not in the most common or just retain the most common tags?

> Say they filter by search:

Currently search works by searching for matching text on the title

1. Should search also look into content? (FlatMap all of it and look through it)
2. Do we also update the tags with the new ones from the filtered news or retain the most common ones

> Say a user visits the page with some params?

1. Should we display the tags for the filtered news articles or the most common ones


### My new Proposed approach
Using NextJS api, get all news, filter out the tags, then loop through them showing say 5 of them for a short period of time.
This should be able to handle both when a user visits with or without params

koechkevin:
> 

@kelvinkipruto your approach would work but my concern would be the performance implication of the algorithm when the articles list grows.

Have we considered option such as 

1. Having the designs tweaked to include more dropdown tag? With a multi select option.
2. What if we have tags displayed match only those that have been used in the current fetched page? (This could require clear all option)

kelvinkipruto:
> the performance implication of the algorithm when the articles list grows

Fetching of the tags will be done on the server side. I don’t think it could affect the performance that much.



> 2\. What if we have tags displayed match only those that have been used in the current fetched page?

I think the idea is to show possibly show all tags and not maybe limit them

@kilemensi any ideas or comments on this?

kelvinkipruto:
@kilemensi @koechkevin 


> In order of priority:

Here is what we can now do. 

- `Tags` are pulled using `Next API`. We then score them and pick 5 of the most used of them. Tags do not change when filtering.
- We can sort by `publishedOn`. We can add ability to sort by `title` if necessary.
- We can search by `title`. I did not manage to get searching by `content` working using any of the Payload Query [operators](https://payloadcms.com/docs/queries/overview#operators).
- The tags are dynamic and work for other slugs `knowledge\news` and `knowledge\research`
- I could not get visiting the page with slugs working. Any ideas?

koechkevin:
@kelvinkipruto good progress. However these issues still exist

1. Search by typing makes an API call on every keystroke. Consider debouncing the fetch.
2. On selecting tags the API call ay take some time creating a notion that filter is not working, can we find a way of having a loader for when the request is being mmade?
3. On page reload, the page loses previously selected tags (We may need to change the approach).

koechkevin (approved):

PR #367, opened 2023-03-03
charterafrica individual grants/fellowship pages

kilemensi:
## Description

This PR implements the individual Grant and Fellowship pages per designs. The big change is introduction of Opportunities block in the CMS that allows content editors to group and show various collections together e.g. show Grants and Fellowships together but Events on their own.

## Type of change

## Screenshots

S

## Checklist:


koechkevin (approved):
LGTM

kilemensi:
@kelvinkipruto ?

kelvinkipruto:
@kilemensi It looks like the [featured event](https://github.com/CodeForAfrica/ui/blob/2b793894ad2c4144e733c4f8a53f30cfc6bb3980/apps/charterafrica/src/payload/blocks/Opportunities.js#L26-L35) repeats on each item. Perhaps it could go into the [`items` ](https://github.com/CodeForAfrica/ui/blob/2b793894ad2c4144e733c4f8a53f30cfc6bb3980/apps/charterafrica/src/payload/blocks/Opportunities.js#L37-L38)array to allow selection of featured on each item.

kelvinkipruto:
Also when visiting the details page for each individual item, say `opportunities/events/event-slug` it returns 404 unless I create an empty events page in the CMS. Not sure if this is caused by how I have set up my CMS or it should be this way.

kilemensi:
> It looks like the [featured event](https://github.com/CodeForAfrica/ui/blob/2b793894ad2c4144e733c4f8a53f30cfc6bb3980/apps/charterafrica/src/payload/blocks/Opportunities.js#L26-L35) repeats on each item. Perhaps it could go into the [items ](https://github.com/CodeForAfrica/ui/blob/2b793894ad2c4144e733c4f8a53f30cfc6bb3980/apps/charterafrica/src/payload/blocks/Opportunities.js#L37-L38)array to allow selection of featured on each item.

Do you mean _each_ `opportunities` block has **one** featured item? That is by design. We now only have featured for events but this design allows setting up featured for grants and fellowships _without_ code change.

If you mean something else, let me know @kelvinkipruto 

> Also when visiting the details page for each individual item, say opportunities/events/event-slug it returns 404 unless I create an empty events page in the CMS. Not sure if this is caused by how I have set up my CMS or it should be this way.

This is correct again. We use grants page for title and any seo metadata. Similarly for fellowships, we use fellowships page, etc.

kelvinkipruto:
@kilemensi here is how I have set up my `opportunities page` in the CMS
opportunities-cms
and this is how it appears on the website:
opportunities-page
The featured event appears on each item. In my opinion, the featured item should be scoped to item group itself, i.e `featured-event` should only appear in `events` and not in `grants` or `fellowships`

kilemensi:
CMS looks fine @kelvinkipruto : a **single** opportunities block that shows all 3 type of opportunities; events, fellowships and grants. The block also has one featured opportunity.

If I have to guess, it's the front-end part that _repeats_ the featured item since `FeaturedPost` is rendered _inside_ `OpportunityPage` component.

kilemensi:
Yep, here is the front-end code that repeats @kelvinkipruto : https://github.com/CodeForAfrica/ui/pull/367/files#diff-35c1cf4c0979b48d2e068c02d156bf3783048ecec612295907f5e163cc2c5acfR34

Needs to move `FeaturedPost` outside of `OpportunityCards` to move it outside.

kelvinkipruto:
@kilemensi Yes I agree. But as per the [designs](https://www.figma.com/file/gbnaO1FYPq16V2gAFbLOkl/CfA---EPD-Charter-Africa---Sep-2022?node-id=9108-28134&t=apU9tu3EzfGelGDy-4), the `featured-post-card` should appear after the `opportunity-card-header` if it exists. Not sure how we are going to achieve this if we move it outside `OpportunityCards`. We may need to refractor it more to achieve that.

kilemensi:
Pushed the change @kelvinkipruto . Retest and let me know.

kelvinkipruto:
@kilemensi Yes it now works ok, except that the `featuredPost` will only display at the top of the page. If we are to go by the designs, it should appear for any type of opportunity.

kilemensi:
> @kilemensi Yes it now works ok, except that the `featuredPost` will only display at the top of the page. If we are to go by the designs, it should appear for any type of opportunity.

It doesn't appear at the top of the page @kelvinkipruto.

It appears at the top of a **block**. You can have as many blocks as you want.

kilemensi:
All good @kelvinkipruto ?

kelvinkipruto:
@kilemensi It looks good. I was testing some implementations I had in mind. Just one thing, say you want to have a `featuredEvent` and a `featuredGrant`, that would not be possible in the current implementation, and would require on to create a new block for it.

kilemensi:
> @kilemensi It looks good. I was testing some implementations I had in mind. Just one thing, say you want to have a `featuredEvent` and a `featuredGrant`, that would not be possible in the current implementation, and would require on to create a new block for it.

Not sure what you mean not possible @kelvinkipruto .

My understanding

A featured item servers as an highlight / appertizer for the rest of items in that group. For example, in news page, you have the feature post and then the rest of posts.

Other than hero on the homepage which uses a carousel, the current designs use a single featured item.

Back to this PR.

You create a group that can contain any combination of events, fellowships and grants. This group can have one highlighted item.

If I want to highlight fellowships and grants separately, I'll create a group for fellowships only, and add a featured fellowship. I'll then create a group for grants, and add a featured grant.

Back to you: What scenario does this implementation doesn't cover?

kelvinkipruto:
@kilemensi Lets take an example of [this page](https://www.figma.com/file/gbnaO1FYPq16V2gAFbLOkl/CfA---EPD-Charter-Africa---Sep-2022?node-id=9108-28134&t=ZF3bG8qREyi6qU22-0) form the design. Only `grants` has a featured post, and not the rest. With the current implementation, the `featuredGrants` will appear at the top of the page in the `events` opportunity type. It should only appear on the `grants` opportunity type.
Here is how it looks in the CMS:
new-cms
and here is how it looks in the web
new Page

In my opinion, to achieve something similar to the design, the `featuredPost` should be for each item. I think the CMS should maybe look like this:
screencapture-localhost-3000-admin-collections-pages-63e4d2681a55b889733c1ee3-2023-03-08-14_14_05

kilemensi:
You have one `Opportunities` block in the CMS @kelvinkipruto so the featured item will appear at the top of that block (which happened to be at the top of the page).

Here, I've set up two `Opportunities` blocks in the dev CMS, each with it's own feature item and 👇🏽 is how it appears on the from end.

Screen Shot 2023-03-08 at 14 21 36


1. Feature appear at the top of a `Opportunities` block
2. A page can have unlimited number of `Opportunities` blocks.

kelvinkipruto:
@kilemensi Now I get it. Adding Multiple `opportunities blocks` solves the issue.

kelvinkipruto (approved):
🚀

PR #379, opened 2023-03-14
@/charterafrica/components/EmbededVideoList #335

### LINKED ISSUE #335: [ON HOLD] @/charterafrica/components/EmbededVideoList
## Description

Implement `EmbededVideoList` component per designs.

 - [x] We can focus on YouTube exclusively,
 - [x] Some items will only contain audio,
 - [x] This could be a youtube embed but since videos will be fetched as youtube urls, this component will only accept youtube video ID as a prop.

### Mobile
[Designs on Figma](https://www.figma.com/file/gbnaO1FYPq16V2gAFbLOkl/CfA---EPD-Charter-Africa---Sep-2022?node-id=10578%3A44859&t=1OqRrbQtsaXyczXE-4)

### Desktop

[Figma Designs](https://www.figma.com/file/gbnaO1FYPq16V2gAFbLOkl/CfA---EPD-Charter-Africa---Sep-2022?node-id=9658%3A37666&t=1OqRrbQtsaXyczXE-4)

koechkevin:
## Description

Adds a Youtube Embed Component  that accepts video Id Prop and width.

This Component will be used with YoutubeComments component to create `EmbededVideoList`.

Example usage.
```js
import YoutubeVideoPlayer from "@/charterafrica/components/YoutubeVideoPlayer";

const EmbededVideoList = () => {
       return  <YoutubeVideoPlayer width="1024" videoId="I1-sz23FHWU" />
}
```
Fixes #335 

## Type of change

## Screenshots
image

## Checklist:

### DIFF
--- a/apps/charterafrica/src/components/YoutubeVideoPlayer/YoutubeVideoPlayer.js
+++ b/apps/charterafrica/src/components/YoutubeVideoPlayer/YoutubeVideoPlayer.js
@@ -1,0 +1,79 @@
+/* eslint-disable jsx-a11y/media-has-caption */
+import { Box, styled } from "@mui/material";
+import PropTypes from "prop-types";
+import React, { useEffect, useRef } from "react";
+import videojs from "video.js";
+import "videojs-youtube";
+import "video.js/dist/video-js.css";
+
+const StyledDiv = styled(Box)({
+  minWidth: "100%",
+  height: "100%",
+  "& .video-js .vjs-big-play-button": {
+    height: "100%",
+    left: 0,
+    top: 0,
+    width: "100%",
+    borderRadius: 0,
+    background: "rgba(62,32,44, 0.9)",
+  },
+  "& .video-js .vjs-big-play-button:hover": {
+    height: "100%",
+    left: 0,
+    top: 0,
+    width: "100%",
+    borderRadius: 0,
+    background: "rgba(62,32,44, 0.8)",
+  },
+  ".vjs-button > .vjs-icon-placeholder:before, .video-js .vjs-big-play-button .vjs-icon-placeholder:before":
+    {
+      position: "relative",
+      display: "flex",
+      justifyContent: "center",
+      alignItems: "center",
+      fontSize: 70,
+    },
+});
+
+function YoutubeVideoPlayer({ videoId, width }) {
+  const videoRef = useRef(null);
+  const playerRef = useRef(null);
+  const parentRef = useRef(null);
+
+  const src = `https://www.youtube.com/watch?v=${videoId}`;
+
+  useEffect(() => {
+    if (videoRef.current && !playerRef.current) {
+      const options = {
+        autoplay: false,
+        controls: true,
+        preload: "auto",
+        sources: [{ src, type: "video/youtube" }],
+        techOrder: ["youtube"],
+        height: "auto",
+        width,
+        youtube: { ytControls: 2 },
+      };
+      playerRef.current = videojs(videoRef.current, options);
+    }
+  }, [src, width]);
+
+  return (
+    <StyledDiv ref={parentRef}>
+      <div data-vjs-player>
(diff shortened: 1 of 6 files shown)


apps/charterafrica/src/components/YoutubeVideoPlayer/YoutubeVideoPlayer.js
```diff
@@ -0,0 +1,79 @@
...
+        width,
+        youtube: { ytControls: 2 },
+      };
+      playerRef.current = videojs(videoRef.current, options);
+    }
+  }, [src, width]);
+
+  return (
+    <StyledDiv ref={parentRef}>
+      <div data-vjs-player>
```
kilemensi, line 63: Can't both of this be on the same div?

apps/charterafrica/src/components/YoutubeVideoPlayer/YoutubeVideoPlayer.js
```diff
@@ -0,0 +1,79 @@
...
+  useEffect(() => {
+    if (videoRef.current && !playerRef.current) {
+      const options = {
+        autoplay: false,
+        controls: true,
+        preload: "auto",
+        sources: [{ src, type: "video/youtube" }],
+        techOrder: ["youtube"],
+        height: "auto",
+        width,
```
kilemensi, line 54: Any reason why we're going for fixed with instead of [`fluid`](https://docs.videojs.com/player#fluid)?

kilemensi (commented):
👍🏽 

-----

1. The play button size doesn't match design size (for various screen sizes).
2. On my end, there is a white border around the video. I don't see this in the designs.

Screenshot 2023-03-15 at 09-20-19 Screenshot

apps/charterafrica/src/components/YoutubeVideoPlayer/YoutubeVideoPlayer.js
koechkevin: If collapsed into 1 `div`, styles failed to apply.

So I had to have them as is.

kilemensi: Did you adjust styles as well? e.g. if on same `div`, `& .video-js` would become `&.video-js`

apps/charterafrica/src/components/YoutubeVideoPlayer/YoutubeVideoPlayer.js
koechkevin: Yes.

100% width didn't work with `height: auto`.

We end up with something like

image

So I thought it would be best if width is supplied as a prop.

apps/charterafrica/src/components/YoutubeVideoPlayer/YoutubeVideoPlayer.js
koechkevin: Yes

kilemensi: That's weird... I guess videojs removes existing `class` names and add it's own on the element with `data-vjs-player` attribute? The smartest thing would have been to _add_.

apps/charterafrica/src/components/YoutubeVideoPlayer/YoutubeVideoPlayer.js
kilemensi: That sounds like a `fill` mode. In `fluid` mode, all you need is an `aspectRatio`, no?

https://videojs.com/guides/layout/

kilemensi: @koechkevin ?

kilemensi (approved):
🚀 

---


Lets rename the PR to just `YouTubeVideoPlayer`. It's not a list just yet.

Issue #376, opened 2023-03-13
@charterafrica Youtube player CMS


koechkevin:
- [ ] As a content editor I want to be able to add youtube playlists.
- [ ]  Users can enter Playlist ID.
- [ ] Implemented this  as a collection.
- [ ] Using Youtube Player and Comments components, Implement [Designs](https://www.figma.com/file/gbnaO1FYPq16V2gAFbLOkl/CfA---EPD-Charter-Africa---Sep-2022?node-id=10578%3A46108&t=6ft5QfjPAhYj3q8o-4) on CMS.

PR #388, opened 2023-03-21
Feature/embed youtube video with comments

koechkevin:
## Description

- This PR combines `YoutubeVideoPlayer` with `CommentsComponent` and fetches comments using google API.
- Adds Consultations collection to CMS

```js
     import EmbedYoutubeVideo from "@/charterafrica/components/EmbedYoutubeVideo";

    <EmbedYoutubeVideo videoId="B0FKcdieE-Q" />
```

Fixes #379 #376 

## Screenshots
image

## Checklist:


apps/charterafrica/package.json
```diff
@@ -58,6 +58,7 @@
     "nanoid": "^4.0.1",
     "next": "^13.2.1",
     "next-seo": "^5.15.0",
+    "node-fetch": "^3.3.1",
```
kilemensi, line 61: Next.js already includes this.

apps/charterafrica/src/pages/api/v1/external/[...service].page.js
```diff
@@ -0,0 +1,35 @@
+import fetch from "node-fetch";
```
Comment (kilemensi) [inline apps/charterafrica/src/pages/api/v1/external/[...service].page.js:1]: Why external?

1. I'm assuming this is service specific i.e. YouTube in this PR.
2. We can probably keep navigation structure, for example, `opportunities/consultations/youtube` or videos or multimedia, etc.

kilemensi (commented):
👍🏽 

-----

Can you add more info on the approach? For example I was under the impression will only be asking for a playlist and featured video (should be in on the videos in the playlist) in the CMS but I see we've opted for a full collection.

koechkevin:
> 👍🏽
> 
> Can you add more info on the approach? For example I was under the impression will only be asking for a playlist and featured video (should be in on the videos in the playlist) in the CMS but I see we've opted for a full collection.

- I thought maybe they could choose to have as many playlists.
- An alternative approach would be to just have them as blocks without collection then users be forced to only have 2 playlists. But what happens when in future there are more categories other than `featured/active` and `previous`?

kilemensi:
> > 👍🏽
> > Can you add more info on the approach? For example I was under the impression will only be asking for a playlist and featured video (should be in on the videos in the playlist) in the CMS but I see we've opted for a full collection.
> 
>     * I thought maybe they could choose to have as many playlists.
> 
>     * An alternative approach would be to just have them as blocks without collection then users be forced to only have 2 playlists. But what happens when in future there are more categories other than `featured/active` and `previous`?

2. We'll be here to address the future when we get there @koechkevin, right? For now, YAGNI.
1. Why even 2 playlists? Just 1 should be enough. For the active/featured, they should just select a one of the video in that same playlist.

My thinking is the content team will create a single playlist that is specifically to contain videos that should appear on the website.

apps/charterafrica/src/payload/blocks/Consultation.js
```diff
@@ -0,0 +1,119 @@
...
+          hooks: {
+            beforeValidate: [mapPlaylistLinkToId],
+          },
+          type: "text",
+          localized: true,
+        },
+      ],
+    },
+    {
+      name: "featuredConsultations",
```
kilemensi, line 70:
```suggestion
      name: "featured",
```
1. Since this is already Consultation block, featured is enough. No need to repeat consultation.
2. I'm just thinking out loud but I think may be we should offer option of featured being latest/most recent or custom. When custom is selected, that's when we ask for video url

apps/charterafrica/src/payload/blocks/Consultation.js
```diff
@@ -0,0 +1,119 @@
...
+            hidden: true,
+          },
+          hooks: {
+            beforeValidate: [mapPlaylistLinkToId],
+          },
+          type: "text",
+          localized: true,
+        },
+      ],
+    },
```
kilemensi, line 68: Playlist content should be a group field (title e.g. Previous Consultations and URL i.e playlist URL) You can use help text to give an example of expected playlist URL.

This way we'll refer to it as playlist.url, playlist.title, etc.

apps/charterafrica/src/payload/blocks/Consultation.js
```diff
@@ -0,0 +1,119 @@
...
+          },
+          type: "text",
+          required: true,
+          localized: true,
+        },
+        {
+          name: "airedOn",
+          label: { en: "Aired On", fr: "Diffusé", pt: "Foi ao ar" },
+          type: "date",
+        },
```
kilemensi, line 93: Doesn't this mean the date it was published? Do we need to ask it manually?

apps/charterafrica/src/payload/blocks/FeaturedConsultation.js
```diff
@@ -0,0 +1,30 @@
+const FeaturedConsultation = {
```
kilemensi, line 1: Is this still needed/used somewhere?

apps/charterafrica/src/payload/utils/mapPlaylistLinkToId.js
```diff
@@ -0,0 +1,16 @@
+export const mapPlaylistLinkToId = ({ siblingData }) => {
+  const { playlistLink } = siblingData;
+  const searchParams = playlistLink?.split("?")?.[1];
+  const playlistId = new URLSearchParams(searchParams).get("list");
```
kilemensi, line 4: Or you can try creating a URL from link and read the [`searchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URL/searchParams) property. This way you may be able to even detect invalid URLs and show validation error in the CMS.

kilemensi (commented):
👍🏽 

---

I think UI is still a bit off. Here is how mobile looks for me.
Screenshot 2023-03-23 at 11-33-36 Consultation charter AFRICA

apps/charterafrica/src/lib/data/common/index.js
```diff
@@ -65,6 +67,59 @@ async function processPageExplainers(page, api, context) {
...
+async function processPageConsultations(page) {
+  const { blocks } = page;
+
+  const consultationIndex = blocks.findIndex(
+    ({ slug }) => slug === "consultation"
+  );
+
+  const consultation = blocks[consultationIndex];
+
+  const otherConsultations = await getVideosFromPlaylist(
```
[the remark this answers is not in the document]
koechkevin: Yes

apps/charterafrica/src/payload/utils/mapPlaylistLinkToId.js
kilemensi: Cool.

```js
  const url = new URL(playlistLink);
  const playlistId = url.searchParams.get("list");
```

No need for `split`, etc. Plus, you can wrap the `new URL` part in a `try`/`catch` to handle invalid URLs.

koechkevin: Yes

apps/charterafrica/src/pages/api/v1/opportunities/consultation/[...service].page.js
```diff
@@ -0,0 +1,49 @@
...
+  } catch (error) {
+    res.status(500).json({ error });
+  }
+};
+
+const youtube = async (req, res) => {
+  try {
+    const { service, ...rest } = req.query;
+    const url = service?.slice(1).join("/");
+    const key = process.env.GOOGLE_MAPS_API_KEY;
```
Comment (kilemensi) [inline apps/charterafrica/src/pages/api/v1/opportunities/consultation/[...service].page.js:18]:
```suggestion
    const key = process.env.GOOGLE_API_KEY;
```

Aren't we going to use a single Google API key for all supported services?

apps/charterafrica/src/pages/api/v1/opportunities/consultation/[...service].page.js
```diff
@@ -0,0 +1,49 @@
...
+    const queryParams = { key, ...rest };
+    const queryString = new URLSearchParams(queryParams).toString();
+    const youtubeBaseUrl = `https://www.googleapis.com/youtube/v3`;
```
Comment (kilemensi) [inline apps/charterafrica/src/pages/api/v1/opportunities/consultation/[...service].page.js:21]: This can be defined outside of this function.

apps/charterafrica/src/pages/api/v1/opportunities/consultation/[...service].page.js
```diff
@@ -0,0 +1,49 @@
...
+    res.status(response.status).json(data);
+  } catch (error) {
+    res.status(500).json(error);
+  }
+};
+
+const serviceMap = {
+  documents,
+  youtube,
+};
```
Comment (kilemensi) [inline apps/charterafrica/src/pages/api/v1/opportunities/consultation/[...service].page.js:33]:
```suggestion
const mediaMap = {
  documents,
  multimedia,
};
```

or audiovideo/av instead of multimedia.

apps/charterafrica/src/pages/api/v1/opportunities/consultation/[...service].page.js
```diff
@@ -0,0 +1,49 @@
...
+
+// We do not want apiKeys etc publicly availabe on the frontend
```
Comment (kilemensi) [inline apps/charterafrica/src/pages/api/v1/opportunities/consultation/[...service].page.js:35]: What is this comment for?

apps/charterafrica/src/pages/api/v1/opportunities/consultation/[...service].page.js
```diff
@@ -0,0 +1,49 @@
...
+      query: { service },
+    } = req;
+    const response = serviceMap[service?.[0]];
+    if (response) {
+      return response(req, res);
+    }
+    return res.status(404).json({ error: "NOT FOUND" });
+  } catch (error) {
+    return res.status(500).json(error);
+  }
```
Comment (kilemensi) [inline apps/charterafrica/src/pages/api/v1/opportunities/consultation/[...service].page.js:48]:
```suggestion
  const {
    query: { media },
  } = req;
  const mediaHandler = mediaMap[media];
  if (mediaHandler) {
    return mediaHandler(req, res);
  }
  return res.status(404).json({ message: "UNKNOWN_MEDIA", media });
```

Since handlers already have `try`/`catch`, I don't think we need it here.

apps/charterafrica/src/payload/blocks/Consultation.js
```diff
@@ -0,0 +1,127 @@
...
+              },
+              admin: {
+                description: () => "Playlist title e.g previous consultation",
+              },
+              type: "text",
+              required: true,
+              localized: true,
+            },
+            {
+              name: "playlistLink",
```
kilemensi, line 48:
```suggestion
              name: "link",
```

apps/charterafrica/src/payload/blocks/Consultation.js
```diff
@@ -0,0 +1,127 @@
...
+              label: {
+                en: "Playlist URL",
+                fr: "URL de liste de lecture",
+                pt: "URL da lista de reprodução",
+              },
+              type: "text",
+              admin: {
+                description: () =>
+                  "Valid youtube playlist URL e.g https://www.youtube.com/watch?list=RDEMPsM0esWCZxNU2OE89iz0kA or https://www.youtube.com/playlist?list=PLd9BS3XfsFcw4zmhpJWANl6HWc4olrKkg",
```
kilemensi, line 57:
```suggestion
                  "Valid YouTube playlist URL e.g https://www.youtube.com/watch?list=RDEMPsM0esWCZxNU2OE89iz0kA or https://www.youtube.com/playlist?list=PLd9BS3XfsFcw4zmhpJWANl6HWc4olrKkg",
```

apps/charterafrica/src/payload/blocks/Consultation.js
```diff
@@ -0,0 +1,127 @@
...
+              },
+              required: true,
+              localized: true,
```
kilemensi, line 60: Do we really see the need of different playlists per each language?

apps/charterafrica/src/payload/blocks/Consultation.js
```diff
@@ -0,0 +1,127 @@
...
+            },
+            {
+              name: "playlistId",
+              label: {
+                en: "Playlist ID",
+                fr: "Identifiant de la liste de lecture",
+                pt: "ID da lista de reprodução",
+              },
```
kilemensi, line 68: Since it's not visible in the admin, do we need labels?

apps/charterafrica/src/payload/blocks/Consultation.js
```diff
@@ -0,0 +1,127 @@
...
+              admin: {
+                hidden: true,
+              },
+              hooks: {
+                beforeValidate: [mapPlaylistLinkToId],
+              },
+              type: "text",
+              localized: true,
```
kilemensi, line 76: Same here, should this be localized?

apps/charterafrica/src/payload/utils/mapPlaylistLinkToId.js
```diff
@@ -0,0 +1,22 @@
...
+  }
+};
+
+export const mapVideoUrlToId = ({ siblingData }) => {
+  try {
+    const { videoLink } = siblingData;
+    const sharedId = videoLink.split("https://youtu.be/")?.[1];
+    const url = new URL(videoLink);
+    const videoId = url.searchParams.get("v");
+    return sharedId || videoId || "";
```
kilemensi, line 18: Is there a difference between returning `""` and `undefined`?

kilemensi (commented):
👍🏽

apps/charterafrica/src/pages/api/v1/opportunities/consultation/[...service].page.js
Comment (koechkevin) [reply apps/charterafrica/src/pages/api/v1/opportunities/consultation/[...service].page.js:18]: @kilemensi please confirm that this change requires [updating this line turbo.json](https://github.com/CodeForAfrica/ui/blob/e138473c8f5f317d8453383878b7c9b9a20b3e9a/turbo.json#L89).

One key can be used provided it is properly enabled on console.

Comment (kilemensi) [reply apps/charterafrica/src/pages/api/v1/opportunities/consultation/[...service].page.js:18]: Yes @koechkevin

apps/charterafrica/src/payload/blocks/Consultation.js
```diff
@@ -0,0 +1,121 @@
+import {
+  mapPlaylistLinkToId,
+  mapVideoUrlToId,
+} from "../utils/mapPlaylistLinkToId";
+
+const Consultations = {
+  slug: "consultation",
```
kilemensi, line 7:
```suggestion
const ConsultationsMultimedia = {
  slug: "consultation-multimedia",
```

Since we'll have a separate `ConsultationDocuments` I assume.

kelvinkipruto (approved):

koechkevin:
> 🚀
> 
> The only thing I think that makes sense is making the manual featured consultation optional. Default should be the latest consultation.
> 
> Bonus point is if you can pull videos in the playlist using one of the hook and convert the featured part from manually entering URL to a select.

@kilemensi I am asking a re review this.

To enable select, I used React Component since no hook works best (None is reactive to playlist change).

kilemensi (approved):
🚀

apps/charterafrica/src/payload/fields/react/customSelect.tsx
```diff
@@ -0,0 +1,44 @@
...
+    `/api/v1/opportunities/consultation/multimedia?${queryString}`,
+    fetcher
+  );
+  const memoOptions = () =>
+    data?.items?.map((video: any) => ({
+      label: video?.snippet?.title,
+      value: video?.snippet?.resourceId?.videoId,
+    })) || [];
+  const options = useMemo(memoOptions, [playlistId, data?.items?.length]);
+  return <Select {...props} options={options} />;
```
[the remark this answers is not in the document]
koechkevin: 🤔 I see not possible from this code,
https://github.com/payloadcms/payload/blob/1b85f194c5c760f959aa94b9ef657ae9ca4fc132/src/admin/components/forms/field-types/Select/Input.tsx#L98.

They have restricted props forwarded to primary component used [react-select](https://react-select.com/props)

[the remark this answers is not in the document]
kilemensi: Cool... lets just use admin.description then.

apps/charterafrica/tsconfig.json
```diff
@@ -20,7 +20,7 @@
       "@/commons-ui/next/*": ["../../packages/commons-ui-next/src/*"]
     }
   },
-  "include": ["next-env.d.ts", "**/*.ts"],
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
```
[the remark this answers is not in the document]
kilemensi: Not a fan but all good... Lets keep `.tsx`.

PR #404, opened 2023-04-06
Github integration

koechkevin:
## Description

This PR adds github integration

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/apps/charterafrica/src/lib/github/fetchGSheet.js
+++ b/apps/charterafrica/src/lib/github/fetchGSheet.js
@@ -1,0 +1,44 @@
+import fetchJson from "@/charterafrica/utils/fetchJson";
+
+export const getSheetsPerSpreadsheet = async ({ spreadSheetId }) => {
+  const params = {
+    key: process.env.GOOGLE_API_KEY,
+    fields: "sheets.properties.title",
+  };
+  const url = `https://sheets.googleapis.com/v4/spreadsheets/${spreadSheetId}`;
+  const d = await fetchJson.get(url, { params });
+  const { sheets } = d;
+  if (sheets && sheets?.length) {
+    return sheets.map((item) => item?.properties?.title);
+  }
+  return [];
+};
+
+export const fetchSpreadsheetPerSheet = async ({
+  spreadSheetId,
+  sheetName,
+}) => {
+  const range = `${sheetName}!a:z`;
+  const params = {
+    key: process.env.GOOGLE_API_KEY,
+  };
+  const url = `https://sheets.googleapis.com/v4/spreadsheets/${spreadSheetId}/values/${range}`;
+  const { values } = await fetchJson.get(url, { params });
+  if (values && values?.length) {
+    const keys = values[0];
+    const output = [];
+    values?.forEach((data, i) => {
+      if (i > 0) {
+        const item = {};
+        data.forEach((val, index) => {
+          item[keys[index]] = val;
+        });
+        output.push(item);
+      }
+    });
+    return output;
+  }
+  return [];
+};
+
+export default fetchSpreadsheetPerSheet;
--- a/apps/charterafrica/src/lib/github/fetchGithub.js
+++ b/apps/charterafrica/src/lib/github/fetchGithub.js
@@ -1,0 +1,36 @@
+import fetchJson from "@/charterafrica/utils/fetchJson";
+
+export const getRepoDetails = async ({ link, ...params }) => {
+  const nameAndOwner = link.replace("https://github.com/", "");
+  const res = await fetchJson.get(
+    `https://api.github.com/repos/${nameAndOwner}`,
+    { params }
+  );
+  return res;
+};
+
+export const getContributorsPerRepo = async ({ link, ...params }) => {
+  const nameAndOwner = link.replace("https://github.com/", "");
+  const res = await fetchJson.get(
+    `https://api.github.com/repos/${nameAndOwner}/contributors`,
+    { params }
+  );
+  return res;
+};
+
+export const getUserDetails = async ({ user, ...params }) => {
+  const res = await fetchJson.get(`https://api.github.com/users/${user}`, {
+    params,
+  });
+  return res;
+};
+
+export const getOrganization = async ({ organization, ...params }) => {
+  const res = await fetchJson.get(
+    `https://api.github.com/orgs/${organization}`,
+    {
+      params,
+    }
+  );
+  return res;
+};


kilemensi (commented):
👍🏽 

---

Haven't dug deep yet but a few things seem to stand out:

1. Why are we doing a lot of "micro" data fetching? On gSheet we seem to have `fetchSpreadsheetPerSheet`, `getSheetsPerSpreadsheet`, etc. On GitHub we seem to fetch repo details separate from commits, etc. Isn't there a single way of getting all (or most) of what we need, and then process and catch locally?
2. Collections. If a collection is called `GithubOrganisation`, why then have a field called `organisationName` and not just `name`? There is even `organisationGitHub` field.
3. `v1/api/github`. I don't think we need to expose github and spreadsheet here. To the front-end, the api should return just the information the UI needs. Whether that comes from gSheet or GitHub is irrelevant to the UI. Unless of course, you've come across a use case that may need this distinction.
4. Have you thought things would change and work if we need to support another tools host e.g. GitLab?

koechkevin:
> 👍🏽
> 
> Haven't dug deep yet but a few things seem to stand out:
> 
> 1. Why are we doing a lot of "micro" data fetching? On gSheet we seem to have `fetchSpreadsheetPerSheet`, `getSheetsPerSpreadsheet`, etc. On GitHub we seem to fetch repo details separate from commits, etc. Isn't there a single way of getting all (or most) of what we need, and then process and catch locally?
> 2. Collections. If a collection is called `GithubOrganisation`, why then have a field called `organisationName` and not just `name`? There is even `organisationGitHub` field.
> 3. `v1/api/github`. I don't think we need to expose github and spreadsheet here. To the front-end, the api should return just the information the UI needs. Whether that comes from gSheet or GitHub is irrelevant to the UI. Unless of course, you've come across a use case that may need this distinction.
> 4. Have you thought things would change and work if we need to support another tools host e.g. GitLab?

1. We are doing this because Github API returns an object with an api link to other properties such as commits, languages and contributors. The organization field also has some other attributes such as location, description, email and twitter missing.  We are using `getSheetsPerSpreadsheet` to get a list of sheets in a document to enable selecting of the specific sheet to source data from.

koechkevin:
> 👍🏽
> 
> Haven't dug deep yet but a few things seem to stand out:
> 
> 1. Why are we doing a lot of "micro" data fetching? On gSheet we seem to have `fetchSpreadsheetPerSheet`, `getSheetsPerSpreadsheet`, etc. On GitHub we seem to fetch repo details separate from commits, etc. Isn't there a single way of getting all (or most) of what we need, and then process and catch locally?
> 2. Collections. If a collection is called `GithubOrganisation`, why then have a field called `organisationName` and not just `name`? There is even `organisationGitHub` field.
> 3. `v1/api/github`. I don't think we need to expose github and spreadsheet here. To the front-end, the api should return just the information the UI needs. Whether that comes from gSheet or GitHub is irrelevant to the UI. Unless of course, you've come across a use case that may need this distinction.
> 4. Have you thought things would change and work if we need to support another tools host e.g. GitLab?

3. The distinction was necessary for fetching sheets per document on Admin (CMS) dashboard.

koechkevin:
> 👍🏽
> 
> Haven't dug deep yet but a few things seem to stand out:
> 
> 1. Why are we doing a lot of "micro" data fetching? On gSheet we seem to have `fetchSpreadsheetPerSheet`, `getSheetsPerSpreadsheet`, etc. On GitHub we seem to fetch repo details separate from commits, etc. Isn't there a single way of getting all (or most) of what we need, and then process and catch locally?
> 2. Collections. If a collection is called `GithubOrganisation`, why then have a field called `organisationName` and not just `name`? There is even `organisationGitHub` field.
> 3. `v1/api/github`. I don't think we need to expose github and spreadsheet here. To the front-end, the api should return just the information the UI needs. Whether that comes from gSheet or GitHub is irrelevant to the UI. Unless of course, you've come across a use case that may need this distinction.
> 4. Have you thought things would change and work if we need to support another tools host e.g. GitLab?

4. If a new source is added(with an api), we'd have a separate file to `fetchFromSource` with functions similar to those from github.

kilemensi:
Thanks @koechkevin.

writing this on mobile so excuse any typos.

My understanding is this:
1. Google sheet settings will be provided via a global in the CMS.
2. One or more corn jobs will run in the background pulling info from the gsheet, combine it with info from GitHub and finally store it in various collections.
3. UI will only need to show info from these collections.

Is my understanding ✅ or ❎?

kilemensi:
If ✅ @koechkevin then we only need to make one call to gsheet and pull the full list. We’ll the iterate over this list locally.

When fetching from GH then we may multiple calls for each tool but, if REST is the issue, doesn’t GraphQL allow us to combine queries? It will be ideal to make just one or two queries per tool and get everything we need.

koechkevin:
> If ✅ @koechkevin then we only need to make one call to gsheet and pull the full list. We’ll the iterate over this list locally.
> 
> When fetching from GH then we may have to make multiple calls for each tool but, if REST is the issue, doesn’t GraphQL allow us to combine queries? It will be ideal to make just one or two queries per tool and get everything we need.

Looks like GraphQL is the deal

koechkevin:
@kilemensi.

An update on this,

I couldn't find a Spreadsheet API to get all data in all sheets. Seems to default to the first sheet. This is what we do on globals.

1. User enters spreadsheet URL.
2. We use the URL to query GOOGLE API to get a list of sheets in that document.
3. User then selects sheet source for the data.
4. The cron(probably hourly) to then uses (Provided url and Sheet) to obtain data for processing(for any changes).
5. We have another cron that does the updates(Will probably run once a day) on all entries in the spreadsheet.

kilemensi:
> @kilemensi.
> 
> An update on this,
>

Sounds good @koechkevin expect:
>     5. We have another cron that does the updates(Will probably run once a day) on all entries in the spreadsheet.

Can you elaborate on this second cron job?

koechkevin:
> > @kilemensi.
> > An update on this,
> 
> Sounds good @koechkevin except:
> 
> > ```
> > 5. We have another cron that does the updates(Will probably run once a day) on all entries in the spreadsheet.
> > ```
> 
> Can you elaborate on this second cron job?

This cron is basically to  update all the records already cached for changes such as views count, stars, forks etc.

kilemensi:
Oh... "on all entries in the spreadsheet" made me think we're planning to update the spreadsheet @koechkevin .

Sounds alright then... this could be what you mean but just for clarity, I'll state it here:

1. **`cron` job one**: Pulls data from gSheet and create/delete collections in the CMS.
2. **`cron` job two**: Updates CMS collections by pulling most resent data from GitHub.

These two cron jobs don't need to know anything about each other as they'll work completely independent of each.

Is this where we're now?

koechkevin:
> Oh... "on all entries in the spreadsheet" made me think we're planning to update the spreadsheet @koechkevin .
> 
> Sounds alright then... this could be what you mean but just for clarity, I'll state it here:
> 
> 1. **`cron` job one**: Pulls data from gSheet and create/delete collections in the CMS.
> 2. **`cron` job two**: Updates CMS collections by pulling most resent data from GitHub.
> 
> These two cron jobs don't need to know anything about each other as they'll work completely independent of each.
> 
> Is this where we're now?

Correct.

kelvinkipruto (approved):

apps/charterafrica/src/lib/payload/index.js
```diff
@@ -21,17 +21,37 @@ async function getCollection(collection, options) {
...
+}
+
 async function findGlobal(slug, options) {
   return payload.findGlobal({
     ...options,
     slug,
   });
 }
 
+async function updateCollection(collection, data, options) {
```
kilemensi, line 39: I'm assuming id is required so
```suggestion
async function updateCollection(collection, id, data, options) {
```

apps/charterafrica/src/lib/tools/fetchGSheet.js
```diff
@@ -0,0 +1,52 @@
...
+    key: process.env.GOOGLE_API_KEY,
+  };
+  const res = await fetchJson.get(
+    `https://sheets.googleapis.com/v4/spreadsheets${path}`,
+    { params }
+  );
+  return res;
+}
+
+export const getSheetsPerSpreadsheet = async ({ spreadSheetId }) => {
```
kilemensi, line 15: Isn't this `getSpreadSheetSheetTitles` or something?

apps/charterafrica/src/lib/tools/fetchGSheet.js
```diff
@@ -0,0 +1,52 @@
...
+  };
+  const d = await fetchSpreadsheet(`/${spreadSheetId}`, params);
+  const { sheets } = d;
+  if (sheets && sheets?.length) {
+    return sheets.map((item) => item?.properties?.title);
+  }
+  return [];
+};
+
+export const fetchSpreadsheetPerSheet = async ({
```
kilemensi, line 27: `fetchSpreadSheetSheetByName` or something

apps/charterafrica/src/lib/tools/fetchGSheet.js
```diff
@@ -0,0 +1,52 @@
...
+  spreadSheetId,
+  sheetName,
+}) => {
+  const range = `${sheetName}!a:z`;
```
kilemensi, line 31: Doesn't putting just the sheet name in range accomplish the same thing?
```suggestion
  const range = `${sheetName}`;
```

apps/charterafrica/src/lib/tools/fetchGSheet.js
```diff
@@ -0,0 +1,52 @@
...
+  const { values } = await fetchSpreadsheet(
+    `/${spreadSheetId}/values/${range}`
+  );
+  if (values && values?.length) {
```
kilemensi, line 35:
```suggestion
  if (values?.length) {
```

apps/charterafrica/src/lib/tools/fetchGSheet.js
```diff
@@ -0,0 +1,52 @@
...
+    const keys = values[0];
+    const output = [];
+    values?.forEach((data, i) => {
```
kilemensi, line 38:
```suggestion
    values.forEach((data, i) => {
```

1. You've already check that value has length,
2. Can't we use `map`/`flatMap` to create `output`?

apps/charterafrica/src/lib/tools/fetchGithub.js
```diff
@@ -0,0 +1,77 @@
+import fetchJson, { FetchError } from "@/charterafrica/utils/fetchJson";
```
kilemensi, line 1: May be this should be lib/github or lib/tools/github. Can't be `fetGithub`. We are not fetching GH, we're fetching _from_ GH/

kilemensi (commented):
👍🏽 

---

Just a quick review to get you going.

koechkevin:
👍

apps/charterafrica/src/lib/payload/index.js
```diff
@@ -21,17 +21,37 @@ async function getCollection(collection, options) {
...
 }
 
+async function updateCollection(collection, id, data, options) {
+  const { id: _, ...rest } = data;
+  return payload.update({
+    collection,
+    id,
+    data: rest,
+    ...options,
+  });
```
kilemensi, line 46:
```suggestion
  return payload.update({
    ...options,
    collection,
    id,
    data,
  });
```

1. We've already made id a standalone arg,
2. `options` shouldn't override `collection`, `id`, etc.

apps/charterafrica/src/lib/tools/github.js
```diff
@@ -0,0 +1,77 @@
...
+  };
+  const data = {
+    variables,
+    query: GET_REPOSITORY,
+  };
+  const res = await fetchJson.post(url, { data, headers });
+  if (res?.data?.repository) {
+    return res.data.repository;
+  }
+  throw new FetchError("An Error occured", res.errors, 400);
```
kilemensi, line 76: Why are we assuming it's a 400?

apps/charterafrica/src/lib/tools/models.js
```diff
@@ -0,0 +1,99 @@
+import api from "../payload";
+
+export const ORGANIZATION_COLLECTION = "organisations";
+export const PEOPLE_COLLECTION = "people";
+export const TOOL_COLLECTION = "tool";
```
kilemensi, line 5:
```suggestion
export const TOOLS_COLLECTION = "tools";
```
Aren't collection in plural?

apps/charterafrica/src/lib/tools/models.js
```diff
@@ -0,0 +1,99 @@
...
+  const toCreate = {
+    ...rest,
+    people: createdPeople.map(({ id }) => id),
+    organisation: createdOrganization?.id,
+  };
+  const res = await api.createCollection(TOOL_COLLECTION, toCreate);
+  return res;
+};
+
+export const updateTool = async (data) => {
```
kilemensi, line 76: update and create seem to be very similar... I think we can achieve much better code reuse and readability.

apps/charterafrica/src/lib/tools/spreadsheet.js
```diff
@@ -0,0 +1,48 @@
...
+  return res;
+}
+
+export const getSpreadSheetSheetTitles = async ({ spreadSheetId }) => {
+  const params = {
+    fields: "sheets.properties.title",
+  };
+  const d = await fetchSpreadsheet(`/${spreadSheetId}`, params);
+  const { sheets } = d;
+  if (sheets && sheets?.length) {
```
kilemensi, line 21:
```suggestion
  if (sheets?.length) {
```

apps/charterafrica/src/pages/api/v1/github/[path].page.js
```diff
@@ -0,0 +1,55 @@
...
+  } catch (error) {
+    return res.status(500).json(error);
+  }
+};
+
+const sourceMap = {
+  "sheets-per-doc": sheetsPerSpreadsheet,
+  "sheet-data": sheetData,
+  processGsheet,
+  updateTools,
```
Comment (kilemensi) [inline apps/charterafrica/src/pages/api/v1/github/[path].page.js:43]: 1. Use single naming scheme; camelCase or kebab-case but not both.

apps/charterafrica/src/payload/collections/People.js
```diff
@@ -0,0 +1,105 @@
...
+        fr: "Source",
+      },
+      defaultValue: "github",
+      options: [
+        { label: "Github", value: "github" },
+        { label: "Gitlab", value: "gitlab" },
+        { label: "Bitbucket", value: "bitbucket" },
+      ],
+      required: true,
+    },
```
kilemensi, line 40: This can be a reusable field.

kilemensi (commented):
👍🏽 

---

1. Instead of TOOLS_CONFIG etc, I think lets call it ecosystem or digital democracy ecosystem.
2. Not sure why spreadsheet functionality is still being exposed in the API.

apps/charterafrica/src/lib/tools/github.js
koechkevin: Because Gitghub graphql errors' response status is 200. So this was default code since all other error codes have been handled by fetchJson

kilemensi: But 400 is a user/client-side error. Can a user do anything about this?

koechkevin: User may have entered a wrong github repository e.g `CodeForAfrica/non-existent-repository`. 
We may also default to `500`.

kilemensi: I don't think users enter anything into the system... content team, yes. Is

1. Throwing error the best approach, or
2. Just logging it somewhere (Sentry?) so that we know there is any issue with a given row?

koechkevin: At the moment, I am throwing all the errors wherever they occur.
All these errors are caught on controller level. Response object has a list of successful processes and failed processes.

PR #412, opened 2023-04-18
Ft/openAfrica API integration

### LINKED ISSUE #411: @charterafrica/openafrica
Add fetching datasets from [openAfrica](https://openafrica.net/)

Detailed document [here](https://docs.google.com/document/d/14iqZy4cAXuav1uxk0fEEb5JTQiM2Oli_o_px6g8bh7M/edit?usp=sharing)

kelvinkipruto:
## Description

This PR adds the feature to fetch documents from [openAfrica](https://openafrica.net/).

Fixes #411

## Type of change

## Checklist:


apps/charterafrica/src/pages/api/v1/data/datasets.js
```diff
@@ -0,0 +1,30 @@
...
+      const datasets = await fetchDatasets(q, {
+        fq: filterQuery,
+      });
+      res.status(200).json(datasets);
+    } catch (error) {
+      res.status(500).json({ error });
+    }
+  }
+
+  return res.json(404).json({ message: "NO ORGANIZATION PROVIDED" });
```
koechkevin, line 29:
```suggestion
  return res.status(404).json({ message: "NO ORGANIZATION PROVIDED" });
```

koechkevin (changes requested):

koechkevin (commented):
Overall, the code appears to be well-structured and follows best practices for handling API requests and responses.

kilemensi (commented):
👍🏽 

---

Haven't run it locally but I'm iffy about 2 things:
1. Why do we need the front-end to send the `organisation` via `v1/data/datasets`? I'm assuming we don't want this to change i.e. it should always point to the one set in the CMS.
2. Don't see anywhere we're handling pagination.

koechkevin (approved):

kilemensi:
> 👍🏽
> 
> Haven't run it locally but I'm iffy about 2 things:
> 
>     1. Why do we need the front-end to send the `organisation` via `v1/data/datasets`? I'm assuming we don't want this to change i.e. it should always point to the one set in the CMS.
> 

Heita @kelvinkipruto .

Where are we addressing point 1 above? i.e. pull organisation id from the CMS and use it for searching. Was hoping to see it somewhere in `src/lib/openafrica`

kilemensi:
@kelvinkipruto ?

kelvinkipruto:
@kilemensi Sorry I was updating how it fetches `tags`
In my design approach `src/lib/openafrica` should only interact with `openAfrica API` and not with the CMS. Getting data from CMS should be handled similarly to how we do in `sourceAfrica`. I have this [file](https://github.com/CodeForAfrica/ui/blob/f0eccf2697a34a474e84a1542ecfcbac211671f4/apps/charterafrica/src/lib/data/common/processPageDatasets.js) to show how we can handle it.
The organization will be passed to the component as [props](https://github.com/CodeForAfrica/ui/blob/e64e9666ec419163ed871078c81071cebc91ad55/apps/charterafrica/src/lib/data/common/processPageDatasets.js#L19) which we can then use when making API calls to `v1/data/datasets`

kelvinkipruto:
@kilemensi We [only fetch data](https://github.com/CodeForAfrica/ui/blob/e64e9666ec419163ed871078c81071cebc91ad55/apps/charterafrica/src/lib/data/common/processPageDatasets.js#L8-L9) if we have an `organization` provided in the CMS. The `organizationId` is required, so we're guaranteed to always have an organization to fetch data from. So yes, we can fetch data from any organization that is set up in the CMS.

kilemensi:
> @kilemensi We [only fetch data](https://github.com/CodeForAfrica/ui/blob/e64e9666ec419163ed871078c81071cebc91ad55/apps/charterafrica/src/lib/data/common/processPageDatasets.js#L8-L9) if we have an `organization` provided in the CMS. The `organizationId` is required, so we're guaranteed to always have an organization to fetch data from. So yes, we can fetch data from any organization that is set up in the CMS.

@kelvinkipruto I mean the API accepts **any** organisation not necessarily the one set in the CMS.

kelvinkipruto:
@kilemensi Yes, now I get you. The API will accept any organization passed to it.
Given that I have set the `datasets` as a page block, it would be difficult to fetch the `organizationId` from `src/lib/openafrica`. To solve it perhaps I could set datasets as a `collection`. This will make it possible to fetch the collection, hence the organization easily

kelvinkipruto:
@kilemensi I have updated it.

apps/charterafrica/src/lib/data/rest/index.js
```diff
@@ -53,6 +53,12 @@ export const api = {
   findPage,
 };
 
+export async function getOrganizationId() {
+  const data = await findGlobal("datasets");
+  const { organizationId } = data;
+  return organizationId;
+}
+
```
kilemensi, line 61: Not sure this is necessary... both pages via lib/data and APIs have access to `payload` object and can hence fetch the organisation `id`/`slug` _**before**_ passing it to `openAfrica` lib.

The `openAfrica` lib shouldn't need to know anything about data/rest, etc.

apps/charterafrica/src/lib/data/common/processPageDatasets.js
```diff
@@ -0,0 +1,23 @@
+import fetchDatasets from "@/charterafrica/lib/openAfrica";
+
+async function processPageDatasets(page) {
+  const { blocks } = page;
+
```
kilemensi, line 5:
```suggestion
```

apps/charterafrica/src/lib/data/common/processPageDatasets.js
```diff
@@ -0,0 +1,23 @@
...
+      rows: 10,
+      start: 0,
+    });
+
+    blocks[datasetsIndex] = {
+      slug: "datasets",
+      datasets,
+    };
+  }
+
```
kilemensi, line 19:
```suggestion
```

kilemensi (commented):
🚀

PR #416, opened 2023-04-25
Ft/datasets component

### LINKED ISSUE #415: @charterafrica/datasets-component
Implement the dataset UI component based on the [design](https://www.figma.com/file/gbnaO1FYPq16V2gAFbLOkl/CfA---EPD-Charter-Africa---Sep-2022?node-id=11922-110710&t=Z7Nx7iRu9mp7goaD-4)
Integrate API #412

kelvinkipruto:
## Description

This PR implement Components for the [datasets page](https://www.figma.com/file/gbnaO1FYPq16V2gAFbLOkl/CfA---EPD-Charter-Africa---Sep-2022?node-id=11727-96336&t=Z7Nx7iRu9mp7goaD-4)

Fixes #415 

## Type of change

## Screenshots
Datasets-Code-for-Africa-mobile
Datasets-Code-for-Africa-tab
Datasets-Code-for-Africa-full

## Checklist:


kelvinkipruto:
**Issues Faced**
I'm using [`useDatasets`](https://github.com/CodeForAfrica/ui/blob/2adaeda931530a77b83e55315fb0d50356f3b5e7/apps/charterafrica/src/components/Datasets/Datasets.js#L84-L92) too fetch data from the NextJS API. This works as expected.
I'm then using another variable [`resource`](https://github.com/CodeForAfrica/ui/blob/2adaeda931530a77b83e55315fb0d50356f3b5e7/apps/charterafrica/src/components/Datasets/Datasets.js#L34) to handle whether to show `datasets` or `documents` on the page. When `resource` is updated, `useDatasets` is called automatically, fetching the data from the API again. This should not be the case as we just want to show the documents.

kilemensi:
I'm not sure I follow @kelvinkipruto :

1. Looking at Figma, we need 2 pages: multiple datasets and single dataset pages (similar to articles and single article page).
2. These pages are completely separate or at least they should use completely different section components. Why do need toggle-able resource state?

kelvinkipruto:
@kilemensi On the first page when the show `documents` or `datasets` button is clicked, the document list rendered should update accordingly.
isuuuus

kilemensi:
> @kilemensi No design. I presume it should be the same as the dataset one.

Nah, I doubt it @kelvinkipruto. I think this section of the site is design to match that of organisation page in openAFRICA: https://open.africa/organization/the-charter-project

As you can see there, there is no "documents" view. Documents are just the links (PDF, CSV, etc.)

kelvinkipruto:
@kilemensi That simplifies it greatly, but then what should happen when the `documents` button above is clicked? Could I make it do nothing for now?

kilemensi:
> @kilemensi That simplifies it greatly, but then what should happen when the `documents` button above is clicked? Could I make it do nothing for now?

Link to the document in openAFRICA @kelvinkipruto

apps/charterafrica/src/components/Datasets/Datasets.js
```diff
@@ -0,0 +1,98 @@
...
+    tags: selectedTags,
+  });
+
+  useEffect(() => {
+    if (!isLoading) {
+      const { datasets: filteredDatasets, count } = data || {};
+      setDatasets(filteredDatasets);
+      setTotalPages(Math.ceil(count / pageSize));
+    }
+  }, [data, isLoading]);
```
kilemensi, line 71: Show linear progress bar when `isLoading` (similar to how Articles component handle it)

apps/charterafrica/src/lib/data/common/processPageResources.js
```diff
@@ -0,0 +1,47 @@
+import fetchDatasets, {
```
kilemensi, line 1: Why not `processPageDatasets`?

apps/charterafrica/src/lib/data/common/processPageResources.js
```diff
@@ -0,0 +1,47 @@
...
+} from "@/charterafrica/lib/openAfrica";
+
+// eslint-disable-next-line import/prefer-default-export
+export async function processPageDatasets(page, api) {
+  const { blocks } = page;
+  const { organizationId } = await api.findGlobal("datasets");
+  const { datasetCount, documentsCount } = await getOrganizationStatistics(
+    organizationId
+  );
+  const pieChartData = [
```
kilemensi, line 12: This should also be set in the CMS.

apps/charterafrica/src/lib/data/common/processPageResources.js
```diff
@@ -0,0 +1,47 @@
...
+  blocks.push({
+    slug: "datasets-charts",
+    data: pieChartData,
+  });
+
+  const data = await fetchDatasets(organizationId, {
+    rows: 10,
+    start: 0,
+  });
+  const { count, datasets, countries, tags } = data;
```
kilemensi, line 36: Check how `processPageArticles` caches the data via fallback so that it's available on the front-end for SWR calls.

kilemensi (commented):
👍🏽 

---

Why node-cache? We have the CMS where we can periodically update the organisation documents count.

apps/charterafrica/src/components/Datasets/Datasets.js
```diff
@@ -0,0 +1,98 @@
...
+          count={totalPages}
+          page={page}
+          onChange={handleChangePage}
+          sx={{ mt: 2.5, bgcolor: "common.white" }}
+        />
+      </Section>
+    </Box>
+  );
+}
+
```
koechkevin, line 97: @kelvinkipruto isn't there a need for proptype validation here?

kilemensi:
Is this ready for review @kelvinkipruto ?

.github/workflows/cron-dev.yml
```diff
@@ -0,0 +1,13 @@
+name: Dev Cron Jobs
```
kilemensi, line 1: Let [Dokku](https://dokku.com/docs/processes/scheduled-cron-tasks/#specifying-commands) handle cron jobs

apps/charterafrica/.env.template
```diff
@@ -13,6 +13,8 @@ S3_SECRET_ACCESS_KEY=
 S3_BUCKET=
 S3_REGION=
 
+CRONJOBS_SECRET_KEY=
```
kilemensi, line 16: Name things based on function e.g. `RESOURCES_EDIT_KEY`, `RESOURCES_SECRET_TOKEN`, etc. We need this key to edit resources and not crons.

apps/charterafrica/src/components/Datasets/useDatasets.js
```diff
@@ -0,0 +1,22 @@
...
+
+import queryString from "@/charterafrica/utils/datasets/queryString";
+
+const fetcher = (url) => fetch(url).then((res) => res.json());
+
+function useDatasets(query) {
+  const qs = queryString(query);
+  const separator = qs ? "?" : "";
+  const { data, error } = useSWR(
+    `/api/v1/data/datasets${separator}${qs}`,
```
kilemensi, line 11:
```suggestion
    `/api/v1/resources/datasets${separator}${qs}`,
```

datasets and ecosystem are all under resources I believe.

apps/charterafrica/src/lib/data/common/processPageDatasets.js
```diff
@@ -1,21 +1,60 @@
...
+    countries,
+    data: datasets,
+    sortOptions: options,
+    tags,
+    totalPages,
+  });
+
+  // SWR fallback
+  const swrKey = `/api/v1/data/datasets`;
+  const qs = `?rows=10&start=0`;
```
kilemensi, line 53: This should be computed from query params and not hard-coded like this.

apps/charterafrica/src/lib/openAfrica.js
```diff
@@ -2,16 +2,28 @@ import fetchJson from "@/charterafrica/utils/fetchJson";
...
+    return error;
+  }
+}
+
+function formatDatasets(data) {
+  const { result: { count, facets: { tags, groups }, results } = {} } =
+    data || {};
+
+  const documents = [];
+  const datasets = results?.map((dataset) => {
```
kilemensi, line 26: There has to be a cleaner way of handling this part... may be a standalone `formatDataset` function or something.

apps/charterafrica/src/lib/openAfrica.js
```diff
@@ -50,37 +73,83 @@ async function formatDatasets(data) {
...
 
 export default async function fetchDatasets(organization, query = {}) {
-  const { tags = [], ...other } = query;
-  const tagsQuery = tags.length ? `tags:(${tags.join(" OR ")})` : "";
+  const { tags = [], countries = [], page = 1, ...other } = query;
+  const tagsQuery = tags.length
+    ? `tags:(${tags.map((tag) => `"${tag}"`).join(" OR ")})`
+    : null;
+  const countriesQuery = countries.length
+    ? `groups:(${countries.map((country) => `"${country}"`).join(" OR ")})`
```
kilemensi, line 97: Instead of converting to array of string first only to call `join`, I think you should be able to use `.reduce` directly on `tags` or `countries` to convert them to a single string.

kilemensi (commented):
👍🏽 

---

Looks good. Haven't paid much attention to the UI just yet but having gone through the code, my immediate thoughts are:

1. Only thing we need in global is the openAFRICA organization-id. Everything else should move into page blocks. We can continue calling the global item `datasets` or we can change it to `openAFRICA` to emphasize the integration.
2. Lets create a `ResourceHeader` block (similar to the `AboutHeader` block) that contains `title` as well as option to select resource to render as chart i.e. `datasets` or `ecosystem` (or `tools`)
3. Lastly, lets have `Datasets` block (if we've renamed the global to `openAFRICA`) that will now contain the `sorting` options. It is in this block that we'll also show the list of datasets.

Let me know if they make sense.

kelvinkipruto:
@kilemensi I have addressed most of the issues we discussed. Only the charts part remain. I will update it once we've got feedback.

apps/charterafrica/src/utils/datasets/queryString.js
```diff
@@ -0,0 +1,32 @@
...
+ * /path?q2=2&q1=1.
+ */
+function queryString(query = {}) {
+  const searchParams = new URLSearchParams();
+  const { countries, page, sort, tags, q } = query;
+
+  if (q) {
+    searchParams.append("q", q);
+  }
+  if (sort && sort !== "") {
```
kilemensi, line 15:
```suggestion
  if (sort) {
```

This should cover against `undefined`, `null`, **and** `""`

apps/charterafrica/src/payload/globals/OpenAfrica.js
```diff
@@ -1,5 +1,6 @@
-const Datasets = {
-  slug: "datasets",
+const OpenAfrica = {
+  slug: "openAfrica",
+  label: "openAfrica",
```
kilemensi, line 3:
```suggestion
  label: "openAFRICA",
```
We talked about this.

apps/charterafrica/src/payload/blocks/Datasets.js
```diff
@@ -0,0 +1,191 @@
...
+                return data?.label || data?.value || data?.id;
+              },
+            },
+          },
+        },
+      ],
+    },
+    {
+      label: {
+        en: "Datasets Labels",
```
kilemensi, line 80:
```suggestion
        en: "Search & Filter Labels",
```

We're in Datasets block already, no need to keep mentioning it.

apps/charterafrica/src/payload/blocks/Datasets.js
```diff
@@ -0,0 +1,191 @@
...
+  "author asc",
+  "author desc",
+];
+
+const Datasets = {
+  slug: "datasets",
+  fields: [
+    {
+      label: {
+        en: "Sort Options",
```
kilemensi, line 21:
```suggestion
        en: "Search & Filter Values",
```

apps/charterafrica/src/payload/blocks/Datasets.js
```diff
@@ -0,0 +1,191 @@
...
+        pt: "Opções de classificação",
+      },
+      type: "collapsible",
+      fields: [
+        {
+          name: "sortOptions",
+          type: "array",
+          minRows: 1,
+          label: {
+            en: "Options",
```
kilemensi, line 32:
```suggestion
            en: "Sort Order",
```
or Sort Options, etc.

apps/charterafrica/src/pages/api/v1/resources/[action].page.js
```diff
@@ -4,23 +4,24 @@ import payload from "@/charterafrica/lib/payload";
 async function datasets(req, res) {
   const {
     query: {
```
Comment (kilemensi) [inline apps/charterafrica/src/pages/api/v1/resources/[action].page.js:6]: `[action].page.js` is only for actions i.e. update, create, resources. What you're doing here is just a plain data retrieval API endpoint.

This should be in `datasets.page.js`. It doesn't share anything with `[actions].page.js`. Otherwise, we're going to make `[actions].page.js` very complex.

apps/charterafrica/src/lib/openAfrica.js
```diff
@@ -23,62 +55,73 @@ async function formatDatasets(data) {
...
+  const { tags = [], countries = [], page = 1, ...other } = query;
+  const tagsQuery = tags.length
+    ? `tags:(${tags.reduce((acc, tag) => {
+        return acc ? `${acc} OR "${tag}"` : `"${tag}"`;
+      }, null)})`
+    : null;
+  const countriesQuery = countries.length
+    ? `groups:(${countries.reduce((acc, country) => {
+        return acc ? `${acc} OR "${country}"` : `"${country}"`;
+      }, null)})`
```
kilemensi, line 108: Isn't this just `countries.join(" OR ")`?

apps/charterafrica/src/lib/openAfrica.js
```diff
@@ -23,62 +55,73 @@ async function formatDatasets(data) {
...
 }
 
 export default async function fetchDatasets(organization, query = {}) {
-  const { tags = [], ...other } = query;
-  const tagsQuery = tags.length ? `tags:(${tags.join(" OR ")})` : "";
+  const { tags = [], countries = [], page = 1, ...other } = query;
+  const tagsQuery = tags.length
+    ? `tags:(${tags.reduce((acc, tag) => {
+        return acc ? `${acc} OR "${tag}"` : `"${tag}"`;
+      }, null)})`
```
kilemensi, line 103: Same here, isn't it just `tags.join(" OR ")`?

apps/charterafrica/src/lib/openAfrica.js
```diff
@@ -23,62 +55,73 @@ async function formatDatasets(data) {
...
+function formatResponse(data) {
+  const { result: { count, facets: { tags, groups }, results } = {} } =
+    data || {};
+
+  const datasets = formatDatasets(results);
+
+  const countries = Object.keys(groups || {}).sort((a, b) =>
+    a.localeCompare(b)
+  );
+  const tagsList = Object.keys(tags || {}).sort((a, b) => a.localeCompare(b));
```
kilemensi, line 87:
```suggestion
  const sortStrings = (a, b) => a.localeCompare(b);
  const countries = Object.keys(groups || {}).sort(sortStrings);
  const tagsList = Object.keys(tags || {}).sort(sortStrings);
```

apps/charterafrica/src/lib/openAfrica.js
```diff
@@ -23,62 +55,73 @@ async function formatDatasets(data) {
...
-        name: resourceName,
-        url,
-      };
-    });
+    const formattedResources = formatResources(resources, author);
 
-    const allDocumentFormats = [
+    const formats = [
       ...new Set(formattedResources?.map((resource) => resource.format)),
     ];
```
kilemensi, line 62:
```suggestion
    const formatsSet = new Set(resources?.map((r) => r.format));
    const formats = Array.from(formatsSet);
```

Shorter code is always easier to follow.

apps/charterafrica/src/lib/openAfrica.js
```diff
@@ -2,16 +2,48 @@ import fetchJson from "@/charterafrica/utils/fetchJson";
...
+      description,
+    } = resource;
+    return {
+      author,
+      created: resourceCreated,
+      description,
+      format,
+      id,
+      name: resourceName,
+      updated: resourceUpdated,
```
kilemensi, line 39: I'm completely lost here: Why are we renaming attributes e.g. `created` to `resourceCreated` on first pass and then rename them again back to the original names on returning e.g. `resourceCreated` to `created`?

apps/charterafrica/src/lib/data/common/processPageDatasets.js
```diff
@@ -1,21 +1,38 @@
...
+
+  return { countries, locale, page, q, tags, sort };
+};
+
+export default async function processPageDatasets(page, api, context) {
   const { blocks } = page;
+  const { organizationId } = await api.findGlobal("openAfrica");
+
+  const data = await fetchDatasets(organizationId);
+  const { count, datasets, countries, tags, totalPages } = data;
```
kilemensi, line 16: 1. Move this data fetching to the inside of `if(datasetsIndex > -1)`. Why fetch data if there isn't `datasets` block?
2. Also, only fetch data if `organizationId` is set.

apps/charterafrica/src/components/Datasets/Datasets.js
```diff
@@ -0,0 +1,119 @@
...
+    if (!isLoading) {
+      const { datasets: filteredDatasets, totalPages: filteredTotalPages } =
+        data || {};
+      setDatasets(filteredDatasets);
+      setTotalPages(filteredTotalPages);
+    }
+  }, [data, isLoading]);
+
+  return (
+    <Box bgcolor="common.white" sx={sx} ref={datasetsRef}>
```
kilemensi, line 85: Make sure this component has `scroll-margin` so that when next/prev is clicked, the full `FilterBar` is visible. See `Articles` component.

apps/charterafrica/src/components/Datasets/DatasetCard.js
```diff
@@ -0,0 +1,107 @@
...
+          <LineClampedRichTypography
+            color="common.black"
+            lineClamp={showAll ? -1 : 3}
+            order={1}
+            variant="p1"
+            sx={{ mb: 1 }}
+          >
+            {notes}
+          </LineClampedRichTypography>
+          <Button
```
kilemensi, line 64: 1. There is some left margin/padding on this button
2. It appears everywhere even where it should e.g. second row on attached image.
Screenshot 2023-05-16 at 09 10 03

apps/charterafrica/src/components/Datasets/DatasetCard.js
```diff
@@ -0,0 +1,107 @@
...
+            color="neutral.dark"
+            lineClamp={1}
+            order={1}
+            variant="h5SmallSemiBold"
+            sx={{ mb: 1 }}
+          >
+            {title}
+          </LineClampedRichTypography>
+          <Typography variant="p1" color="neutral.main" sx={{ mb: 1 }}>
+            Updated {formatDateTime(updated, { includeTime: false })} | Created{" "}
```
kilemensi, line 47: No hard-coded content.

apps/charterafrica/src/components/Datasets/DatasetCard.js
```diff
@@ -0,0 +1,107 @@
+import { Box, Grid, Typography, Chip, Button } from "@mui/material";
+import { useState } from "react";
+
+import LineClampedRichTypography from "@/charterafrica/components/LineClampedRichTypography";
+import formatDateTime from "@/charterafrica/utils/formatDate";
+
+function DatasetCard({
```
kilemensi, line 7: I don't see us using `Card` component here. Why call this a `Card`?

apps/charterafrica/src/components/Datasets/DatasetCard.js
```diff
@@ -0,0 +1,107 @@
...
+          </Button>
+        </Grid>
+        <Grid
+          item
+          xs={12}
+          md={4}
+          order={{ xs: 3, sm: 4 }}
+          display="flex"
+          justifyContent={{ xs: "flex-start", sm: "flex-end" }}
+          alignItems="center"
```
kilemensi, line 84: Are you trying to turn this into a container? Just use `<Grid item .... container justifyContent...`

apps/charterafrica/src/components/Datasets/DatasetCard.js
```diff
@@ -0,0 +1,107 @@
...
+        >
+          {formats?.map((format) => (
```
kilemensi, line 86: This whole `Grid` shouldn't be there if there are no `formats` to display, right?

kilemensi (commented):
👍🏽 

---

I'm still not sure where does the list of countries/tags is supposed to come from.

apps/charterafrica/src/components/Datasets/DatasetCard.js
kelvinkipruto: @kilemensi On my end I don't have the padding. Perhaps it's a space from the CMS? @koechkevin How does yours look like?
image

kilemensi: Duude focus. Who puts spaces in the CMS?

No there are no space(s) in the CMS
Screenshot 2023-05-16 at 12 25 54

Yes, it's styling issue. By default a button has `min-size`. Yours doesn't show probably because the text you're using is more than `min-size` i.e. `64px`. Mine shows because it's less than that.
Screenshot 2023-05-16 at 12 25 31

kilemensi:
My question still stands @koechkevin 

> 👍🏽
> 
> I'm still not sure where does the list of countries/tags is supposed to come from.

And if you've decided to use Card in DatasetCard, please make sure you've gone through the MUI [docs](https://mui.com/material-ui/react-card/#ui-controls) on how to use various card components (CardMedia, CardContent, and if the card is clickable, CardActionArea).

kilemensi:
> @kilemensi By adding the [`facet.field`](https://github.com/CodeForAfrica/ui/blob/c31c261521ce64c19418f1d3e5b6df363b27eeef/apps/charterafrica/src/lib/openAfrica.js#L112) param to the request, the response will include a list of all tags and groups. In this case, groups contain countries.

So how come they've been empty all this time for me @kelvinkipruto ?

kelvinkipruto:
Not sure I get you, but by empty, do you mean the `countries` and `tags` dropdown is empty?

kilemensi:
> Not sure I get you, but by empty, do you mean the `countries` and `tags` dropdown is empty?

Yes, that's what I'm asking @kelvinkipruto: where does the list of countries and tags come from?

kelvinkipruto:
Here is how we're currently pulling it out in the [formatResponse](https://github.com/CodeForAfrica/ui/blob/c31c261521ce64c19418f1d3e5b6df363b27eeef/apps/charterafrica/src/lib/openAfrica.js#L78-L79) function

kilemensi:
Put another way @kelvinkipruto: Do the countries and tags dropdown now include content other than just labels from the CMS?

kelvinkipruto:
@kilemensi Yes. Here is how it looks on my end:
image
image

kilemensi:
Cool @kelvinkipruto , will pull latest code and re-test once you've addressed:
>And if you've decided to use Card in DatasetCard, please make sure you've gone through the MUI [docs](https://mui.com/material-ui/react-card/#ui-controls) on how to use various card components (CardMedia, CardContent, and if the card is clickable, CardActionArea).

kelvinkipruto:
@kilemensi I have addressed it.

koechkevin (commented):
Check the font of autocomplete component

apps/charterafrica/src/components/Datasets/DatasetCard.js
```diff
@@ -0,0 +1,137 @@
...
+              <Grid
+                item
+                xs={12}
+                md={4}
+                order={{ xs: 3, sm: 4 }}
+                container
+                justifyContent={{ xs: "flex-start", sm: "flex-end" }}
+                alignItems="center"
+              >
+                {formats?.map((format) => (
```
kilemensi, line 113:
```suggestion
                {formats.map((format) => (
```
You've already checked `formats.length`

kilemensi (commented):
👍🏽

apps/charterafrica/src/components/Datasets/Datasets.js
```diff
@@ -0,0 +1,128 @@
...
+          onQChange={handleChangeQ}
+          onSortChange={handleChangeSort}
+          onChangeCountries={handleChangeCountries}
+          onChangeTags={handleChangeTags}
+        />
+        {isLoading ? <LinearProgress color="secondary" /> : null}
+        {datasets?.map((dataset) => (
+          <DatasetCard
+            key={dataset.name}
+            {...dataset}
```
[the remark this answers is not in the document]
kelvinkipruto: The name is also unique

kelvinkipruto:
@kilemensi I have addressed the pending issues.

apps/charterafrica/src/components/Datasets/DatasetCard.js
```diff
@@ -48,7 +49,8 @@ function DatasetCard({
           },
         }}
         component={Link}
-        href="#"
+        // TODO: Update URL to single dataset page
+        href={url}
```
kilemensi, line 53: Isn't this URL pointing to open.africa directly?

kelvinkipruto: @kilemensi Yes, just for now. In the other PR I'm working on now, I'm updating it to the single datasets page. I could also update it to # now if need be.

kilemensi: Nah, you don't need to update it here @kelvinkipruto ... Need to make sure you're working on the issue.

kilemensi:
The dropdown font isn't fixed @kelvinkipruto 
Screenshot 2023-05-18 at 14 57 49

kilemensi:
Changing filters (`countries`, `tags` at least ... and may be `sort`) should reset `page` back to **1** @kelvinkipruto.

kilemensi:
`Charter-Africa` isn't a country @kelvinkipruto ... We may need to filter it our 
Screenshot 2023-05-18 at 15 01 30

apps/charterafrica/src/lib/data/common/processPageDatasets.js
```diff
@@ -1,21 +1,38 @@
...
+      data: datasets,
+      tags,
+      totalPages,
     };
   }
+  const swrKey = `/api/v1/resources/datasets`;
+  const qs = datasetsQuery(getDatasetsQuery(context));
+  // eslint-disable-next-line no-param-reassign
+  page.fallback = {
+    [`${swrKey}${qs}`]: [],
```
kilemensi, line 34: You need to provide data here instead of `[]`. Otherwise, on first render, the page will not have any data.

apps/charterafrica/src/utils/datasets/queryString.js
```diff
@@ -0,0 +1,32 @@
...
+  if (sort) {
+    searchParams.append("sort", sort);
+  }
+  if (countries?.length > 0) {
+    searchParams.append("countries", countries);
+  }
+  if (tags?.length > 0) {
+    searchParams.append("tags", tags);
+  }
+
```
kilemensi, line 24:
```suggestion
```

apps/charterafrica/src/components/Datasets/Datasets.js
```diff
@@ -0,0 +1,128 @@
...
+          labels={labels}
+          sortOptions={sortOptions}
+          onQChange={handleChangeQ}
+          onSortChange={handleChangeSort}
+          onChangeCountries={handleChangeCountries}
+          onChangeTags={handleChangeTags}
+        />
+        {isLoading ? <LinearProgress color="secondary" /> : null}
+        {datasets?.map((dataset) => (
+          <DatasetCard
```
kilemensi, line 108: Since DatasetCard has a border, this way, 2 DatasetCards next to each other are going to have 2 borders: border-bottom of top card and border-top of bottom card.

You need to but these cards inside say a MUI `Stack` and hide on of the borders e.g. show only top border for first item but hide for the rest, etc., using css selectors `:first-child` or :first-of-type

kilemensi (commented):
👍🏽

kelvinkipruto:
@kilemensi Addressed all pending issues

apps/charterafrica/src/lib/data/common/processPageDatasets.js
```diff
@@ -1,21 +1,39 @@
...
+      data: datasets,
+      tags,
+      totalPages,
+    };
+
+    const swrKey = `/api/v1/resources/datasets`;
+    const qs = datasetsQuery(getDatasetsQuery(context));
+    // eslint-disable-next-line no-param-reassign
+    page.fallback = {
+      [`${swrKey}${qs}`]: [],
```
kilemensi, line 34: This is still empty (`[]`) Should be something like: https://github.com/CodeForAfrica/ui/blob/96257d7c90e8147c92f45766676c165251d61fe0/apps/charterafrica/src/lib/data/common/processPageArticles.js#L248-L251

apps/charterafrica/src/lib/openAfrica.js
```diff
@@ -1,84 +1,131 @@
...
+function formatResponse(data) {
+  const { result: { count, facets: { tags, groups }, results } = {} } =
+    data || {};
+
+  const datasets = formatDatasets(results);
+  const sortStrings = (a, b) => a.localeCompare(b);
+  const uniqueCountries = Object.keys(groups || {}).sort(sortStrings);
+  const validCountries = uniqueCountries.filter((country) =>
+    allCountries.africa
+      .map((c) => c.toLowerCase())
```
kilemensi, line 91: We have to lower-case **everytime**? Why not just convert them to lower case **once** and store them in allCountries?

apps/charterafrica/src/lib/openAfrica.js
```diff
@@ -1,84 +1,131 @@
...
+      .includes(country.toLowerCase())
+  );
+
```
kilemensi, line 94:
```suggestion
```

apps/charterafrica/src/lib/openAfrica.js
```diff
@@ -1,84 +1,131 @@
...
-        url,
-      };
-    });
+    const formattedResources = formatResources(resources, author);
 
-    const allDocumentFormats = [
-      ...new Set(formattedResources?.map((resource) => resource.format)),
-    ];
+    const formatsSet = new Set(resources?.map((r) => r.format));
+    const formats = Array.from(formatsSet);
```
kilemensi, line 64:
```suggestion
    const formats = Array.from(formatsSet).sort();
```
Small thing but I think it will better if we consistently show formats in one order e.g. CSV, PDF or PDF, CSV every time there is more than one format. Mixing them up doesn't really look... neat.

kilemensi (commented):
👍🏽

kilemensi (approved):
🚀

PR #427, opened 2023-05-23
Ft/single dataset component

### LINKED ISSUE #426: @charterAfrica/single-datasets-view
Implement single datasets components as per the [designs](https://www.figma.com/file/gbnaO1FYPq16V2gAFbLOkl/CfA---EPD-Charter-Africa---Sep-2022?type=design&node-id=11727-96642&t=cziNtAp6k61caDOA-4)

kelvinkipruto:
## Description
This PR implements the single datasets components

Fixes #426 

## Type of change

## Checklist:


kilemensi:
I thought this was ready @kelvinkipruto ?

kelvinkipruto:
@kilemensi It is now. I was updating a few thing.

apps/charterafrica/src/components/Datasets/DatasetCard.js
```diff
@@ -41,11 +39,7 @@ function DatasetCard({
         ...sx,
       }}
     >
-      <CardActionArea
-        component={Link}
-        // TODO: Update URL to single dataset page
-        href={url}
-      >
+      <CardActionArea component={Link} href={`${pageUrl}/${id}`}>
```
kilemensi, line 42: Just create a `href` prop when formatting the dataset.

apps/charterafrica/src/lib/openAfrica.js
```diff
@@ -123,3 +123,24 @@ export default async function fetchDatasets(organization, query = {}) {
...
 }
+
+export async function fetchDataset(id) {
+  try {
+    const response = await fetchJson.get(
+      `${BASE_DOCUMENTS_URL}package_show?id=${id}`
+    );
+    const { result: dataset } = response;
+    const { tags = [] } = dataset;
+    const tagsNames = tags.map((tag) => tag.name);
```
kilemensi, line 134: Have we confirmed that all datasets have tags? Otherwise we may have to have a backup plan e.g. tags first if available, groups if no tags, etc.

apps/charterafrica/src/components/Datasets/index.js
```diff
@@ -1,3 +1,5 @@
+import Dataset from "./Dataset";
```
kilemensi, line 1: Isn't `Dataset` a standalone component? i.e. Why does it need to be a child of `Datasets`?

apps/charterafrica/src/payload/blocks/Datasets.js
```diff
@@ -207,6 +207,62 @@ const Datasets = {
                     width: "50%",
                   },
                 },
+                {
+                  name: "openDataset",
```
kilemensi, line 211: Are all of these extra labels you're adding part of **search and filter** because that's our title.

apps/charterafrica/src/lib/openAfrica.js
```diff
@@ -123,3 +123,24 @@ export default async function fetchDatasets(organization, query = {}) {
...
+    const formattedDataset = formatDatasets([dataset]);
+    const relatedDatasets = await fetchDatasets(dataset.organization.name, {
+      tags: tagsNames,
+    });
+    return {
+      ...formattedDataset[0],
+      relatedDatasets: relatedDatasets.datasets.slice(0, 3),
```
kilemensi, line 141: This can also be just `related`. We know it's dataset.

kilemensi (commented):
👍🏽 

---

In the following images,

Screenshot 2023-05-25 at 09 32 06
Screenshot 2023-05-25 at 09 32 38

1. One of the **Open** is supposed to take you to the source of the dataset, and another to openAFRICA. I believe the one with big PDF image is probably the one that should take you to openAFRICA.
2. Have you confirmed with the design team why we need this big PDF image? openAFRICA doesn't have embed functionality/thumbnails similar to sourceAFRICA.
3. Not sure why we have **GitHub** link in a dataset.

kilemensi:
@kelvinkipruto ?

apps/charterafrica/src/components/Datasets/DatasetCard.js
kelvinkipruto: @kilemensi This will require us to use a hardcoded string `/resources/datasets/` when formatting datasets. Won't this cause an issue if for example the `dataset `block is used in a different page?

kelvinkipruto: @kilemensi Not sure I get you. Here is my quick view:
Say we are on the page `/resources/datasets` if we click on one of the datasets, it should take us to `/resources/datasets/id`. We can easily add this in the `processPageDatasets` function, but this would not work as subsequent calls to `/api/v1/resources/datasets` would not know which page it was called from to append the id of each dataset.
To add the individual URL in each dataset in the [formatDataset](https://github.com/CodeForAfrica/ui/blob/fb3692fe328c3210e46d2d210ebefb1f958a19d6/apps/charterafrica/src/lib/openAfrica.js#L46) function, we'll need to pass around the url to append the dataset id to, both in the `processPageDatasets` funtion and the `useDatasets` hook, which I think is quite unnecessary as it is not information required to fetch data from openAfrica.

kilemensi: 3. Every `req` on API side also comes with payload i.e. `req.payload`. You can load the `datasets` page as well without requiring anything passed-in from the front-end.

kelvinkipruto:
@kilemensi Looks ready. Still waiting for info from the design team about the pdf and Github button

apps/charterafrica/src/lib/openAfrica.js
```diff
@@ -72,17 +72,19 @@ function formatDatasets(datasets) {
       name,
       notes,
       url: url?.trim(),
+      href: `https://openafrica.net/${type}/${id}`,
+      localHref: `${path}/${id}`,
```
kilemensi, line 76: `href` should be to charter.africa page. This is what the card should open when clicked. You can use `url` and `source` to mean either openAFRICA link or link to where openAFRICA got the dataset from i.e. source of the data.

apps/charterafrica/src/lib/openAfrica.js
```diff
@@ -96,7 +98,7 @@ function formatResponse(data) {
 }
 
 export default async function fetchDatasets(organization, query = {}) {
```
kilemensi, line 100:
```suggestion
export default async function fetchDatasets(organization, pathname, query = {}) {
```

Since `pathname` is now required, make it a standalone arg.

apps/charterafrica/src/lib/openAfrica.js
```diff
@@ -117,9 +119,35 @@ export default async function fetchDatasets(organization, query = {}) {
...
     const response = await packageSearch(params);
-    const formattedData = formatResponse(response);
+    const formattedData = formatResponse(response, path);
     return formattedData;
   } catch (error) {
     return error;
   }
 }
+
+export async function fetchDataset(id, path) {
```
kilemensi, line 129:
```suggestion
export async function fetchDataset(id, pathname) {
```

[`path`](https://nodejs.org/api/url.html#urlobjectpath) is some old school pathname + query or something like that.

kilemensi (approved):
:rocket

apps/charterafrica/src/pages/api/v1/resources/datasets.page.js
```diff
@@ -9,6 +9,7 @@ async function datasets(req, res) {
       countries,
       q = "",
       page = 1,
+      path,
```
kilemensi, line 12:
```suggestion
      pathname,
```

apps/charterafrica/src/components/Dataset/Dataset.js
```diff
@@ -0,0 +1,275 @@
...
+                mb: 2,
+              }}
+            >
+              {commonLabels.backToDatasets}
+            </Typography>
+            <Typography variant="h4" component="h2" gutterBottom>
+              {title}
+            </Typography>
+            <Typography variant="p1" color="neutral.main" sx={{ mb: 1 }}>
+              {commonLabels.updated}{" "}
```
koechkevin, line 58: Why not use string literal instead?

PR #432, opened 2023-05-26
Fix sourceAFRICA group label

### LINKED ISSUE #431: Allow Multiple Groups for sourceAfrica documents
Allow creation of multiple sourceAfrica groups for use in different sections of the CMS

kelvinkipruto:
## Description

This PR fixes the sourceAFRICA label to [clarify](https://dc.sourceafrica.net/help/searching) that what we need is organisation short name i.e. group `id` and not `name`.

## Type of change

## Screenshots

N/A

## Checklist:


kilemensi:
Can you may be describe the approach @kelvinkipruto?

kelvinkipruto:
@kilemensi Updated.

kilemensi:
Thanks @kelvinkipruto but I'm still not sure why we need a collection though.

My understanding is the only thing needed for sourceAFRICA integration is just the organisation/group id. Why not just type it in the block? What does having a separate collection & extra relationship add to the mix?

kelvinkipruto:
Ahaa, So simple. Thanks.

apps/charterafrica/src/payload/blocks/EmbeddedDocuments.js
```diff
@@ -46,14 +46,13 @@ const EmbeddedDocuments = {
       type: "group",
       fields: [
         {
-          name: "group",
+          name: "groupId",
```
kilemensi, line 49: What about the current code that's using group? Deploying this code as is would break that functionality.

kilemensi (approved):
🚀

PR #434, opened 2023-05-29
Airtable

koechkevin:
## Description
This PR Integrates AIRTABLE API, GITHUB API to the CMS to conclude github integration.

To test this PR,
- Set up environment variables `AIRTABLE_API_TOKEN` and `RESOURCES_SECRET_TOKEN`.
- Update global configs http://localhost:3000/admin/globals/ecosystem-config
- Hit the endpoint http://localhost:3000/api/v1/resources/ecosystem/update-ecosystem-content
## Type of change

## Checklist:


kilemensi (commented):
👍🏽 

-------

Haven't dug deep but:
1. Are we doing both Airtable -> CMS and GitHub -> CMS in this one PR?
2. Any reason why we're going with raw fetch instead of [`Airtable.js`](https://github.com/Airtable/airtable.js)?

apps/charterafrica/app.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "charterafrica",
+  "scripts": {
+    "predeploy": "export NEXT_PUBLIC_APP_URL=$(dokku config:get charterafrica NEXT_PUBLIC_APP_URL) && export RESOURCES_SECRET_TOKEN=$(dokku config:get charterafrica RESOURCES_SECRET_TOKEN)",
+    "update-ecosystem-list": "curl -H \"x-api-key: $RESOURCES_SECRET_TOKEN\" $NEXT_PUBLIC_APP_URL/api/v1/resources/ecosystem/update-ecosystem-list",
+    "update-ecosystem-content": "curl -H \"x-api-key: $RESOURCES_SECRET_TOKEN\" $NEXT_PUBLIC_APP_URL/api/v1/resources/ecosystem/update-ecosystem-content"
+  },
```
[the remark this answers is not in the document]
koechkevin: > 

1. This PR is only for Airtable integration. 
2. I didn't use `Airtable.js` just not to install an extra dependancy for something we can write.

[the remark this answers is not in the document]
koechkevin: > Is `update-ecosystem-content` part of this PR? I'm assuming this will be done in GitHub integration.

I had this impleented to provide for a way to test this PR @kilemensi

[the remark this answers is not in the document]
kilemensi: Cool, cool... I think airtable.js handles a few things:

1. Retry: https://airtable.com/developers/web/api/rate-limits
2. Parameter encoding: https://airtable.com/developers/web/api/list-records

If we're sure we can manage, all good.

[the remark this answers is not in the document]
koechkevin: We can manage parameter encoding since we are using tableID instead of tableName but Retry could be needed if the list grows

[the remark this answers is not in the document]
kilemensi: > Is update-ecosystem-content part of this PR? I'm assuming this will be done in GitHub integration.
> 
> I had this impleented to provide for a way to test this PR @kilemensi

I was talking about updating content. This PR is all about updating list, no?

apps/charterafrica/src/lib/ecosystem/airtable.js
```diff
@@ -0,0 +1,134 @@
...
+
+export const processToolFromAirtable = async (data) => {
+  const config = await api.findGlobal(ECOSYSTEM_CONFIG, {});
+  const theme = {
+    en: data[config.toolTheme.english]?.[0],
+    pt: data[config.toolTheme.portuguese]?.[0],
+    fr: data[config.toolTheme.french]?.[0],
+  };
+  const description = {
+    en: data[config.toolsDescription.english] || "",
```
kilemensi, line 90: Why are we setting description to `""` but leaving theme to `undefined` (when not set in global that is)?

apps/charterafrica/src/lib/ecosystem/airtable.js
```diff
@@ -0,0 +1,134 @@
...
+  const homeCountry = data[config.homeCountry];
+  const defaultData = {
+    airtableId: data.id,
+    externalId: data[config.toolExternalId] || " ",
+    repoLink: data[config.toolRepoLink],
+    name: data[config.toolName],
+    link: data[config.toolLink],
+    operatingCountries,
+    contributors: contrib.map(({ id }) => id),
+    organisation: org[0]?.id,
```
kilemensi, line 110: Where are we checking to make sure we have at least 1 org and 1 contributor?

apps/charterafrica/src/lib/ecosystem/ecosystem.js
```diff
@@ -0,0 +1,73 @@
...
+} from "@/charterafrica/lib/ecosystem/models";
+import api from "@/charterafrica/lib/payload";
+
+export const updateEcosystemList = async (req, res) => {
+  const config = await api.findGlobal(ECOSYSTEM_CONFIG, {});
+  const args = {
+    baseId: config.airtableBase,
+    tableIdOrName: config.toolsTableName,
+  };
+  const response = await getListFromAirtable(args);
```
kilemensi, line 22: Are we returning the (whole) list or are we supposed to create collection items from this list?

apps/charterafrica/src/pages/api/v1/resources/ecosystem/[action].page.js
```diff
@@ -0,0 +1,29 @@
...
+  "update-ecosystem-content": updateEcosystemContent,
+};
+
+export default async function handler(req, res) {
+  const {
+    query: { action },
+  } = req;
+  const key = req.headers["x-api-key"];
+  if (!isApiKeyValid(key)) {
+    return res.status(403).json({ message: "INVALID API KEY" });
```
Comment (kilemensi) [inline apps/charterafrica/src/pages/api/v1/resources/ecosystem/[action].page.js:21]:
```suggestion
    return res.status(403).json({ message: "INVALID_API_KEY" });
```

apps/charterafrica/src/pages/api/v1/resources/ecosystem/proxy.page.js
```diff
@@ -0,0 +1,23 @@
+import fetchJson from "@/charterafrica/utils/fetchJson";
```
kilemensi, line 1: We can just have a schema action rather than this proxy. Our actions would then be

1. `list` if we use **PUT** method or `update-list` if we want to stick to **GET**.
2. `content` / `update-content`
3. `schema` (GET only. Again, this going to external service I feel like it should be protected but not sure how to pass the secret key to Payload without exposing it publicly)

PS: I think we can still use airtable.js to fetch bases and tables, no?

apps/charterafrica/src/payload/fields/airtableId.js
```diff
@@ -0,0 +1,8 @@
+const airtableId = () => ({
```
kilemensi, line 1: This is a normal text field. Why do we need a separate field for it?

apps/charterafrica/src/payload/fields/ecosystem/AirtableBaseSelect.js
```diff
@@ -0,0 +1,33 @@
...
+  return options;
+};
+
+export const validateSelect = async (value, { hasMany, required, t }) => {
+  const options = await getOptions();
+  try {
+    return select(value, { hasMany, options, required, t });
+  } catch (error) {
+    return select(value, { hasMany, options, required, t });
+  }
```
kilemensi, line 23: Same thing in try and catch? Not sure I follow.

apps/charterafrica/src/payload/fields/ecosystem/AirtableBaseSelect.js
```diff
@@ -0,0 +1,33 @@
...
+};
+
+function AirtableBaseSelect(props) {
```
kilemensi, line 26: Not a component, should be named like other functions i.e. `airtableBaseSelect`

apps/charterafrica/src/payload/fields/ecosystem/AirtableTableSelect.js
```diff
@@ -0,0 +1,44 @@
...
+import { select } from "payload/dist/fields/validations";
+import { createElement } from "react";
+import useSWR from "swr";
+
+import fetchJson from "../../../utils/fetchJson";
+
+const fetcher = (url) => fetchJson.get(url);
+const getOptions = async (baseId) => {
+  const url = `/api/v1/resources/ecosystem/proxy${baseId}/tables`;
+  const { bases } = await fetcher(url);
```
kilemensi, line 15:
```suggestion
const getOptions = async (baseId) => {
  const url = `/api/v1/resources/ecosystem/proxy${baseId}/tables`;
  const { bases } = await fetchJson.get(url);
```
Not sure fetcher adds anything here.

kilemensi (commented):
👍🏽 

----

Thinking about the CMS/i18n, I think we need to make i18n optional. This means the first group should probably ask for base and if 18n should be enabled.

If it should then for every field that we need l10n, we show the field. Otherwise, we just show default fields (Similar to how we ask if somebody wants to show documents in the DocumentsAndDatasets block).

apps/charterafrica/src/payload/fields/ecosystem/AirtableBaseSelect.js
koechkevin: it is a component @kilemensi only that it uses `React.createElement`.
Also uses React Hook `useSwr`

kilemensi: Yo! I didn't see that @koechkevin. Why are we mixing naming-conventions then `AirtableBaseSelect` vs `airtableColumnSelect`?

koechkevin: `airtableColumnSelect` is a function that builds a field but `AirtableBaseSelect` is only a field Component exported.
This informs the conventions

kilemensi: Check the code sir... there is `columnSelect` which I assume is the component that `airtableColumnSelect` function uses, no?

apps/charterafrica/src/lib/ecosystem/ecosystem.js
koechkevin: We are supposed to create/update a collection @kilemensi.
This part will be handled by PR #444

apps/charterafrica/app.json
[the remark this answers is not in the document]
koechkevin: Update content.
- Content source is airtable so update content implies getting data from airtable.

Update List.
- Pull data from github and update locally existing list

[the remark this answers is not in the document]
kilemensi: **List** is from airtable i.e. which contributors, organisations and tools we need to track.
**Content** is the actual details of an entity that we need to pull from sources such as GitHub. Content depends on List.

apps/charterafrica/src/lib/ecosystem/ecosystem.js
kilemensi: 🤔 ... I think creating collections should be done here @koechkevin . i.e. given the list of this contributors/organisations/tools, create the respective items in the CMS.

#444 is supposed to add flesh to these items but these items need to already exist in the CMS.

apps/charterafrica/src/payload/fields/ecosystem/AirtableBaseSelect.js
koechkevin: `columnSelect` That's a Higher Order Function that returns a component `ColumnSelect`

kilemensi: Man this code gets complex by the minute! No way I'm going to remember how all of this fits together if somebody wakes me up in a middle of the night three months from now.

There has to be a simpler way... Let me think on it for a minute.

apps/charterafrica/src/pages/api/v1/resources/ecosystem/proxy.page.js
koechkevin: Looks like Airtable.js doesn't have an API to list all bases unless we use REST API

kilemensi: 😱 ... that's just sad!

apps/charterafrica/src/payload/fields/ecosystem/AirtableBaseSelect.js
```diff
@@ -0,0 +1,28 @@
+import { Select } from "payload/components/forms";
```
kelvinkipruto, line 1: @koech is it necessary to create custom components for select? Can the default payload select field work, all that is needed is to fetch the options.

apps/charterafrica/src/payload/fields/ecosystem/airtableColumnSelect.js
```diff
@@ -0,0 +1,51 @@
...
+  };
+};
+
+const airtableColumnSelect = ({ tableField, baseField, ...overrrides }) => {
+  const defaults = {
+    required: true,
+    type: "text",
+    admin: {
+      description: () =>
+        "Enter Airtable Base above and Table to select a column from",
```
kelvinkipruto, line 43: Shouldn't this be translated?

koechkevin: Not sure how description accepts a locale param

koechkevin:
closed by #474

PR #435, opened 2023-05-31
Ft/add sourceAfrica documents to datasets page CMS setup

kelvinkipruto:
## Description
This PR implements the ability to add documents on the datasets page.
This change includes introducing a new field in the `datasets` block `showDocuments` which allows the configuration of the sourceAfrica group to fetch the documents.

Fixes
This is a partial fix for #441

## Type of change

## Screenshots
image
image

## Checklist:


koechkevin (approved):
LGTM

kelvinkipruto:
@kilemensi This is ready for review

kilemensi (commented):
👍🏽 

---

1. What's the thinking behind offering different options between datasets and documents?
   1. Why can I set document group but not dataset group?
   2. Why can I show/hide documents but I can't show/hide datasets?
Screenshot 2023-06-02 at 14 28 57
Screenshot 2023-06-02 at 14 29 06
2. Not all labels are common e.g. Read more/less, back to datasets, etc., seem to be applicable only to datasets and not documents
3. Data structure can also be improved a bit I believe. This is what I get:

```json
    {
      "commonLabels": {
        "readMore": "Read More",
        "readLess": "Read Less",
        "updated": "Updated",
        "created": "Created",
        "openDataset": "Open Dataset",
        "backToDatasets": "Back to Datasets",
        "seeMoreDatasets": "See More Datasets",
        "shareDataset": "Share Dataset"
      },
      "datasets": {
        "sortOptions": [
        ],
        "labels": {
          "search": "Search",
          "sort": "Sort",
          "countries": "Countries",
          "tags": "Tags"
        }
      },
      "documents": {
        "showDocuments": true,
        "documents": {
          "groupID": "the-charter-project",
          "options": {
          }
        },
        "sortOptions": [
        ],
        "labels": {
          "search": "Search",
          "sort": "Sort"
        }
      },
    }
```

But I think something like :point_down: would work better:

```json
    {
      "labels": {
          // If there are any common labels
      },
      "datasets": {
        "show": true,
        "organisation": {
          "groupID": "the-charter-project",
        },
        "filterBar": {
          "search": {
              label: "Search",
          },
          "sort": {
            "label": "Sort",
            "options": [
            ],
          },
          "countries": {
              label: "Countries",
          },
          "tags": {
            "labels":  "Tags"
          },
        },
        // datasets-only labels
        "labels": {
          "readMore": "Read More",
          "readLess": "Read Less",
          "updated": "Updated",
          "created": "Created",
          "openDataset": "Open Dataset",
          "backToDatasets": "Back to Datasets",
          "seeMoreDatasets": "See More Datasets",
          "shareDataset": "Share Dataset"
        },
      },
      "documents": {
        "show": true,
        "organisation": {
          "groupID": "the-charter-project",
          "options": {
          }
        },
        "filterBar": {
          "search": {
              label: "Search",
          },
          "sort": {
            "label": "Sort",
            "options": [
            ],
          },
        },
        // documents-only labels
        "labels: {
        }
      },
    }
```

kelvinkipruto:
> 1. Why can I set document group but not dataset group?

@kilemensi This is because the `datasets` group/organization is set in the `globals` section, but we can definitely let it be set in this block. One of the reasons why we set it in the `globals` was because of the flourish charts section where we had to develop a CRON job to update datasets statistics.

kilemensi:
> > 1. Why can I set document group but not dataset group?
> 
> @kilemensi This is ...

Thanks @kelvinkipruto ... Since we no longer need datasets statistics, removing it from global seems like the logical next step, no?

apps/charterafrica/src/payload/utils/documents.js
```diff
@@ -0,0 +1,6 @@
+export default function getDocumentsQuery(context, options) {
```
kilemensi, line 1: Why is this id `payload/utils` and not `utils`?

apps/charterafrica/src/lib/data/common/processPageDatasets.js
```diff
@@ -19,15 +21,21 @@ async function processSingleDataset(page, api, context) {
     return null;
   }
 
-  const { commonLabels = {} } = blocks.find(({ slug }) => slug === "datasets");
+  const {
+    datasets: { labels },
+    labels: commonLabels,
+  } = blocks.find(({ slug }) => slug === "datasetsAndDocuments");
```
kilemensi, line 27: What happens when there isn't such a block?

apps/charterafrica/src/lib/data/common/processPageDatasets.js
```diff
@@ -42,30 +50,77 @@ export default async function processPageDatasets(page, api, context) {
...
+          labels: {
+            ...commonLabels,
+            ...labels,
+          },
+          organizationId,
+        },
+      };
+    }
+
+    if (showDocuments) {
```
kilemensi, line 87: As we discussed, documents and datasets need to be different pages (see https://pesayetu.pesacheck.org/data).

... an alternative single page implementation is possible (e.g. /data page) but one would have to manage what "tab" is current active between documents and datasets, etc.

kilemensi (commented):
👍🏽

kelvinkipruto:
@kilemensi After reviewing Pesayetu, we may have to rethink this slightly. I thought they would all be on one page `/resources/datasets`. With the current implementation, we're managing which [tab is active](https://github.com/CodeForAfrica/ui/blob/83d96a185303dc307d0c06c3ab4a353de8d6d3ff/apps/charterafrica/src/components/Datasets/Datasets.js#L39)
I really like Pesayetu's approach where we can use `/resources/datasets` and `/resources/documents`. This will require that I refactor this a little bit but it should work well.

kilemensi:
> @kilemensi After reviewing ...

Cool @kelvinkipruto, I also like the PesaYetu approach. Let me let you cook. Ping me when it's ready for 👀 .

kilemensi:
QQ @kelvinkipruto : Looks like you've separated Datasets and Documents block. Where the tap/button switching info come from?

kelvinkipruto:
@kilemensi The buttons will switch between two pages `resources/datasets` and `/resources/documents`. This will also require the creation of two pages in the CMS with parent `resources`. The `/resources/datasets` page will continue to be handled as before, and `/resources/documents` will be handled by a new function [processPageDocuments](https://github.com/CodeForAfrica/ui/blob/a57a89ded8aa0dba9f4c03c5103338e3adda3baa/apps/charterafrica/src/lib/data/common/processPageDocuments.js).

kilemensi:
I meant as far as the CMS is concerned, where does the content team see this button that switches between datasets and documents pages @kelvinkipruto .

kelvinkipruto:
@kilemensi If in the CMS this has to be visible, then the best approach would be to manage the state of which tab is active instead of using different pages.

kelvinkipruto:
@kilemensi If in the CMS this has to be visible, then the best approach would be to manage the state of which tab is active instead of using different pages.

kilemensi:
> @kilemensi If in the CMS this has to be visible, then the best approach would be to manage the state of which tab is active instead of using different pages.

Why? In datasets block, you can have a show documents checkbox and if selected, we show a dropdown with a list of pages for editors to select the documents page to switch to. We do the opposite in documents block.

No state management necessary.

kelvinkipruto:
> In datasets block, you can have a show documents checkbox and if selected, we show a dropdown with a list of pages for editors to select the documents page to switch to. We do the opposite in documents block.

@kilemensi I'm a little bit lost here. Why would they need to select a page? The data required for both the `resources/datasets` and `resources/documents` are all in the `DatasetsAndDocuments` block.
Here are two approaches I'm thinking about:

1. (Current Implementation) In the CMS, the editors will create a `resources/datasets` page and add the `DatasetsAndDocuments` block as is currently. This allows them to configure both `datasets` and `documents` in a single block. This is then processed in the `processPageDatasets` function. The only issue I have with this is that the editor will have to create an empty `resources/documents` page in the CMS if we're to go with Pesayetu's approach and not the state-managed way. 
2. In the CMS, the user will have to create two pages, `resources/datasets` and `resources/documents`. We then split the `DatasetsAndDocuments` block into two, one for datasets and another for documents. These pages will then be processed independently of each other.

kilemensi:
> @kilemensi I'm a little bit lost here. Why would they need to select a page? The data required for both the...

Because clicking show documents/datasets will then load that page that they've selected @kelvinkipruto. You're assuming pages **have** to always be named `resources/datasets` and `resources/documents` but if we switch to block-based rendering, this block can be added into any page linking into any other page.

On the approach:
I prefer the stateless approaches i.e. everything you need is rendered on the server and you just get a page so approach 2 is more to my liking.

kelvinkipruto:
Ok @kilemensi I maybe get you a little bit
Here is how I'm interpreting it:
First, the editors create two pages in the CMS, say `datasets` and `documents`. These are the pages to be rendered when the user clicks the switch button.
They then create a third page and add the `DatasetsAndDocuments` block, from where they can select the two created pages as pages for datasets and documents.

kilemensi:
No @kelvinkipruto . Based on approach 2 you outlined above:

>In the CMS, the user will have to create two pages, resources/datasets and resources/documents. We then split the DatasetsAndDocuments block into two, one for datasets and another for documents. These pages will then be processed independently of each other.

1. We have a `Datasets` block with ability to link to documents via `showDocuments`, and
2. We have a `Documents` block with ability to link to datasets via `showDatasets`.

Content editors can choose to have these in any page that they want (for now of course it will be `/datasets` and `/documents` page until block-based rendering is fully implemented). The `Datasets` block can even be used in the consultation page by toggling `showDocuments` to `false`.

kelvinkipruto:
@kilemensi Yes. Now Cooking...

kelvinkipruto:
@kilemensi It is now ready for re-review

apps/charterafrica/src/payload/blocks/Documents.js
```diff
@@ -0,0 +1,265 @@
...
+                fr: "Page des ensembles de données",
+                pt: "Página de conjuntos de dados",
+              },
+              admin: {
+                condition: (_, siblingData) => siblingData?.showDatasets,
+              },
+              type: "relationship",
+              relationTo: "pages",
+              hasMany: false,
+              required: true,
```
kilemensi, line 256: `link`/`linkGroup` field already handles this.

kilemensi:
How can you say this is ready for review @kelvinkipruto when I still see [tabs](https://github.com/CodeForAfrica/ui/blob/35f6b19474d688a892d72b57e2031c70072c2eeb/apps/charterafrica/src/payload/blocks/Datasets.js#L22-L23) are still there?

Have you actually fully tested it or you expect others to do that for you?

kelvinkipruto:
@kilemensi I have tested it. I have used the `tabs` to separate the `datasets` and the pages to link to. See the above screenshots.

kilemensi:
> @kilemensi I have tested it. I have used the `tabs` to separate the `datasets` and the pages to link to. See the above screenshots.

Good thinking but (my thoughts are) tabs are for showing multiple **_independent_** things on the same window @kelvinkipruto. What we're going for is more of "would you like a coke with your fries?" i.e. if in documents, we're saying datasets would go well with documents. Hence they should be in the same panel, probably as a `group`.

... what do you think? (@koechkevin also if you have any thoughts).

kelvinkipruto:
@kilemensi I have removed the tabs.

apps/charterafrica/src/payload/blocks/Documents.js
```diff
@@ -0,0 +1,254 @@
...
+                        fr: "Options de tri",
+                        pt: "Opções de classificação",
+                      },
+                      fields: [
+                        {
+                          name: "value",
+                          type: "select",
+                          unique: true,
+                          required: true,
+                          options: documentSortOptions.map((value) => {
```
koechkevin, line 124: `options`  Can either be an array of strings, or an array of objects containing a label string and a value string.
So you could as well have this as 
```suggestion
                          options: documentSortOptions,
```

kilemensi (approved):
🚀

PR #443, opened 2023-06-05
Block based rendering

kilemensi:
## Description

Our current implementation kinds of cheat: Looks at a page and processes all blocks that it knows will more or less be in a page. This PR flips that process: process all blocks in a given page without caring which page it is.

**DO NOT MERGE**: This PR is up to kick-start discussions and it's not ready to merge just yet. Pages that currently use block-based rendering include:
 - Privacy policy,
 - Knowledge > News, and
 - About.

## Type of change

- [x] Chore (non-breaking change that improves code quality)

## Screenshots

N/A

## Checklist:


koechkevin (commented):
@kilemensi I try to access http://localhost:3000 I get an error 

image

kilemensi:
> @kilemensi I try to access http://localhost:3000 I get an error
> 

That's why it's not ready for merging just yet @koechkevin ... Pushed a fix for it.

apps/charterafrica/src/lib/data/common/index.js
```diff
@@ -124,7 +120,10 @@ export async function getPageProps(api, context) {
     return null;
   }
 
-  const [page] = pages;
+  let [page] = pages;
+  if (params?.slugs?.length === 3) {
+    page = await pagify(page, api, context);
```
koechkevin, line 125: There's a possibility of pagify returning `null`.

Setting page.blocks on null will throw an error

koechkevin (commented):
@kilemensi I have tested this approach and it looks good.

apps/charterafrica/src/lib/data/common/index.js
kilemensi: Nice catch. `null` would mean `notFound`. Let me push a fix.

kilemensi:
> @kilemensi I have tested this approach and it looks good.

Aaaight @koechkevin but what are you thoughts? Is this approach better that what we have? Can we improve the implementation, etc?

The idea is to get everyone's input?

apps/charterafrica/src/components/Articles/Articles.js
```diff
@@ -65,7 +66,14 @@ const Articles = React.forwardRef(function Articles(props, ref) {
   if (filtering && articlesRef.current) {
     articlesRef.current.scrollIntoView({ behavior: "smooth" });
   }
-  const res = useArticles(slug, { locale, q, sort, page, pageSize });
+  const res = useArticles(collection || slug, {
+    locale,
+    q,
```
saminegash, line 71: What is this variable q? the other variables represent the actual representation I believe but q is not clear?

apps/charterafrica/src/components/LongForm/LongForm.js
```diff
@@ -33,7 +33,7 @@ const LongForm = React.forwardRef(function LongForm(props, ref) {
         {content.map((c) => {
```
saminegash, line 33: using plurals for naming lists then while using the map method we can use the singular reference
e.g `contents.map((content) => { ...`

apps/charterafrica/src/lib/data/common/blockify/articles.js
```diff
@@ -0,0 +1,203 @@
...
+    locale,
+    where: {
+      _status: { equals: "published" },
+    },
+  });
+
+  if (!docs?.length) {
+    return null;
+  }
+  const tags = docs.flatMap((doc) => doc.tags).filter((tag) => tag?.slug);
```
saminegash, line 17: using typescript might help a lot with unsafe type issues here.

apps/charterafrica/src/lib/data/common/blockify/articles.js
```diff
@@ -0,0 +1,203 @@
...
+
+  return {
+    results,
+    totalPages,
+  };
+}
+
+const filtersLabelsPerLocale = {
+  en: {
+    "-publishedOn": "Most recent",
```
saminegash, line 125: separating translation data to it's own file might help reduce line of code a lot

apps/charterafrica/src/lib/data/common/blockify/articles.js
```diff
@@ -0,0 +1,203 @@
...
+      }
+    }
+  }
+  const foundArticles = await getArticles(page, api, context);
+  const tags = await getTags(page, api, context);
+  const filterLabels = filtersLabelsPerLocale[locale];
+  const { slug } = page;
+
+  // SWR fallback
+  let swrKey = `/api/v1/knowledge/${slug}`;
```
saminegash, line 169: base end points like this shall change time to time so it should be added to configuration files

saminegash (changes requested):

apps/charterafrica/src/components/Articles/Articles.js
kilemensi: `q` = `query` e.g. `https://www.google.com/search?q=code+for+africa` or `https://github.com/search?q=code+for+africa&type=repositories`

apps/charterafrica/src/components/LongForm/LongForm.js
kilemensi: 💯 ... except our CMS has chosen to call the page contents `content` so we're stuck with that for now.

apps/charterafrica/src/lib/data/common/blockify/articles.js
kilemensi: 💯 ... that's the next step: monorepo -> ci/cd -> typescript -> end-to-end testing. Happy to hear your suggestions as well.

apps/charterafrica/src/lib/data/common/blockify/articles.js
kilemensi: 💯 ... ideally all of these should be managed by/via the CMS but alas, we are still not sure where/how to really manage it in our CMS. For now, it's still a TODO.

apps/charterafrica/src/components/LongForm/LongForm.js
saminegash: What CMS are we using?

apps/charterafrica/src/lib/data/common/blockify/articles.js
saminegash: you mean we are going to convert our codebase based on this steps

saminegash:
I think I cannot say more as I am just looking into the codebases and diving in I believe.

apps/charterafrica/src/components/LongForm/LongForm.js
kilemensi: [Payload CMS](https://payloadcms.com/)... Speaking of CMS, I'd also like to hear you views and experiences with various CMSes.

apps/charterafrica/src/lib/data/common/blockify/articles.js
kilemensi: Yes-ish.

1. We've moved about 3 apps into this monorepo (charter.africa, codeforafrica and promisetracker). We still have a few to go (pesayetu, trolltracker, etc.)
2. We've already implemented CI/CD for charter.africa and codeforafrica (you can see the workflows in the repo).
3. We have included Jest (for component testing) and Playwright (for end-to-end testing) but we're not really using them that much (just snapshot testing w/ Jest).

So the idea is come up with a comprehensive plan that will allow us to move everything UI into this repo and make sure we have ci/cd -> typescript -> end-to-end testing for all the apps.

kilemensi:
Your review and thoughts @kelvinkipruto

kilemensi:
> @kilemensi ... the [`pagify`](https://github.com/CodeForAfrica/ui/blob/b91555657503622b482737f761ee5935dd46e201/apps/charterafrica/src/lib/data/common/pagify/index.js#L7) function means some block will still be processed per page?
>

`pagify` is for those pages that **do not exist** in the CMS @kelvinkipruto. For example single news item page. `pagify` creates the block(s) that should be displayed in this page but doesn't process/render the page. The processing/rendering still happens in `blockify`.

Not sure if there is anything we can do about these sort of pages but any ideas are welcomed.

> I like how [`blockify`](https://github.com/CodeForAfrica/ui/blob/b91555657503622b482737f761ee5935dd46e201/apps/charterafrica/src/lib/data/common/blockify/index.js#L23) works, and I think it's a better way than our current method.
>
👍🏽 

>One issue I have faced though is when I add a `news` block to any that does not have a parent page, it breaks. ( I will investigate further) but given this approach, shouldn't it be processed regardless of the page it is on?
>
Yes, `news` block should be processed regardless of which page it's on. If it doesn't work, do investigate and report. It's most likely an edge case we haven't considered.

kilemensi:
Taking this back to draft to focus on current PROD issues.

kilemensi:
This was implemented in codeforafrica, etc. Since charterafrica is stable at the moment, there is no need for this PR... unless we're doing v2.0

PR #449, opened 2023-06-13
Ecosystem block

### LINKED ISSUE #447: @charterafrica CMS Ecosystem block
Implement the CMS Block that can be used to configure how we pull contributors, organisations and tools from an externally managed Airtable base.

The expected data structure from the block should be similar to the one shown below:
```jsonc
{
  "baseId": "",
  "localized": false,
  // show schema once baseId is set
  "schema": {
    "contributorTableId": "",
    // show TableColumns once TableId is set
    "contributorTableColumns": {
      // To have standard structure, lets use slug to mean username, etc.
      "slug": "",
      "source": {
        // Now we only support GitHub but this is what we'll check to know where to pull data from
        // We can check the full url as backup?
        "type": "github",
        "url": ""
      },
      "name": "",
      "description": {
        "en": "",
        // other locales if localized: true
      },
      "role": "",
      "url": ""
    },
    "organisationTableId": "",
    "organisationTableColumns": {
      "slug": "",
      "source": {
        "type": "github",
        "url": ""
      },
      "name": "",
      "description": { "en": "" },
      "type": "",
      "role": "",
      "url": "",
      "tools": []
    },
    "toolTableId": "",
    "toolTableColumns": {
      "slug": "",
      "source": {
        "type": "github",
        "url": ""
      },
      "name": "",
      "description": { "en": "" },
      "theme": { "en": " " },
      "url": "",
      "contributors": [],
    }
  }
}
```

Depends on #445

koechkevin:
## Description

Fixes #447 

## Type of change

## Screenshots
image

## Checklist:


koechkevin:
@kelvinkipruto please review this

kelvinkipruto:
@koechkevin I'm having some errors saving the `Ecosystem` global

kelvinkipruto (approved):
LGTM
Perhaps the descriptions can be updated to match the actual fields, and not be repetitive.

koechkevin:
@kilemensi

apps/charterafrica/src/lib/ecosystem/airtable.js
```diff
@@ -0,0 +1,15 @@
...
+export async function airtableSchema(req) {
+  const { url } = req.query;
+  const headers = {
+    "Content-Type": "application/json",
+    Authorization: `Bearer ${process.env.AIRTABLE_API_TOKEN}`,
+  };
+  const response = await fetchJson.get(`https://api.airtable.com/v0${url}`, {
+    headers,
+  });
+  return response;
```
kilemensi, line 14:
```suggestion
  return fetchJson.get(`https://api.airtable.com/v0${url}`, {
    headers,
  });
```

apps/charterafrica/src/pages/api/v1/resources/ecosystem/[action].page.js
```diff
@@ -0,0 +1,43 @@
+import { airtableSchema } from "@/charterafrica/lib/ecosystem/airtable";
+
+const isApiKeyValid = (key) => {
+  return key && key === process.env.RESOURCES_SECRET_TOKEN;
+};
+
+export async function schema(req, res) {
+  const { source } = req.query;
+  const sourceMap = { airtable: airtableSchema };
```
Comment (kilemensi) [inline apps/charterafrica/src/pages/api/v1/resources/ecosystem/[action].page.js:9]: Why do we need `airtable` as source? Is there any other schema we'll be exposing here?

apps/charterafrica/src/pages/api/v1/resources/ecosystem/[action].page.js
```diff
@@ -0,0 +1,43 @@
...
+const actionMap = {};
+
+export default async function handler(req, res) {
+  const {
+    query: { action },
+  } = req;
+  // Because we do not need to expose API KEY
+  if (action === "schema") {
+    return schema(req, res);
+  }
```
Comment (kilemensi) [inline apps/charterafrica/src/pages/api/v1/resources/ecosystem/[action].page.js:29]: Use the `actionMap` above.

kilemensi (commented):
👍🏽

apps/charterafrica/src/pages/api/v1/resources/ecosystem/[action].page.js
Comment (koechkevin) [reply apps/charterafrica/src/pages/api/v1/resources/ecosystem/[action].page.js:29]: Are we going to share SECRET_KEY on payloadCMS admin console?

Comment (kilemensi) [reply apps/charterafrica/src/pages/api/v1/resources/ecosystem/[action].page.js:29]: May be I'm missing something... context please.

apps/charterafrica/src/pages/api/v1/resources/ecosystem/[action].page.js
```diff
@@ -0,0 +1,43 @@
...
+  if (!isApiKeyValid(key)) {
+    return res.status(403).json({ message: "INVALID_API_KEY" });
+  }
+  const actionFunc = actionMap[action];
+  if (actionFunc) {
+    const response = await actionFunc(req, res);
+    if (response) {
+      return res.status(200).json(response);
+    }
+    return null;
```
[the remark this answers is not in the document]
Comment (koechkevin) [reply apps/charterafrica/src/pages/api/v1/resources/ecosystem/[action].page.js:40]: Just  a workaround for some eslint error, every action block must have a return statement

apps/charterafrica/src/pages/api/v1/resources/ecosystem/[action].page.js
Comment (kilemensi) [reply apps/charterafrica/src/pages/api/v1/resources/ecosystem/[action].page.js:29]: I think this just makes the code a bit harder to follow @koechkevin ... Since every `actionFunc` receives `req`, they can check if valid key exists if they need to protect the `action`.

apps/charterafrica/src/payload/fields/ecosystem/airtableColumnSelect.js
```diff
@@ -0,0 +1,78 @@
...
+    );
+
+    return createElement(Select, { ...props, options });
+  };
+};
+
+function airtableColumnSelect({ tableField, overrides = {} }) {
+  const defaultSelect = {
+    type: "text",
+    validate: validateColumnSelect(tableField),
```
koechkevin, line 64: @kilemensi we have a very interesting limitation on this.

it is being called on save for every field therefore exceeding airtable api limit. We may need to find a way of caching airtable response.

kilemensi: 🤯 ... makes sense though @koechkevin .

My initial thinking is we should have an hidden field, `baseData` or `baseTables` where by once a user has selected the base, we make a single API request and store the tables in this field. The rest of the fields in `schema` would then refer to this data i.e. siblingData.baseTables instead of making API requests.

i.e. only one extra request for every change in base selection.

koechkevin: > 🤯 ... makes sense though @koechkevin .
> 
> My initial thinking is we should have an hidden field, `baseData` or `baseTables` where by once a user has selected the base, we make a single API request and store the tables in this field. The rest of the fields in `schema` would then refer to this data i.e. `siblingData.baseTables` instead of making API requests.
> 
> i.e. only one extra request for every change in base selection.

Looks like the payloadcms api doesn't offer a way to mutate sibling data inside field hooks.

Tried in memory cache that expires after 5 minutes and it looks to work.

i.e 
```js
const cache = {};

function getFromCache(key) {
  const rawData = cache[key];
  if (rawData) {
    const { lastUpdated, value } = rawData;
    const fiveMinutesAgo = new Date();
    fiveMinutesAgo.setMinutes(fiveMinutesAgo.getMinutes() - 5);
    if (fiveMinutesAgo < lastUpdated) {
      return value;
    }
  }
  return null;
}

export async function schema(req) {
  const { url } = req.query;
  const headers = {
    "Content-Type": "application/json",
    Authorization: `Bearer ${process.env.AIRTABLE_API_TOKEN}`,
  };
  if (getFromCache(url)) {
    return getFromCache(url);
  }
  const value = fetchJson.get(`https://api.airtable.com/v0${url}`, {
    headers,
  });
  cache[url] = { value, lastUpdated: new Date() };
  return value;
}
```

kilemensi: Cool, lets implement this approach for now @koechkevin but I think the approach of may be storing cache data in the public folder is may be best moving forward (we do this in the codeforafrica app).

Create an issue to track this so that we can revisit once we've deployed current implementation

apps/charterafrica/src/payload/fields/ecosystem/sourceField.js
```diff
@@ -0,0 +1,37 @@
...
+        tableField,
+        overrides: {
+          name: "url",
+          label: { en: "Source URL", fr: "URL de source", pt: "URL da fonte" },
+        },
+      }),
+      airtableColumnSelect({
+        tableField,
+        overrides: {
+          name: "type",
```
kilemensi, line 21: Do we have the type column in the Airtable? I'm assuming this will be github, gitlab, etc., and we only support github for now.

kilemensi (approved):
🚀

apps/charterafrica/src/payload/fields/ecosystem/sourceField.js
koechkevin: @kilemensi  yes.

Can either be
For-Profit, Non-Profit, Academics & Research etc.

So we can't fetch this from github

kilemensi: Unless I'm missing something, I'm asking about `source` type @koechkevin. Not `organisation` type.

koechkevin: @kilemensi.
My thinking is,

There will be 1 Airtable base to collect this data.
So data team will input all the data in the tables then have a column tthat specifies source of data in each row

koechkevin: I had asked Keziah to add this column.

PR #458, opened 2023-06-22
Fix ecosystem block

koechkevin:
## Description

Ecosystem global in the CMS is still extremely slow and makes hundreds of API calls (takes more than 20sec for everything to complete and the page to render). 

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/apps/charterafrica/src/payload/fields/ecosystem/airtableBaseSelect.js
+++ b/apps/charterafrica/src/payload/fields/ecosystem/airtableBaseSelect.js
@@ -8,6 +8,15 @@ import fetchJson from "../../../utils/fetchJson";
 
 const url = `${process.env.PAYLOAD_PUBLIC_APP_URL}/api/v1/resources/ecosystem/schema?source=airtable&url=/meta/bases`;
 
+export function getUrl(baseId) {
+  if (baseId) {
+    return `${process.env.PAYLOAD_PUBLIC_APP_URL}/api/v1/resources/ecosystem/schema?source=airtable&url=/meta/bases/${baseId}/tables`;
+  }
+  return null;
+}
+
+export const schema = {};
+
 function basesToOptions(bases) {
   return bases?.map((item) => ({ value: item.id, label: item.name })) || [];
 }
@@ -18,6 +27,11 @@ const getOptions = async () => {
 };
 
 const validateBaseSelect = async (value, { hasMany, required, t }) => {
+  if (value) {
+    const link = getUrl(value);
+    const { tables } = await fetchJson.get(link);
+    schema.tables = tables;
+  }
   const options = await getOptions();
   return select(value, { hasMany, options, required, t });
 };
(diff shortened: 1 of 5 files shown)


apps/charterafrica/src/payload/fields/ecosystem/airtableBaseSelect.js
```diff
@@ -8,6 +8,15 @@ import fetchJson from "../../../utils/fetchJson";
 
 const url = `${process.env.PAYLOAD_PUBLIC_APP_URL}/api/v1/resources/ecosystem/schema?source=airtable&url=/meta/bases`;
 
+export function getUrl(baseId) {
```
kilemensi, line 11: 1. Why do we have `url` and getUrl? Seems like `getUrl` should probably be `getTablesUrl` (or similar)
2. `getTablesUrl` (or similar) can reuse `url` and just append `/${baseId}/tables`, no?

apps/charterafrica/src/payload/fields/ecosystem/airtableBaseSelect.js
```diff
@@ -18,6 +27,11 @@ const getOptions = async () => {
 };
 
 const validateBaseSelect = async (value, { hasMany, required, t }) => {
+  if (value) {
+    const link = getUrl(value);
+    const { tables } = await fetchJson.get(link);
+    schema.tables = tables;
+  }
   const options = await getOptions();
   return select(value, { hasMany, options, required, t });
```
kilemensi, line 36:
```suggestion
  const options = await getOptions();
  const valid = select(value, { hasMany, options, required, t });
  if (valid) {
    const link = getUrl(value);
    const { tables } = await fetchJson.get(link);
    schema.tables = tables;
  }
  return valid;
```

1. We should set schema only when we are sure it's a valid selection
2. Do we need to load bases every time? Can't we cache this too i.e. only load it the first time.

apps/charterafrica/src/payload/fields/ecosystem/airtableColumnSelect.js
```diff
@@ -6,16 +6,8 @@ import {
...
 
-import fetchJson from "../../../utils/fetchJson";
-
-function getUrl(baseId) {
-  if (baseId) {
-    return `${process.env.PAYLOAD_PUBLIC_APP_URL}/api/v1/resources/ecosystem/schema?source=airtable&url=/meta/bases/${baseId}/tables`;
-  }
-  return null;
-}
+import { schema as globalSchema } from "./airtableBaseSelect";
```
kilemensi, line 10: Rather than importing schema from `airtableBaseSelect` which tightly couples `airtableBaseSelect` to `airtableColumnSelect`, can't we pass `schema` to `airtableColumSelect` similar to how we're passing `tableField`?

kilemensi (commented):
👍🏽

kilemensi (approved):
LGTM!

PR #459, opened 2023-06-23
Refactor ecosystem backend

### LINKED ISSUE #457: @charterafrica Ecosystem backend structure
## Description

While the current implementation seems to work fine, the code behind it seems to be a bit  difficult to follow and reason about. This means it will most likely be very hard to remember how everything fit together a few months from now.

This issue proposes a structure that we should try to adopt if we all agree. The proposal tries to stay very close to the current implementation so that changes are minimal and therefore we can move faster.

### Structure

```
  - lib
      - ecosystem
          - airtable
              - index.js
          - ecosystem.js
          - index.js
          - github
              - index.js
          - gitlab
          - ...
          - payload
```
`ecosystem/airtable` will contain all the necessary functionality for us to interact with Airtable for ecosystem purposes. `ecosystem/airtable/index.js` will expose the 3 main functionality we need: list bases, schema, and data (all data in a given base).

1. list bases is to be used in the CMS Ecosystem global section to select the correct base to use for ecosystem data,
2. schema is to be used to query tables and fields for the selected schema above, and
3. data is to be used on the application side when fetching the list of entities to track in the CMS/show on the front-end.

`ecosystem/payload` will contain all the necessary functionality we need to store, update, and retrieve entities in the CMS. This should be built on top of the general `lib/payload` implementation. `ecosystem/payload/index.js` will expose only the necessary functions we need.

The structure should be clear and this is what `ecosystem/github`, `ecosystem/gitlab`, and any other ecosystem integration should follow. One thing to note is if there is any field mapping necessary, this should be done within that respective integration. For example, if ecosystem/ecosystem.js expects a tool to have a name, then ecosytem/airtable should make sure it returns data with name in tool. Same with ecosystem/github, it should convert whatever field GitHub uses to name since that's what `ecosystem/ecosystem.js` expects. Hopefully, we'll be able to keep the mappings to a minimum.

This way `ecosystem/ecosystem.js` defines the interface that all these integration needs to strictly follow.

`ecosystem/index.js` will expose the 2 main functionality required from the ecosystem: `update list` and `update content`. Nothing less, nothing more. How this functions actually work will be in `ecosystem/ecosystem.js`.

update list should start by first checking when the last update was done and if it's still considered fresh (less than 30 mins ago), it should probably just skip. Otherwise, it should do a full refresh using data from Airtable.

update content should do similar thing (but fresh for update content is probably an hour or two) and do a full refresh from GitHub, GitLab, etc.

This proposal also suggests we should return objects rather than individual methods when exposing functionality. For example

```js
import airtable from "@/charterafrica/lib/ecosystem/airtable"

const bases = airtable.bases();
...
const schema = airtable.schema(baseId); // or airtable.schema(config)
...
const data = airtable.data(baseId, schema); // or airtable.data(config)
```

is much easier to work with that
```js
import { getAirTableBases, listSchemaFromAirtable} from "@/charterafrica/lib/ecosystem/airtable"

const bases = getAirTableBases();
...
const schema = listSchemaFromAirtable(baseId); // or listSchemaFromAirtable(config)
...
```

### API

Single `api/v1/resources/ecosystem/entities` end-point but accessible via 2 methods: `POST` and `PUT`

`POST` will create or update the entities list (the current `update-list` end-point). `PUT` will only update the content of the entities list if it already exists (the current `update-content` end-point).

koechkevin:
## Description

Refactors Ecosystem backend

Fixes #457

## Type of change

## Checklist:


kilemensi (approved):
🚀

kilemensi (commented):
:+1:

---

Aren't we returning data in the format ecosystem expects/can use directly? From the issue:

>One thing to note is if there is any field mapping necessary, this should be done within that respective integration. For example, if ecosystem/ecosystem.js expects a tool to have a name, then ecosytem/airtable should make sure it returns data with name in tool. Same with ecosystem/github, it should convert whatever field GitHub uses to name since that's what ecosystem/ecosystem.js expects. Hopefully, we'll be able to keep the mappings to a minimum.
>
>This way ecosystem/ecosystem.js defines the interface that all these integration needs to strictly follow.

koechkevin:
> 👍
> 
> Aren't we returning data in the format ecosystem expects/can use directly? From the issue:
> 
> > One thing to note is if there is any field mapping necessary, this should be done within that respective integration. For example, if ecosystem/ecosystem.js expects a tool to have a name, then ecosytem/airtable should make sure it returns data with name in tool. Same with ecosystem/github, it should convert whatever field GitHub uses to name since that's what ecosystem/ecosystem.js expects. Hopefully, we'll be able to keep the mappings to a minimum.
> > This way ecosystem/ecosystem.js defines the interface that all these integration needs to strictly follow.

I expect this file to only have fetching functionalities .
@kilemensi  To map data to format in which ecosystem expects, I intend to have another file `processDataFromAirtable`

kilemensi:
> I expect this file to only have fetching functionalities . @kilemensi To map data to format in which ecosystem expects, I intend to have another file `processDataFromAirtable`
>

Where will this processDataFromAirtable file leave and which part of the code will be responsible for call it @koechkevin?

koechkevin:
> > I expect this file to only have fetching functionalities . @kilemensi To map data to format in which ecosystem expects, I intend to have another file `processDataFromAirtable`
> 
> Where will this processDataFromAirtable file leave and which part of the code will be responsible for call it @koechkevin?

Resides inside ecosystem folder.
- Exports processContributor, processTool and processOrganisation (Takes in an airtable tableRow and returns data as expected by the ecosystem).
- To be imported in `ecosystem/index.js`

kilemensi:
> > > I expect this file to only have fetching functionalities . @kilemensi To map data to format in which ecosystem expects, I intend to have another file `processDataFromAirtable`
> > 
> > 
> > Where will this processDataFromAirtable file leave and which part of the code will be responsible for call it @koechkevin?
> 
> Resides inside ecosystem folder.
> 
> * Exports processContributor, processTool and processOrganisation (Takes in an airtable tableRow and returns data as expected by the ecosystem).
> * To be imported in `ecosystem/index.js`

I want us to move away from this approach and move to the feature-based approach `ecosystem/airtable` should contain all the functionality of working with data from airtable; not just fetching, but also processing it into a format useful to the rest of the system. Other parts of ecosystem should know as little as possible about the inner workings/structure of airtable data.

kilemensi:
> @kilemensi We will then be forced to export contributors, tools, organisations separately because of relationships as opposed to data
>
Could be @koechkevin. Or it could be data method pulls raw data, process it, and return an object like
```js
  {
    contributors: [/* list of contributors*/],
    organisations: [/* list of organisations*/],
    tools: [/* list of tools*/],
  }
```
So, it's still `data` but now containing processed entities.

koechkevin (commented):
@kilemensi [This is how I have handled processing data from airtable](apps/charterafrica/src/lib/ecosystem/airtable/processData.js).

However, at this point, nothing has been created in the database so tools will have 0 contributors and organisation will have 0 tools.

koechkevin:
> > @koechkevin ?
> 
> I am about to make a commit that refactors `ecosystem.js`

@kilemensi

koechkevin:
@kilemensi

PR #466, opened 2023-06-27
Refactor Airtable

koechkevin:
## Description

This is to allow Review of Airtable refactor
Fixes # (issue)

## Type of change

## Checklist:


koechkevin:
@kilemensi,

- Created another branch [chore/main-refactor](https://github.com/CodeForAfrica/ui/tree/chore/main-refactor)
- Granular changes will be merged onto this branch for easy review.
- It is expected for builds to fail at this point until all changes are merged

apps/charterafrica/src/lib/ecosystem/airtable/index.js
```diff
@@ -0,0 +1,113 @@
...
+      toolTableId,
+      organisationTableId,
+      contributorTableId,
+      partnersTableId,
+      socialMediaTableId,
+    },
+  } = config;
+  const airtableTools = await table(baseId, toolTableId);
+  const airtableContributors = await table(baseId, contributorTableId);
+  const airtableOrganisations = await table(baseId, organisationTableId);
```
kilemensi, line 76:
```suggestion
  const toolsData = await table(baseId, toolTableId);
  const contributorsData = await table(baseId, contributorTableId);
  const organisationsData = await table(baseId, organisationTableId);
```

apps/charterafrica/src/lib/ecosystem/utils.js
```diff
@@ -0,0 +1,24 @@
+export function localizeData(data = {}) {
```
kilemensi, line 1: Where is this file being used?

apps/charterafrica/src/lib/ecosystem/airtable/processData.js
```diff
@@ -0,0 +1,194 @@
+import * as Sentry from "@sentry/nextjs";
+
+function getter(data, key) {
+  return key ? data?.[key] : null;
+}
```
kilemensi, line 5:
```suggestion
function getValue(data, key, defaultValue=null) {
  return data?.[key] ?? defaultValue
}
```

But I think typing `getter(fields, name)` or `getValue(fields, name)` is not shorter that just `fields?.[name] ?? null`

apps/charterafrica/src/lib/ecosystem/airtable/processData.js
```diff
@@ -0,0 +1,194 @@
...
+
+function mapSupportersToFields(supporters, config, tableData) {
```
kilemensi, line 7:
```suggestion
function mapSupporterIdsToObjects(supporterIds, config, { partnersData }) {
```

Since it's raw data from Airtable (i.e. contains fields, etc.), I think we should call it `partnersData` rather than just partners to emphasise it hasn't been processed.

apps/charterafrica/src/lib/ecosystem/airtable/processData.js
```diff
@@ -0,0 +1,194 @@
...
+      const { fields } = partners.find((item) => id === item.id);
+      return {
+        name: getter(fields, name),
+        website: getter(fields, url),
+        logo: getter(fields, logo),
+      };
+    } catch (error) {
+      Sentry.captureMessage(error.message);
+      return null;
+    }
```
kilemensi, line 24:
```suggestion
    const partnerData = partners?.find((item) => id === item.id);
    if (partner) {
      const { fields } = partnerData;
      return {
        name: getter(fields, name),
        website: getter(fields, url),
        logo: getter(fields, logo),
      };
    } else {
      return null;
    }
```

Not sure why we need the try/catch

apps/charterafrica/src/lib/ecosystem/airtable/processData.js
```diff
@@ -0,0 +1,194 @@
...
+    const theme = {
+      en: getter(data, toolTableColumns.theme.en)?.[0],
+      pt: getter(data, toolTableColumns.theme.pt)?.[0],
+      fr: getter(data, toolTableColumns.theme.fr)?.[0],
+    };
+    const description = {
+      en: getter(data, toolTableColumns.description.en),
+      pt: getter(data, toolTableColumns.description.pt),
+      fr: getter(data, toolTableColumns.description.fr),
+    };
```
kilemensi, line 64: I'd have gone with something like this... Only process data that's going to be used.

```suggestion
    const { localized } = config;
    const locales = localized ? ['en', 'fr', 'pt'] : ['en'];
    const theme = locales.reduce((curr, acc) => {
      acc[curr] = getter(data, toolsTableColumns.theme[curr])?.[0];
      return acc;
    }, {};
    const description = locales.reduce((curr, acc) => {
      acc[curr] = getter(data, toolsTableColumns.description[curr]);
      return acc;
    }, {};
```

apps/charterafrica/src/lib/ecosystem/airtable/processData.js
```diff
@@ -0,0 +1,194 @@
...
+      partners,
+      homeCountry,
+      socialMedia,
+      theme,
+      description,
+    };
+  } catch (error) {
+    Sentry.captureMessage(error.message);
+    return null;
+  }
```
kilemensi, line 105: We're not making any external calls, not sure where we expect these errors to come from.

apps/charterafrica/src/lib/ecosystem/airtable/processData.js
```diff
@@ -0,0 +1,194 @@
...
+      fr: getter(data, contributorTableColumns.description.fr),
+    };
+    return {
+      airtableId: data.id,
+      avatarUrl: getter(data, contributorTableColumns.avatarUrl)?.[0]?.url,
+      externalId: getter(data, contributorTableColumns.slug),
+      repoLink: `https://github.com/${getter(
+        data,
+        contributorTableColumns.slug
+      )}`,
```
kilemensi, line 130: you should use source field to figure out if it's GitHub repo or not.

apps/charterafrica/src/lib/ecosystem/airtable/index.js
```diff
@@ -0,0 +1,113 @@
...
+  const socialMedia = await table(baseId, socialMediaTableId);
+  const partners = await table(baseId, partnersTableId);
+  const tableData = { socialMedia, partners };
+  const tools = airtableTools.map((item) =>
+    processTool(
+      {
+        ...item.fields,
+        id: item.id,
+      },
```
kilemensi, line 85:
```suggestion
    processTool(
      item,
```

All process functions know this is Airtable data. I'm sure they know about id and fields, no?

kilemensi (commented):
👍🏽

apps/charterafrica/src/lib/ecosystem/utils.js
koechkevin: It is anticipated to be used in `ecosystem`.

apps/charterafrica/src/lib/ecosystem/airtable/processData.js
koechkevin: I think it is important to catch errors at every level to avoid termination if subsequent processes @kilemensi

apps/charterafrica/src/lib/ecosystem/utils.js
koechkevin: payload api does not allow create/update with data in the form 
```js
{
   description: {
    en: "English",
    pt: "Portuguese",
    fr: "French"
   }
}
```
Data should look like

```js
 createCollection(collection, toCreate, { locale })
```
So this function will be used to structure data in this format @kilemensi

apps/charterafrica/src/lib/ecosystem/airtable/processData.js
kilemensi: No, it's not important to catch errors @koechkevin, it's important to handle errors. It's more important to understand why errors are thrown and if possible, write code that doesn't run into exerptions.

Hence, if we've pulled all data from Airtable already, what kind of errors can be thrown here?

apps/charterafrica/src/lib/ecosystem/airtable/processData.js
```diff
@@ -0,0 +1,188 @@
...
+    socialMedia,
+    tools,
+    description,
+  };
+  if (!commonData.externalId) {
+    const message = `Missing external ID for ${data.id}`;
+    Sentry.captureMessage(message);
+    return null;
+  }
+  if (!tools.length) {
```
kilemensi, line 182: Can't these be checked at the top so we skip the whole processing?

kilemensi (commented):
👍🏽

kilemensi (approved):
:rocket:

PR #468, opened 2023-06-27
Chore/payload-refactor

koechkevin:
## Description
Refactors models (Payload)
Fixes # (issue)

## Type of change

## Checklist:


apps/charterafrica/src/lib/ecosystem/payload.js
```diff
@@ -0,0 +1,70 @@
+import * as Sentry from "@sentry/nextjs";
+
+import api from "../payload";
```
kilemensi, line 3:
```suggestion
import api from ""@/charterafrica/lib/payload";
```

apps/charterafrica/src/lib/payload/constants.js
```diff
@@ -0,0 +1,4 @@
+export const ORGANIZATION_COLLECTION = "organisations";
```
kilemensi, line 1: Put this in `src/payload/utils` or somewhere in `src/payload`: `lib` uses `payload` but `payload` shouldn't depend on `lib`.

apps/charterafrica/src/lib/ecosystem/utils.js
```diff
@@ -0,0 +1,24 @@
+export function localizeData(data = {}) {
```
kilemensi, line 1: What is the structure of `data`? I was expecting keys would be `en`/`pt`/`fr`.

kilemensi (commented):
👍🏽

apps/charterafrica/src/lib/ecosystem/utils.js
koechkevin: I have changed this to 

```js
 {
 [key]: {
  en: "",
  pt: "",
  fr: ""
}
}
``` @kilemensi

Comment (kilemensi) [reply apps/charterafrica/src/lib/ecosystem/utils.js:1]: I'm still not 100% sure I get it @koechkevin :
```js
        ...result,
      en: {
        ...result.en,
        [key]: data[key].en || data[key],
      },
```

Why do we need .`..result` and `...result.en` if our data is in `data`?

My understanding is: The output of airtable module is data in the following format: For a field that supports localization, for example `description`, 
```js
  {
    name: "",
    description: {
      en: "",
    }
  }
```
if `localized` is `false`, and:
```js
  {
    name: "",
    description: {
      en: "",
      fr: "",
      pt: ""
    }
  }
```
If `localized` is `true`.

How does this method deal with the above?

apps/charterafrica/src/lib/ecosystem/payload.js
```diff
@@ -0,0 +1,91 @@
...
+        airtableId: { equals: toCreate.airtableId },
+      },
+    });
+    if (docs.length) {
+      const data = await api.updateCollection(
+        collection,
+        docs[0]?.id,
+        {
+          ...toCreate,
+          updatedAt: new Date(),
```
kilemensi, line 20: My understanding Collections automatically supports `createdAt` and `updatedAt` unless specifically disabled via [`timestamps`](https://payloadcms.com/docs/configuration/collections#options) option. Why are we manually updating `updatedAt`?

apps/charterafrica/src/lib/ecosystem/payload.js
```diff
@@ -0,0 +1,91 @@
...
+          Sentry.captureMessage(error.message);
+        }
+      })
+    );
+  } catch (e) {
+    Sentry.captureMessage(e.message);
+  }
+}
+
+export async function getCollectionPerAirtableId(collection, ids) {
```
kilemensi, line 79: 1. Where is this used?
2. Looks like it returns `collectionId` and not actual `collection` so `getCollection` naming is not 100% correct.

apps/charterafrica/src/lib/ecosystem/payload.js
```diff
@@ -0,0 +1,91 @@
...
+        },
+        { locale }
+      );
+      return data;
+    }
+    const data = await api.createCollection(collection, toCreate, { locale });
+    return data;
+  } catch (e) {
+    Sentry.captureMessage(e.message);
+    return {};
```
kilemensi, line 30: Shouldn't we just return `null`?

apps/charterafrica/src/lib/ecosystem/utils.js
koechkevin: @kilemensi this method is intended to convert data from 
```js
{
 description: {
  en: "",
  fr: "",
  pt: ""
}
...
}
```

to 

```js
{
 en: {
 description : ""
},
pt: {
 description : ""
},
fr: {
 description : ""
},
}
```

Just because payload doesn't allow create or update with objects on a text field

apps/charterafrica/src/lib/ecosystem/payload.js
koechkevin: @kilemensi  This function is intended to be used to link tools to organisation.

from airtable data we have tools = [...airtableId] but we need to store in collection as [...ids]

kilemensi: Cool @koechkevin but again, if it's not being used NOW, don't include it in current PR but include it in the PR that actually uses the functionality. [YAGNI](https://martinfowler.com/bliki/Yagni.html)

apps/charterafrica/src/lib/ecosystem/utils.js
kilemensi: Cool @koechkevin , so there 1 use of this function. But what about those `...result`? Lets take a full example. If my data from Airtable is as 👇🏽, how should I transform it so that Payload accepts it?
```js
{
  name: "Koetch",
  description: {
    en: "Regular Dev",
    fr: "Dev ordinaire",
    pt: "Desenvolvedor regular",
  },
  languages: ["JavaScript", "CSS"],
}
```

koechkevin: @kilemensi 
```js
{
  en: {
    name: "Koetch",
    description: "Regular Dev",
    languages: ["JavaScript", "CSS"],
  },
  pt: {
    name: "Koetch",
    description: "Desenvolvedor regular",
    languages: ["JavaScript", "CSS"],
  },
  fr: {
    name: "Koetch",
    description: "Dev ordinaire",
    languages: ["JavaScript", "CSS"],
  },
}
```

kilemensi: Then I think you need to pass the locales as well @koechkevin , based on if config.localize is enabled (similar to how we handled it in Airtable). Haven't tested the code but I think something like 👇🏽 should do it
```js
function localize(data, locales) {
    return locales.reduce((localized, locale) => {
        localized[locale] = {};
        Object.keys(data).reduce((localeData, key) => {
            localeData[key] = data[key]?.[locale] ?? data[key];
            return localeData;
        }, localized[locale]);
        return localized;
    }, {});
}
```

PS: I also see `delocalize` function that is not used anywhere, and can't think of why we would need it.

koechkevin: @kilemensi this also works fine

koechkevin: @kilemensi this also works fine

kilemensi: Do you mean `localizeData` also works fine @koechkevin ?

I don't doubt it but:
1. It creates more data than needed (e.g. if `localized` if `false`, we only need `en` but it will create for `fr` and `pt`)
2. It has too much/unnecessary code duplication. The only thing that changes between this nine lines is `en`, `fr`, and `pt`: https://github.com/CodeForAfrica/ui/pull/468/files#diff-a49b86971ff1a23df48a437020db40fb15e8fb658fbdfa5e0f877409e28dc850R14-R25. It can be rewritten in 3 lines or less.

The above suggestion addressed these 2 issues... now, leave me alone! It's my off day.

apps/charterafrica/src/lib/ecosystem/payload.js
```diff
@@ -0,0 +1,75 @@
...
+        docs[0]?.id,
+        toCreate,
+        { locale }
+      );
+      return data;
+    }
+    const data = await api.createCollection(collection, toCreate, { locale });
+    return data;
+  } catch (e) {
+    Sentry.captureMessage(e.message);
```
kilemensi, line 26:
```suggestion
    Sentry.captureException(e.message);
```

apps/charterafrica/src/lib/ecosystem/payload.js
```diff
@@ -0,0 +1,75 @@
...
+    const localizedData = localize(toCreate || {}, locales);
+    if (!localizedData) {
+      return null;
+    }
+    if (!localized) {
+      return updateOrCreate(collection, localizedData?.en);
+    }
+    const promises = Object.keys(localizedData).map((key) =>
+      updateOrCreate(collection, localizedData?.[key], key)
+    );
```
kilemensi, line 43:
```suggestion
    // All collections must have airtableId
    if (!toCreate?.airtableId) {
      return null;
    }
    const locales = localized ? ["en", "pt", "fr"] : ["en"];
    const dataPerLocale = localize(toCreate, locales);
    const promises = Object.keys(dataPerLocale).map((locale) =>
      updateOrCreate(collection, dataPerLocale.[locale], locale)
    );
```

Since we've only initialised `locales` based on `localized`, we no longer need to worry about it when calling `updateOrCreate`.

apps/charterafrica/src/lib/ecosystem/payload.js
```diff
@@ -0,0 +1,75 @@
...
+      toDelete.map(async ({ id }) => {
+        try {
+          await api.updateCollection(collection, id, {
+            deletedAt: new Date(),
+          });
+        } catch (error) {
+          Sentry.captureMessage(error.message);
+        }
+      })
+    );
```
kilemensi, line 71:
```suggestion
    Promise.all(
      toDelete.map(async ({ id }) => {
        try {
          await api.updateCollection(collection, id, {
            deletedAt: new Date(),
          });
        } catch (error) {
          Sentry. captureException(error.message);
        }
      })
    );
```

If we're catching the error and sending it to Sentry (which is correct), how do we tell the promise to reject?

apps/charterafrica/src/lib/ecosystem/utils.js
```diff
@@ -0,0 +1,23 @@
+function localize(data, locales = ["en"]) {
+  const getValue = (value, locale) => {
+    if (value?.[locale] !== undefined) {
+      return value?.[locale];
+    }
+    if (value?.en !== undefined) {
+      return value?.en;
+    }
+    return value;
+  };
```
kilemensi, line 10: If I'm reading this function correctly, you're returning en data if say pt data is missing, correct?

If so, is it the right approach? If `localized` is enabled in Airtable, then it's up to the content/data team to ensure all locales have data, no?

apps/charterafrica/src/payload/utils/constants.js
```diff
@@ -0,0 +1,4 @@
+export const ORGANIZATION_COLLECTION = "organisations";
```
kilemensi, line 1: I'm not a fun of functional-based grouping (`utis/constant`s, `utils/functions`, etc.). I prefer feature-based e.g. `utils/collections`, etc.

kilemensi (commented):
👍🏽

apps/charterafrica/src/lib/ecosystem/payload.js
koechkevin: @kilemensi do we need to terminate the whole process because a deletion failed?

Thinking we can only skip if an error occurred during deletion and have it reported in sentry

apps/charterafrica/src/lib/ecosystem/utils.js
koechkevin: Remember this function from `airtable`.

```js
 const locales = localized ? ["en", "fr", "pt"] : ["en"];
  const theme = locales.reduce((curr, acc) => {
    acc[curr] = getValue(data, toolTableColumns.theme[curr])?.[0] ?? null;
    return acc;
  }, {});
  const description = locales.reduce((curr, acc) => {
    acc[curr] = getValue(data, toolTableColumns.description[curr]);
    return acc;
  }, {});
```

means if lacalised is false, object will look like

```js
{
 description: {en: ""},
 theme: {en: ""}
}
```
 so we need to sanitize this as well @kilemensi

kilemensi: Yes, but if localized was false there, it will be false here as well @koechkevin

apps/charterafrica/src/lib/ecosystem/payload.js
kilemensi: 1. Aren't we going to end up with wrong data?
2. If we need to skip, then `Promise.allsettled` is what we should be using, no?

koechkevin: Perhaps promise.allSettled can work here

apps/charterafrica/src/lib/ecosystem/utils.js
kilemensi: @koechkevin

apps/charterafrica/src/lib/ecosystem/payload.js
kilemensi: But what are we going to do about the failed deletes?

apps/charterafrica/src/lib/ecosystem/utils.js
koechkevin: @kilemensi If I understood you I have this updated as 

```js
function localize(data, locales = ["en"], localized = false) {
  const getValue = (value, locale) => {
    if (value?.[locale] !== undefined) {
      return value?.[locale];
    }
    if (value?.en !== undefined && !localized) {
      return value?.en;
    }
    return value;
  };

  return locales.reduce((loc, locale) => {
    const localizedData = loc;
    localizedData[locale] = {};
    Object.keys(data).reduce((locData, key) => {
      const localeData = locData;
      localeData[key] = getValue(data[key], locale);
      return localeData;
    }, localizedData[locale]);
    return localizedData;
  }, {});
}
```

kilemensi: Why are we returning en value if say pt value is undefined @koechkevin? This is what I'm asking?

The point of `localized` is to make sure we're reading and writing only the data we need. If `localized` is enabled in Airtable, then `fr` and `pt` columns *MUST* contain data. Otherwise we only need `en` data.

koechkevin: @kilemensi  We are returning `en` only if localized is `false`

_If localized is enabled in Airtable, then fr and pt columns MUST contain data._ Correct(For only fields that are localisable).

apps/charterafrica/src/lib/ecosystem/utils.js
```diff
@@ -0,0 +1,21 @@
+function localize(data, locales = ["en"]) {
+  const getValue = (value, locale) => {
+    if (value?.[locale] !== undefined) {
+      return value?.[locale];
+    }
+    return value;
```
kilemensi, line 6:
```suggestion
    return value?.[locale] ?? value;
```

kilemensi (approved):
🚀

apps/charterafrica/src/lib/ecosystem/utils.js
koechkevin: @kilemensi  I am iffy about this.

```js
{
 theme: { pt: null }
}
```

will result in something like

```js
{
 pt: {  theme: { pt: null } }
}
```

kilemensi: Yes @koechkevin, `??` only handles `undefined`. If you want to handle `undefined` _and_ `null`, repalce it wil `||`

PR #469, opened 2023-06-27
Flourish Charts Setup

### LINKED ISSUE #467: @charterafrica EmbeddedChart
## Description
Using https://help.flourish.studio/article/145-how-to-embed-flourish-charts-in-your-cms as guide and example, implement the EmbeddedChart CMS block and front-end function per designs

While it looks like title and subtitle would come from the CMS with the embedded chart, please confirm with the design time.

## Screenshot.

Image

kelvinkipruto:
## Description

This PR adds the ability to embed Flourish charts.

Fixes #467

## Type of change

## Screenshots
image
image

## Checklist:


apps/charterafrica/src/payload/blocks/FlourishCharts.js
```diff
@@ -0,0 +1,29 @@
...
+  },
+  fields: [
+    {
+      name: "code",
+      label: {
+        en: "Embed Code",
+        fr: "Code d'intégration",
+        pt: "Código de incorporação",
+      },
+      type: "textarea",
```
kilemensi, line 23: I think it should be [`code`](https://payloadcms.com/docs/fields/code) with `html` as language

apps/charterafrica/src/components/FlourishCharts/FlourishCharts.js
```diff
@@ -0,0 +1,29 @@
...
+        backgroundColor: "#fff",
+      }}
+      ref={ref}
+    >
+      <Section
+        sx={{
+          px: { xs: 5, sm: 0 },
+          py: { xs: 5, md: 7.5 },
+          width: "100%",
+          height: "500px",
```
kilemensi, line 20: `backgroundColor`, `height` and `width` should be passed in via `props`.

kilemensi (commented):
:+1:

---

1. Are we using any Flourish specific features? If not, lets call this component EmbeddedChart rather than FlourishChart since it should be reusable for embedding any chart.
2. Did you confirm about title/subtitle?

koechkevin (approved):

kilemensi:
> 👍
> 
> 1. Are we using any Flourish specific features? If not, lets call this component EmbeddedChart rather than FlourishChart since it should be reusable for embedding any chart.
> 2. Did you confirm about title/subtitle?

Did you see the above questions @kelvinkipruto? I don't see where either of them has been addressed.

kelvinkipruto:
Hello @kilemensi I have renamed the components as suggested.
As for the title and subtitle, it would be great to have them provide some context on the charts, so I have retained them.
The only issue is with the switch buttons in the chart. Since datasets and documents will be on different pages, each can have its own charts, and no need to switch between them.

kilemensi:
> Hello @kilemensi I have renamed the components as suggested. As for the title and subtitle, it would be great to have them provide some context on the charts, so I have retained them.
>

We're talking about different things here @kelvinkipruto. I asked you to confirm with the design team whether all embedded charts will include titles or titles will be entered via the CMS when embedding? The main reason being localization. Does Flourish support localization? If yes, have we integrated with it? If no, what's the plan? Are we going to show an English chart when someone is browsing the site in Portuguese?

> The only issue is with the switch buttons in the chart. Since datasets and documents will be on different pages, each can have its own charts, and no need to switch between them.
>
Are you talking about the switch button in the embedded chart itself? If yes, I'm assuming that's just for illustration purposes. We already handle switching via show documents/datasets filter bar.

kelvinkipruto:
@kilemensi Flourish charts support adding titles and subtitles. It looks like it does not support localization out of the box. 
>  Are we going to show an English chart when someone is browsing the site in Portuguese?

Since don't have control over the embedded code, the simplest solution would be to use already localized data when creating the charts and set the `html` as localized in the CMS allowing for different embeds for each language. The only downside to this is if there is a large amount of data involved and the effort to translate them.

kilemensi:
Do we know the way forward @kelvinkipruto ?

kelvinkipruto:
@kilemensi Yes. Titles/subtitles in the CMS, then different flourish charts for each locale. Updating it in a few.

kelvinkipruto:
@kilemensi Ready for re review

apps/charterafrica/src/payload/blocks/EmbeddedChart.js
```diff
@@ -0,0 +1,92 @@
+import { validateHexColor } from "../utils/colors";
+
+const EmbeddedChart = {
+  slug: "embedded-chart",
+  labels: {
+    singular: {
+      en: "Flourish Charts",
```
kilemensi, line 7: Embedded Chart

apps/charterafrica/src/components/EmbeddedChart/EmbeddedChart.js
```diff
@@ -0,0 +1,40 @@
...
+        <Typography variant="h5SemiBold" sx={{ mb: 1 }} color="neutral.dark">
+          {title}
+        </Typography>
+        <Typography variant="p1" sx={{ mb: 1 }} color="neutral.dark">
+          {subtitle}
+        </Typography>
+        <div
+          style={{
+            width: `${width}%`,
+            height: `${height}px`,
```
kilemensi, line 31: Units should probably be set with prop i.e. height should come in a "500px" or "10vh", etc.

apps/charterafrica/src/payload/blocks/EmbeddedChart.js
```diff
@@ -0,0 +1,92 @@
...
+      localized: true,
+    },
+    {
+      name: "html",
+      label: {
+        en: "Embed Code",
+        fr: "Code d'intégration",
+        pt: "Código de incorporação",
+      },
+      type: "textarea",
```
kilemensi, line 47: Why did you just ignore https://github.com/CodeForAfrica/ui/pull/469#discussion_r1244806021?

kilemensi (commented):
👍🏽

apps/charterafrica/src/payload/blocks/EmbeddedChart.js
kelvinkipruto: 😮

kilemensi:
Looking at the designs, do the charts have the same `height` across various screen sizes @kelvinkipruto ?

kelvinkipruto:
@kilemensi Flourish charts allow you to set the height/width before publishing a chart, it then handles its responsiveness—no need for us to update it manually.

kilemensi:
You're manually setting the `width` and `height` [here](https://github.com/CodeForAfrica/ui/blob/1d5be36b5b280638736e9fab1d62b00061174f86/apps/charterafrica/src/components/EmbeddedChart/EmbeddedChart.js#L30-L31) @kelvinkipruto. It has nothing to do with Flourish.

kelvinkipruto:
Yes @kilemensi When publishing a chart in flourish charts, you set its height and width. These are the dimensions we will set in the CMS
image

kilemensi:
What I mean is this @kelvinkipruto :

1. We set height to 400px in desktop. Cool.
2. What happens when we switch to mobile? The height will still remain 400px, no?

kelvinkipruto:
@kilemensi Flourish charts only allow us to export one height. I have found out that it does handle responsiveness quite well.
It does allow us to set some height settings, but I'm not quite sure how to handle them on the front end.
image

One quick hack can be updating the size on the fly: for example when exporting the embed HTML, the code is as follows:
`<div class=".." data-src=".." data-height="400px"><script src="..."></script></div>`
We can maybe run some regex to update the `data-height` depending on the screen size.

kilemensi:
Before we start looking into _hacks_ @kelvinkipruto, what is the _proper_ way of handling this?

1. Are there breakpoints similar to ours? If not can we set the breakpoints to match hours?
2. Does the responsiveness maintain aspect ratio so that if given height for desktop, we can calculate aspect ratio and maintain that for mobile?

kelvinkipruto:
@kilemensi There is just one breakpoint, between desktop and mobile, and it is only available for certain charts. Most chart types do not allow you to update the breakpoint. 
All of them allow the setting of automatic height.
My observation so far is if, we give the chart a specific height, and set it to automatic height, the responsiveness handles quite well for all the breakpoints. The height will however stay the same across all breakpoints.

kelvinkipruto:
@kilemensi Not sure I get you correctly, but are you referring to the (width/height) of the chart itself?
If so, you can customize both to a specific value. We can then use it to calculate. By default, the width is 100% of its parent element.

kilemensi:
My understanding of the problem is this @kelvinkipruto : We can set `height` at say desktop screen size to make sure the chart fits well with the full width of the screen. The problem comes when a user views the chart on smaller screen: while the chart resizes to fit the new width, the `height` remains the same.

Or goal it to find a way to adjust the height at various screen sizes. I'm I right so far?

kilemensi:
> @kilemensi You are right. Q.A: How would we adjust the height using the aspect ratio 🤔??

CSS already supports [`aspect-ratio`](https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio). We only need to ask for _actual_ chart `width` and `height` at desktop size and we can compute aspect-ratio from that I guess

kelvinkipruto:
@kilemensi After some back-and-forth testing, here is what I have managed to figure out. Once you set the chart height in flourish charts, say set it to `400px`, the chart will maintain this height regardless of screen size. The only way to adjust its size is by updating the attribute `data-height="400px"` to match the size we want. Using CSS to update it will cause it to overlap with other elements.

Also not setting the height in the flourish charts causes it to be very large. 

So the only way I have found to work is by updating the `data-height` attribute.

kelvinkipruto:
> So you're saying flourish chart doesn't respect parent div's height?

@kilemensi Yes. 
I have tried [updating the data-height attribute](https://github.com/CodeForAfrica/ui/pull/469/commits/8387ab61d806ce1b1f1ca9d31fd66dd0ef2a64ba), and it works only on desktop, on resize you have to reload it to work.

kilemensi:
This is straight from the doc I've shared on the issue @kelvinkipruto :

>We recommend using our script embed because it adjusts its height dynamically to fit its contents. This means your chart won't appear flattened. You can change the width and height settings by clicking More options prior to exporting.

kelvinkipruto:
@kilemensi The reason we need to give it a specific height is for it to match our desktop design height. If we leave it at the default height(auto) it is very big and covers too much of the space available.

kilemensi:
> > @kilemensi The r....
> 
> So the `height` isn't related to the content @kelvinkipruto ?

And for completeness sake, what I mean is may be you're testing this using a very big chart. Have you tried embedding charts of various sizes?

kelvinkipruto:
@kilemensi I have tested it with two types of charts.

1. The default one by flourish when you publish the chart. This turns out to be very big and will cover a large area than in our designs.
2. One with updated height (using more options). This works very well and we can match our design. The only problem is that the height set will remain across all screen sizes. But the responsiveness of the chart is quite good.

kilemensi (approved):
🚀 

---

Lets let Flourish dictate `height` for now. If that's not what designers want, we can revisit the implementation.

PR #471, opened 2023-06-27
Chore/ecosystem-refactor

### LINKED ISSUE #457: @charterafrica Ecosystem backend structure

koechkevin:
## Description

This PR Refactors ecosystem and ensures Update from airtable works.

To test out this PR make a POST request to http://localhost:3000/api/v1/resources/ecosystem/entities

Fixes #457

## Type of change

## Checklist:


apps/charterafrica/src/pages/api/v1/resources/ecosystem/[action].page.js
```diff
@@ -1,25 +1,49 @@
...
 
+const apiKeyExcludedActions = ["schema", "bases"];
+
 export default async function handler(req, res) {
   const {
     query: { action },
   } = req;
   const key = req.headers["x-api-key"];
-  if (!isApiKeyValid(key) && action !== "schema") {
+  if (!isApiKeyValid(key) && !apiKeyExcludedActions.includes(action)) {
```
Comment (kilemensi) [inline apps/charterafrica/src/pages/api/v1/resources/ecosystem/[action].page.js:46]: Looks like we only need this key in `entities`. Can't we move the check for key there?

apps/charterafrica/src/payload/fields/ecosystem/airtableBaseSelect.js
```diff
@@ -6,11 +6,11 @@ import useSWR from "swr";
 
 import fetchJson from "../../../utils/fetchJson";
 
-const baseUrl = `${process.env.PAYLOAD_PUBLIC_APP_URL}/api/v1/resources/ecosystem/schema?source=airtable&url=/meta/bases`;
+const baseUrl = `${process.env.PAYLOAD_PUBLIC_APP_URL}/api/v1/resources/ecosystem/bases`;
```
kilemensi, line 9:
```suggestion
const baseUrl = `${process.env.PAYLOAD_PUBLIC_APP_URL}/api/v1/resources/ecosystem/`;
```

Then basesUrl becomes baseUrl/bases, schemaUrl becomes baseUrl/schema?...

kilemensi (approved):
🚀

PR #474, opened 2023-07-03
Update Ecosystem Content

koechkevin:
## Description

This PR fetches Github API to update Ecosystem content

Fixes # 457

## Type of change

## Checklist:

### DIFF
--- a/apps/charterafrica/src/lib/ecosystem/ecosystem/index.js
+++ b/apps/charterafrica/src/lib/ecosystem/ecosystem/index.js
@@ -5,6 +5,9 @@ import {
   processContributors,
   processTools,
   processOrganisations,
+  updateContributor,
+  updateTool,
+  updateOrganisation,
 } from "@/charterafrica/lib/ecosystem/ecosystem/processData";
 import api from "@/charterafrica/lib/payload";
 import { ECOSYSTEM_GLOBAL } from "@/charterafrica/payload/utils/collections";
@@ -30,5 +33,12 @@ export async function updateList() {
 }
 
 export async function updateContent() {
+  async function execute() {
+    const contributors = await updateContributor();
+    const organisations = await updateOrganisation();
+    const tools = await updateTool();
+    return { tools, contributors, organisations };
+  }
+  execute();
   return { message: "PROCESS_STARTED" };
 }
--- a/apps/charterafrica/src/lib/ecosystem/ecosystem/processData.js
+++ b/apps/charterafrica/src/lib/ecosystem/ecosystem/processData.js
@@ -1,8 +1,14 @@
 import {
+  tool,
+  organisation,
+  contributor,
+} from "@/charterafrica/lib/ecosystem/github";
+import {
   createCollection,
   bulkMarkDeleted,
   getCollectionIdsPerAirtableId,
 } from "@/charterafrica/lib/ecosystem/payload";
+import api from "@/charterafrica/lib/payload";
 import {
   ORGANIZATION_COLLECTION,
   CONTRIBUTORS_COLLECTION,
@@ -12,8 +18,10 @@ import {
 export async function processContributors(airtableData, config) {
   const { contributors } = airtableData;
   await bulkMarkDeleted(CONTRIBUTORS_COLLECTION, contributors);
-  const toProcess = airtableData?.contributors?.map((item) => {
-    return createCollection(CONTRIBUTORS_COLLECTION, item, config);
+  const toProcess = airtableData?.contributors?.map(async (item) => {
+    const gitData = await contributor(item);
+    const toCreate = { ...item, gitData };
+    return createCollection(CONTRIBUTORS_COLLECTION, toCreate, config);
   });
   return Promise.allSettled(toProcess);
 }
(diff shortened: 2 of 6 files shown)


apps/charterafrica/src/lib/ecosystem/ecosystem/index.js
```diff
@@ -5,6 +5,9 @@ import {
   processContributors,
   processTools,
   processOrganisations,
+  updateContributor,
```
kilemensi, line 8: If these new methods are updateXXXX, should we call the old ones createXXXX? Or what does processXXXX stand for?

apps/charterafrica/src/lib/ecosystem/ecosystem/index.js
```diff
@@ -30,5 +33,12 @@ export async function updateList() {
 }
 
 export async function updateContent() {
+  async function execute() {
+    const contributors = await updateContributor();
+    const organisations = await updateOrganisation();
+    const tools = await updateTool();
+    return { tools, contributors, organisations };
+  }
+  execute();
```
kilemensi, line 42: Don't we need to time this as well?

kilemensi (commented):
👍🏽 

---

Just had a quick look but I see we have methods in ecosystem/github saving data to payload, etc. Are you sure we're following https://github.com/CodeForAfrica/ui/issues/457?

apps/charterafrica/src/lib/ecosystem/ecosystem/index.js
koechkevin: processXXX processes data from Airtable together with github data using airtableId and github ID then stores this data in payload.
updateXXX on the other hand fetches data for every item from github and updates the collection accordingly.

kilemensi: I know what they do @koechkevin ... I'm talking about naming. To add, I'm ok with updateXXXX. Isn't:
>processXXX processes data from Airtable together with github data using airtableId and github ID then stores this data in payload.
>
Just mean createXXXX or prepareXXXX?

kilemensi:
> > 👍🏽
> > Just had a quick look but I see we have methods in ecosystem/github saving data to payload, etc. Are you sure we're following #457?
> 
> * @kilemensi no method in `github.js` is saving to payload. We are simply passing update boolean to tell if we need to fetch entirely from github or we can use what we already have.
>

I mean why does `github` module need to even know that `payload` module exists @koechkevin? Isn't that the job of the `ecosystem` module?

kilemensi:
I see we've switched to REST for contributors and organization but we're still fetching tools via GraphQL. But the tool query still include contributors and organization @koechkevin

koechkevin:
> I see we've switched to REST for contributors and organization but we're still fetching tools via GraphQL. But the tool query still includes contributors and organization @koechkevin

Retained graphql for tools because of languages and last commit. Using REST would have required  an extra 2 API calls to get them.

I will remove contributors and organization in the query @kilemensi.

apps/charterafrica/src/lib/ecosystem/ecosystem/processData.js
```diff
@@ -1,24 +1,76 @@
...
+  const { airtableId } = airtableData;
+  const { docs } = await api.getCollection(ORGANIZATION_COLLECTION, {
+    where: {
+      airtableId: { equals: airtableId },
+    },
+  });
+  if (docs.length) {
+    return docs[0];
+  }
+  return github.fetchOrganisation(airtableData);
```
kilemensi, line 43: I'm still not sure about this @koechkevin:

1. If we're updating, fetch fresh data from GitHub (Called from `updateContent`). ✅ 
2. If we're not updating, I'm assuming we're creating/preparing data i.e. called from `updatedList`. Yet, if `airtableId` is missing from collection, we're again fetching from GitHub. ❓❓❓ . Should we query GitHub **ONLY** for entities that we are sure are in the CMS collection already?

kilemensi (commented):
👍🏽

apps/charterafrica/src/lib/ecosystem/ecosystem/processData.js
koechkevin: @kilemensi 
The reason  why I had it this way is 
When `updateList` is called and the  data already exists in the collection, then we only use it to link during preparation. No need to make an API call at this point.

koechkevin: Do you suggest that we don't make any github api call from `updateList` only to be done by `updateContent`?

kilemensi: Yes... `updateList` will be called more often than `updateContent` I believe.

koechkevin: Yes,

But if `updateList` is called and new Items are added then it means for the period between calling `updateList` and `updateContent` the newly added items will have github data missing @kilemensi

kilemensi: That's alright for now... we have different API limits on Airtable and GitHub so we can't assume just because we can pull from Airtable, we can also pull from GitHub @koechkevin .

As long as missing data doesn't break the system, we'll experiment & figure out what's the best way to call these 2 methods.

kilemensi (approved):
🚀

PR #477, opened 2023-07-06
Fix creating multiple pages failing

### LINKED ISSUE #473: BUG: Can't create multiple pages on a new database
**Describe the bug**
When running a clean installation of `charterafrica`, you can only create one page. Trying to create a second page fails with error:
```
{
    "errors": [
        {
            "name": "ValidationError",
            "message": "The following field is invalid: blocks.filterBar.sort.options.value",
            "data": [
                {
                    "message": "Value must be unique",
                    "field": "blocks.filterBar.sort.options.value"
                }
            ]
        }
    ]
}
```

**To Reproduce**
Steps to reproduce the behaviour:

1. Switch to the main branch
2. Update the `MONGO_URL` environment variable to point to a clean database
3. Run `charterafrica` and log in to the admin
4. Try creating more than 1 page

**Expected behaviour**
You should be able to create more than 1 page.

kelvinkipruto:
## Description
This PR fixes a bug that prevents the creation of multiple pages in a new DB.

Fixes #473 

## Type of change

## Checklist:

### DIFF
--- a/apps/charterafrica/src/payload/blocks/Datasets.js
+++ b/apps/charterafrica/src/payload/blocks/Datasets.js
@@ -128,7 +128,6 @@ const Datasets = {
                       {
                         name: "value",
                         type: "select",
-                        unique: true,
                         required: true,
                         options: sortOptions,
                         validate: (val, options) => {
--- a/apps/charterafrica/src/payload/fields/documentCloudFilterBar.js
+++ b/apps/charterafrica/src/payload/fields/documentCloudFilterBar.js
@@ -80,7 +80,6 @@ function documentCloudFilterBar({ overrides } = {}) {
                 {
                   name: "value",
                   type: "select",
-                  unique: true,
                   required: true,
                   options: documentSortOptions,
                   validate: (val, options) => {


kilemensi:
Why/How did `unique: true` prevent creating multiple pages @kelvinkipruto?

kelvinkipruto:
@kilemensi According to the [docs](https://payloadcms.com/docs/fields/select), adding `unique` to `select` will:
> Enforce that each entry in the Collection has a unique value for this field.

I found removing this would fix the issue #473

kelvinkipruto:
Not sure about it on your side, but can you reproduce issue #473?

kilemensi:
> @kilemensi According to the [docs](https://payloadcms.com/docs/fields/select), adding `unique` to `select` will:
> 
> > Enforce that each entry in the Collection has a unique value for this field.
> 
> I found removing this would fix the issue #473

Yes but we don't even have that block in a new page @kelvinkipruto or am I misunderstanding the issue?

kelvinkipruto:
@kilemensi Perhaps its an issue on my side alone, can you try and see if you get the same result as me for issue #473

kelvinkipruto:
> Yes but we don't even have that block in a new page @kelvinkipruto or am I misunderstanding the issue?

@kilemensi Yes, even without adding any block to any page, it would just fail.

kilemensi (approved):
🚀

koechkevin (approved):

PR #483, opened 2023-07-11
Setup pesayetu in monorepo

kelvinkipruto:
## Description
This PR experiments with moving [PesaYetu](https://github.com/CodeForAfrica/pesaYetu) into this integrated monorepo

Fixes # (issue)

## Type of change

## Checklist:


apps/pesayetu/.eslintrc.js
```diff
@@ -1,4 +1,7 @@
 module.exports = {
   root: true,
   extends: ["eslint-config-commons-ui/next"],
+  rules: {
+    "no-undef": "off",
+  },
```
kilemensi, line 6: Why?

kilemensi (commented):
👍🏽 

---

This PR is too huge for me to test on my current connection.

kelvinkipruto:
@kilemensi This looks ok, but there is one issue, our [custom theme](https://github.com/CodeForAfrica/ui/blob/b5a36badda335fe0196869b7a97bbeefd3cae0f7/apps/pesayetu/src/theme/index.js) is not being [applied](https://github.com/CodeForAfrica/ui/blob/3835e8a524f7dc9689e3661433bde98dcfcb9ea5/apps/pesayetu/src/pages/_app.js#L47), and it instead uses the default theme. It works well on the original [Pesayetu](https://github.com/CodeForAfrica/PesaYetu/pull/229).

kilemensi:
> @kilemensi This looks ok, but there is one issue, our ...

Have you reviewed and made sure the theme here is created and applied the say way as say in @/charterafrica app @kelvinkipruto ?

kelvinkipruto:
@kilemensi Yes. The creation of the themes looks similar. I think the only difference is that in `charterafrica`, it is handled from the new `@commons-ui/core` in the monorepo. For Pesayetu we're  still using the older [`@commons-ui/core`](https://github.com/CodeForAfrica/COMMONS-UI).

kilemensi:
> @kilemensi Yes. The creation of the themes looks similar...

So how do the two @commons-ui/core packages co-exists @kelvinkipruto ?

kilemensi:
... and do we need to use the old `@commons-ui/core` if we've already upgraded it to the latest MUI?

kelvinkipruto:
> So how do the two @commons-ui/core packages co-exists

For pesayetu we're getting it from [npm](https://github.com/CodeForAfrica/ui/blob/3835e8a524f7dc9689e3661433bde98dcfcb9ea5/apps/pesayetu/package.json#L35).

kelvinkipruto:
> and do we need to use the old `@commons-ui/core` if we've already upgraded it to the latest MUI?

Yes @kilemensi There are some components in the old one that are required and are not in the new one.

kelvinkipruto:
> There are some components in the old one that are required and are not in the new one.

They are not that many of them. I guess we can recreate them as components to get rid of the older `@commons-ui`

kilemensi:
If it works on the old repo but not here @kelvinkipruto then my first assumption would be the two `@commnos-ui/core`. If we can get rid of the old one that would be the way to go.

My advice: test it
1. Remove the old `@commons-ui/core` and point everything to the new `@common-ui/core`.
2. Some components will fail but try disabling them and see if the rest works correctly e.g. the theme is applied correctly.
3. If point 2 above is true, then create issues to track those failing component that you can update on separate PR(s).
4. If point 2 above is not true, then we dig further to figure out what is going on.

kilemensi:
... and @kelvinkipruto ?

kelvinkipruto:
Further changes i.e [migration to emotion](#548 ) to be handled in a new PR

kilemensi (approved):

PR #486, opened 2023-07-12
Rename collection to item

### LINKED ISSUE #485: Collection rename and migrations
[MONGOOSE] Warning: `collection` is a reserved schema pathname and may break some functionality.

Consider using another name e.g `item`.
Add a plan for migration of production data to new name

koechkevin:
## Description
[MONGOOSE] Warning: `collection` is a reserved schema pathname and may break some functionality.
Fixes #485 

The following commands  were added
- Generate migrations file `pnpm migrate-mongo create {name_of_migrations}`.
- Migrate `turbo run db-migrate --filter=charterafrica`
- Rollback `turbo run db-rollback --filter=charterafrica`

## Type of change

## Checklist:


kilemensi:
@koechkevin, when you say:
>We will need to update database column i.e
>

When/How are we supposed to be doing that? That's the crux of what you were supposed to figure out.

For example, did you look at the Payload migration [example](https://github.com/payloadcms/migrate-mongo-example)?

koechkevin:
> 



> @koechkevin, when you say:
> 
> > We will need to update database column i.e
> 
> When/How are we supposed to be doing that? That's the crux of what you were supposed to figure out.
> 
> For example, did you look at the Payload migration [example](https://github.com/payloadcms/migrate-mongo-example)?

@kilemensi I thought the scope of this task was only to have the name collection renamed and the existing data updated but I can update this PR as per the link you shared. I see it works the same way as _Sequelize_ normally handles migrations.

kilemensi:
> @kilemensi I thought the scope of this task was only to have the name collection renamed and the existing data updated but I can update this PR as per the link you shared. I see it works the same way as _Sequelize_ normally handles migrations.
>

You're right on the scope, it's the how to go about it that I was questioning @koechkevin i.e. having to manually update DEV and then PROD DBs. Isn't it better to automate it?

koechkevin:
@kilemensi 

I have updated this PR.
- Added `migrate-mongo` dependancy.
- Added script `db-migrate` for up migrations and `db-rollback` to undo migrations.

kilemensi (commented):
Do we need to loop and update one doc at a time @koechkevin? We can update many or something?

koechkevin:
> Do we need to loop and update one doc at a time @koechkevin? We can update many or something?

@kilemensi 
tried something like this 

```js
updateMany(
  {},
  { $rename: { "blocks.$[].items.$[].collection": "blocks.$[].items.$[].item" } }
)
```

and got this error
`The source field for $rename may not be dynamic: blocks.$[].items.$[].collection`

saminegash (commented):
I think for the renaming  rather renaming the collection to items I believe it is better to fist create Items collection using proper migration then copy everthing from the collection table to items table then update the usage of collection to items after that deleting the collection database is a good way to handle this migration

saminegash:
> I think for the renaming rather renaming the collection to items I believe it is better to fist create Items collection using proper migration then copy everthing from the collection table to items table then update the usage of collection to items after that deleting the collection database is a good way to handle this migration

If we do this I think we are going to have zero failed jobs on production

kilemensi:
> I think for the renaming rather ...
>
@koechkevin correct me if I'm wrong but we're not trying to rename a collection @saminegash. There is a **field** in a block or collection that is named `collection`. This PR is about renaming that field.

apps/charterafrica/migrations/20230717085545-update_blocks_rename_collection_to_item.js
```diff
@@ -0,0 +1,71 @@
...
+
+  async down(db) {
+    const bulkOps = [];
+
+    await db
+      .collection("pages")
+      .find()
+      .forEach((doc) => {
+        if (doc.blocks && doc.blocks.length > 0) {
+          doc.blocks.forEach((block) => {
```
saminegash, line 45: here you have to use for of rather than foreach as foreach doesn't support blocking

saminegash (changes requested):

apps/charterafrica/migrations/20230717085545-update_blocks_rename_collection_to_item.js
koechkevin: @saminegash not sure why you would propose a for...of over forEach when we do not need to break/continue out of loop.

koechkevin:
@kilemensi @saminegash I found a way to rename nested arrays using [Aggregation pipeline](https://www.mongodb.com/docs/manual/tutorial/update-documents-with-aggregation-pipeline). Tested and works fine

apps/charterafrica/turbo.json
```diff
@@ -70,6 +70,8 @@
         "SENDGRID_FROM_EMAIL",
         "SENDGRID_FROM_NAME"
       ]
-    }
+    },
+    "db-rollback": {},
+    "db-migrate": {}
```
kilemensi, line 75: Do we need these in turbo right now?

kilemensi (approved):
🚀

saminegash (approved):

PR #534, opened 2023-07-28
Update ecosystem crons

koechkevin:
## Description

Updates cron jobs to ensure they run as expected

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/apps/charterafrica/package.json
+++ b/apps/charterafrica/package.json
@@ -29,6 +29,8 @@
     "start": "PAYLOAD_CONFIG_PATH=dist/payload.config.js NODE_ENV=production node dist/server.js",
     "lint-check": "TIMING=1 eslint './'",
     "lint": "TIMING=1 eslint --fix './'",
+    "update-ecosystem-list": "node scripts/ecosystem.js POST",
+    "update-ecosystem-content": "node scripts/ecosystem.js PUT",
     "jest": "jest",
     "playwright": "npx playwright test",
     "clean": "rm -rf .next .turbo build dist node_modules",
@@ -36,6 +38,7 @@
     "db-rollback": "migrate-mongo down"
   },
   "dependencies": {
+    "@actions/exec": "^1.1.1",
     "@aws-sdk/abort-controller": "^3.310.0",
     "@aws-sdk/client-s3": "^3.327.0",
     "@aws-sdk/lib-storage": "^3.327.0",
--- a/apps/charterafrica/scripts/ecosystem.js
+++ b/apps/charterafrica/scripts/ecosystem.js
@@ -1,0 +1,12 @@
+const { exec } = require("@actions/exec");
+
+require("dotenv").config({ path: "./.env" });
+require("dotenv").config({ path: "./.env.local" });
+
+(async () => {
+  const { RESOURCES_SECRET_TOKEN } = process.env;
+  const { NEXT_PUBLIC_APP_URL } = process.env;
+  const METHOD = process.argv[2];
+  const command = `curl -X ${METHOD} -H "x-api-key: ${RESOURCES_SECRET_TOKEN}" ${NEXT_PUBLIC_APP_URL}/api/v1/resources/ecosystem/entities`;
+  await exec(command);
+})();
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -39,6 +39,9 @@ importers:
 
   apps/charterafrica:
     dependencies:
+      "@actions/exec":
+        specifier: ^1.1.1
+        version: 1.1.1
       "@aws-sdk/abort-controller":
         specifier: ^3.310.0
         version: 3.310.0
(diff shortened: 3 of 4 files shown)


apps/charterafrica/scripts/ecosystem.js
```diff
@@ -0,0 +1,12 @@
...
+require("dotenv").config({ path: "./.env" });
+require("dotenv").config({ path: "./.env.local" });
+
+(async () => {
+  const { RESOURCES_SECRET_TOKEN } = process.env;
+  const { NEXT_PUBLIC_APP_URL } = process.env;
+  const METHOD = process.argv[2];
+  const command = `curl -X ${METHOD} -H "x-api-key: ${RESOURCES_SECRET_TOKEN}" ${NEXT_PUBLIC_APP_URL}/api/v1/resources/ecosystem/entities`;
+  await exec(command);
+})();
```
kilemensi, line 12: Why `@actions/exec`? Are we running this on GitHub Actions?

kilemensi (commented):
👍🏽 

----

Have we tested this? Do we know if it solves the current problem?

koechkevin:
> 👍🏽
> 
> Have we tested this? Do we know if it solves the current problem?
Tested this on local dokku server and it works @kilemensi.

Hourly crons run at the top of every hour whereas daily crons run at midnight
image

kilemensi:
> > 👍🏽
> > Have we tested this? Do we know if it solves the current problem?
> > Tested this on local dokku server and it works @kilemensi.
> 
> Hourly crons run at the top of every hour whereas daily crons run at midnight image

Cool @koechkevin but I meant does Dokku clean up afterwards? My understanding is: currently, Dokku doesn't clean up old jobs and hence eventually, the instance run out of resources.

apps/charterafrica/scripts/ecosystem.js
```diff
@@ -0,0 +1,18 @@
...
+  const { RESOURCES_SECRET_TOKEN } = process.env;
+  const { NEXT_PUBLIC_APP_URL } = process.env;
+  const METHOD = process.argv[2];
+  const command = `curl -X ${METHOD} -H "x-api-key: ${RESOURCES_SECRET_TOKEN}" ${NEXT_PUBLIC_APP_URL}/api/v1/resources/ecosystem/entities`;
+  exec(command, (err, stdout) => {
+    if (err) {
+      console.error(err);
+      return;
+    }
+    console.log(stdout);
```
kilemensi, line 16: Do we even need curl? [Node 18](https://nodejs.org/es/blog/announcements/v18-release-announce) supports `fetch` natively.

apps/charterafrica/scripts/ecosystem.js
```diff
@@ -0,0 +1,18 @@
...
+  });
+})();
```
kilemensi, line 18: Why do we need IIFE here?

koechkevin: @kilemensi this file is going to be executed via package json script so I think having it as an invoked function makes sense

kilemensi: Could be but I'm looking at a bunch of scripts in say MUI [repo](https://github.com/mui/material-ui/blob/master/scripts/) & they don't use IIFE.

apps/charterafrica/package.json
```diff
@@ -29,13 +29,16 @@
     "start": "PAYLOAD_CONFIG_PATH=${PAYLOAD_CONFIG_PATH:-dist/payload.config.js} NODE_ENV=${NODE_ENV:-production} node dist/server.js",
     "lint-check": "TIMING=1 eslint './'",
     "lint": "TIMING=1 eslint --fix './'",
+    "update-ecosystem-list": "node scripts/ecosystem.js POST",
+    "update-ecosystem-content": "node scripts/ecosystem.js PUT",
```
kilemensi, line 33: We no longer need these since we're directly running scripts/<script name>

apps/charterafrica/package.json
```diff
@@ -29,13 +29,16 @@
...
     "jest": "jest",
     "playwright": "npx playwright test",
     "clean": "rm -rf .next .turbo build dist node_modules",
     "db-migrate": "migrate-mongo up",
     "db-rollback": "migrate-mongo down"
   },
   "dependencies": {
+    "@actions/exec": "^1.1.1",
```
kilemensi, line 41: We don't use this anymore

kilemensi (commented):
👍🏽

apps/charterafrica/scripts/ecosystem.js
```diff
@@ -0,0 +1,21 @@
...
+    `${NEXT_PUBLIC_APP_URL}/api/v1/resources/ecosystem/entities`,
+    {
+      method,
+      headers,
+    },
+  );
+  return res.json();
+}
+
+main().then(console.log);
```
kilemensi, line 21: Since you're using `async`, why not `await`?

```suggestion
const responseJson = await main()
console.log(responseJson)
```

pnpm-lock.yaml
```diff
@@ -39,6 +39,9 @@ importers:
 
   apps/charterafrica:
     dependencies:
+      "@actions/exec":
+        specifier: ^1.1.1
+        version: 1.1.1
```
kilemensi, line 44: This file shouldn't contain any changes.

apps/charterafrica/scripts/ecosystem.js
koechkevin: @kilemensi do we support top level await yet?

kilemensi: It [is](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await#browser_compatibility) in modules @koechkevin ... you may have to save the file as `ecosystem.mjs`.

kilemensi (approved):
🚀

PR #549, opened 2023-08-10
Fix Charterafrica issues

koechkevin:
## Description
This PR fixes the following issues

1. Links to organisation and contributor pages from single tool page not working.
2.  AGA infographic not loaded on about page despite it being added as a block.

[Slack](https://code4africa.slack.com/archives/C04AJ0124H1/p1691503091967479?thread_ts=1691502758.677989&cid=C04AJ0124H1)

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/apps/charterafrica/src/components/Infographic/Infographic.js
+++ b/apps/charterafrica/src/components/Infographic/Infographic.js
@@ -2,13 +2,15 @@ import { Box } from "@mui/material";
 import Script from "next/script";
 import React from "react";
 
+import site from "@/charterafrica/utils/site";
+
 const Infographic = React.forwardRef(function Infographic(props, ref) {
   const {
     countriesFile: countriesFileProp,
     dataFile: dataFileProp,
     infographicId: id,
     sx,
-    url,
+    url = site.url,
   } = props;
 
   if (!(id && url)) {
--- a/apps/charterafrica/src/components/Infographic/Infographic.js
+++ b/apps/charterafrica/src/components/Infographic/Infographic.js
@@ -10,9 +10,9 @@ const Infographic = React.forwardRef(function Infographic(props, ref) {
     dataFile: dataFileProp,
     infographicId: id,
     sx,
-    url = site.url,
   } = props;
 
+  const { url } = site;
   if (!(id && url)) {
     return null;
   }
--- a/apps/charterafrica/src/components/Tool/Contributors.js
+++ b/apps/charterafrica/src/components/Tool/Contributors.js
@@ -1,4 +1,5 @@
 import { RichTypography } from "@commons-ui/core";
+import { Link } from "@commons-ui/next";
 import { Avatar, AvatarGroup, Box } from "@mui/material";
 import PropTypes from "prop-types";
 import React from "react";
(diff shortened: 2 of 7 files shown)


kilemensi:
Can you describe the bug @koechkevin ?

apps/charterafrica/src/components/Infographic/Infographic.js
```diff
@@ -2,13 +2,15 @@ import { Box } from "@mui/material";
...
+import site from "@/charterafrica/utils/site";
+
 const Infographic = React.forwardRef(function Infographic(props, ref) {
   const {
     countriesFile: countriesFileProp,
     dataFile: dataFileProp,
     infographicId: id,
     sx,
-    url,
+    url = site.url,
```
kilemensi, line 13: If the component is aware of `site`, then there is no point of passing `url` as prop. It can access `site` object directly.
```suggestion
```

i.e. this is why the original plan was to process `url` in the backend.

apps/charterafrica/src/components/Tool/Contributors.js
```diff
@@ -16,7 +17,13 @@ const Contributors = React.forwardRef(function Contributors(props, ref) {
         </RichTypography>
         <AvatarGroup sx={{ justifyContent: { xs: "center", sm: "right" } }}>
           {list.map((item) => (
-            <Avatar key={item.id} alt={item.name} src={item.avatarUrl} />
+            <Avatar
+              key={item.id}
+              component={Link}
+              href={item.link?.href}
```
kilemensi, line 23: If we're using Link we need to make sure there is **always** `link.href`. Otherwise, Next.js will throw errors in the console.

apps/charterafrica/src/components/Tool/OrgThemeAndOperatingCountries.js
```diff
@@ -23,14 +23,16 @@ const OrgThemeAndOperatingCountries = React.forwardRef(
             width: "auto",
           }}
         />
-        <RichTypography
+        <Link
+          href={organisation?.link?.href || "#"}
+          component={RichTypography}
```
kilemensi, line 28: Not sure I follow here: `Link` using `RichTypography` as component? `Link` is already built on [top](https://mui.com/material-ui/react-link/#basic-links) of `Typography`, do we need both?

apps/charterafrica/src/lib/data/common/processPageTools.js
```diff
@@ -53,6 +53,13 @@ async function processPageSingleTool(page, api, context) {
   });
   const tools = [];
   const filterLabels = labelsPerLocale[locale];
+  const organisationPage = await getPageUrl(api, "organisations");
+  const organisation = orgDocs?.[0]
+    ? {
+        ...orgDocs?.[0],
+        link: { href: `${organisationPage}/${orgDocs?.[0].slug}` },
+      }
+    : null;
```
kilemensi, line 62: Cool if we're pressed for time but I think this should be handled in the CMS code. Similar to how we have `formatPagePath`, we should have something similar to resources we pull from the CMS.

kilemensi (commented):
👍🏽

apps/charterafrica/src/components/Tool/OrgThemeAndOperatingCountries.js
koechkevin: Then we do not need RichTypography

apps/charterafrica/src/components/Tool/Contributors.js
koechkevin: Can never be `null` or `undefined` since we are appending this at `processPageTools`

kilemensi: Then don't use `link?.href` @koechkevin, it implies `link` can be `undefined`/`null`. Just `link.href`.

kilemensi (approved):
🚀

PR #555, opened 2023-08-17
Refactor page links to use field hooks

### LINKED ISSUE #550: Refactor page links to use field hooks
Cool if we're pressed for time but I think this should be handled in the CMS code. Similar to how we have `formatPagePath`, we should have something similar for resources we pull from the CMS.

_Originally posted by @kilemensi in https://github.com/CodeForAfrica/ui/pull/549#discussion_r1290040032_

saminegash:
## Description

We current have to manually compute page links for resource collections (Contributors, Organisations and Tools) on Next.js server side because these collections are not linked to the pages that they need to appear in. This PR addresses that using `afterRead` hooks and ensures data received from Payload CMS already has these links computed.

Fixes #550 

## Type of change

- [x] Chore

## Screenshots

N/A

## Checklist:

resolve #550


kilemensi (commented):
👍🏽 @saminegash 

---

From quickly reading this, it looks like we're setting the organisation collection item href at create/update time. Since this is read-only property, I think it makes more sense to set it at read time.

For example, if content team decides organisations _page_ should be renamed to teams, all organisation collection items stored in the DB will now be pointing to an invalid page (we're not tracking when organisation page is changed).

The better approach would be: when organisation collection item is read, get the current organisation page and compute the collection item href from it. This way, the collection item href will never go stale.


PS: What about other resources, tools, contributors, etc.? Don't we need to compute href for them as well?

apps/charterafrica/src/payload/utils/mapLinkAndNameforContributorsInToolsCollection.js
```diff
@@ -0,0 +1,25 @@
+import getPageUrlUsingPayload from "../../lib/data/common/getPageUrlUsingPayload";
+
+const mapLinkAndNameforContributorsInToolsCollection = async ({
+  data,
```
koechkevin, line 4: @saminegash afterReadCollectionHook passes doc as a parameter and not data

apps/charterafrica/src/payload/utils/mapLinkAndNameforContributorsInToolsCollection.js
```diff
@@ -0,0 +1,25 @@
...
+      payload,
+      "contributors",
+    );
+    const contributors = tool?.contributors?.map((person) => ({
+      ...person,
+      link: { href: `${contributorPage}/${person.slug}` },
+      name: person.name || person?.fullName || person.username || null,
+    }));
+
+    return contributors;
```
koechkevin, line 19: Here you will return doc and should work as expected

apps/charterafrica/src/payload/utils/mapLinkToHrefForOrgCollection.js
```diff
@@ -0,0 +1,18 @@
...
+const mapLinkToHrefForOrgCollection = async ({ data, req: { payload } }) => {
+  try {
+    const doc = { ...data.doc };
+    const organisationPage = await getPageUrlUsingPayload(
+      payload,
+      "organisations",
+    );
+    const href = `${organisationPage}/${doc?.slug}`;
+
+    return href;
```
koechkevin, line 12: Also here, modify `doc` and return `doc`

apps/charterafrica/src/lib/data/common/getPageUrlUsingPayload.js
```diff
@@ -0,0 +1,21 @@
+const getPageUrlUsingPayload = async (payload, slug) => {
```
koechkevin, line 1: move this file to `payload/utils`

koechkevin (approved):

apps/charterafrica/src/payload/utils/mapLinkAndNameforContributorsInToolsCollection.js
```diff
@@ -0,0 +1,29 @@
+import getPageUrlUsingPayload from "./getPageUrlUsingPayload";
+
+const mapLinkAndNameforContributorsInToolsCollection = async ({
```
kilemensi, line 3: Why do we need to do this? Doesn't having the hook execute on contributors will automatically appear anywhere contributors are pulled? Do we need really to write something specific for contributors in tools?

apps/charterafrica/src/payload/utils/mapLinkForToolsCollection.js
```diff
@@ -0,0 +1,17 @@
+import getPageUrlUsingPayload from "./getPageUrlUsingPayload";
+
+const mapLinkForToolsCollection = async ({ doc, req: { payload } }) => {
```
kilemensi, line 3: We can have a single higher-order function that receives page slug and return a function that adds href.

I mean all these hooks look exactly the same with exception of page slug (`tools`, `contributors` and `organisations`), right

apps/charterafrica/src/payload/utils/getPageUrlUsingPayload.js
```diff
@@ -0,0 +1,21 @@
...
+        equals: slug,
+      },
+    },
+  };
+
+  const { docs } = await payload.find({
+    limit: 0,
+    ...options,
+    collection,
+  });
```
kilemensi, line 15: Why not simplify all of this?

```suggestion
  const options = {
    collection: "pages",
    where: {
      slug: {
        equals: slug,
      },
    },
    limit: 0,
  };

  const { docs } = await payload.find(options);
```
```

File: apps/charterafrica/src/payload/utils/getPageUrlUsingPayload.js [hunk 8]
```diff
@@ -0,0 +1,21 @@
...
+
+  const breadcrumbs = docs[0]?.breadcrumbs || [];
+  return breadcrumbs[breadcrumbs.length - 1]?.url;
```
kilemensi, line 18: We already have a [function](https://github.com/CodeForAfrica/ui/blob/55ee12c0a53855c3055fd788a80bae204722a26d/apps/charterafrica/src/payload/utils/formatPagePath.js#L1) to return page path/url given a page.

apps/charterafrica/src/payload/utils/mapLinkForToolsCollection.js
kilemensi: Yes, and my question [here](https://github.com/CodeForAfrica/ui/pull/555#discussion_r1301832395) is why is this different?

kilemensi: Yes, I know that code, and that's what I'm asking about. This is how we're loading a tool:

```js
  const { docs } = await api.getCollection(collection, {
    locale,
    where: {
      slug: {
        equals: slug,
      },
    },
  });
```
If found, `tool = docs[0]`, right? My questions is: since `tool` contain `contributors` i.e. `tools.contributors`, doesn't this call also trigger the `afterRead` hook on contributors?

kilemensi: If it triggers the hook then we can just add a hook in Contributors collection the same way we’ve added on in Organisations, no?

kilemensi: No you haven't @saminegash . Please review the current implementation.

apps/charterafrica/src/payload/collections/Contributors.js
```diff
@@ -3,6 +3,15 @@ import dateField from "../fields/dateField";
 import slug from "../fields/slug";
 import source from "../fields/source";
 import { CONTRIBUTORS_COLLECTION } from "../utils/collections";
+import nestCollectionUnderPage from "../utils/nestCollectionUnderPage";
+
+function useFullNameOrExternalId({ doc }) {
+  if (doc) {
+    const name = doc.name ?? doc?.fullName ?? doc.externalId ?? null;
```
kilemensi, line 10:
```suggestion
    const name = doc.name ?? doc.fullName ?? doc.externalId ?? null;
```

kilemensi:
@saminegash / @koechkevin

kilemensi:
Your reviews @koechkevin @saminegash

apps/charterafrica/src/lib/data/common/processPageOrganisations.js
```diff
@@ -24,19 +23,11 @@ async function processPageSingleOrganisation(page, api, context) {
   if (!docs?.length) {
     return null;
   }
-  const organisation = docs[0] || {};
 
-  const pageUrl = await getPageUrl(api, "tools");
+  const organisation = docs[0] || {};
   const tools = organisation.tools.map((tool) => {
```
koechkevin, line 28: @saminegash I think we need safe navigation here

kilemensi:
Your review @kelvinkipruto

kelvinkipruto (approved):

kelvinkipruto:
@saminegash Update the readme description to match the changes.

kilemensi:
Your review @koechkevin. We need to make sure we don't miss any edge case before merging.

koechkevin (approved):

kilemensi (approved):

PR #583, opened 2023-09-14
Add Page SEO

### LINKED ISSUE #562: @/codeforafrica seo
Implement SEO

The implementation should also keep in mind https://github.com/CodeForAfrica/ui/pull/443 since this approach promises to minimise changes should content editors move blocks around.

Current content:
https://github.com/CodeForAfrica/ui/blob/daacc18ac4e896494b8c7ae9fe84a978b2c2afd8/apps/codeforafrica/src/lib/api.netlify-cms/seo.js

kelvinkipruto:
## Description
This PR adds individual Page SEO. This PR does not handle SEO for nested pages. This will be handled by #584 

Fixes #562

## Type of change

## Checklist:


apps/codeforafrica/src/lib/data/rest/index.js
```diff
@@ -37,6 +37,6 @@ export const api = {
 export async function getPageStaticProps(context) {
   const props = await getPageProps(api, context);
   return {
-    props,
+    props: props || {},
```
kilemensi, line 40: Why are we forcing it here?

apps/codeforafrica/src/lib/data/seo.js
```diff
@@ -0,0 +1,61 @@
+import site from "@/codeforafrica/utils/site";
+
+function stringifyDescription(description) {
+  if (!description || !Array.isArray(description)) {
```
kilemensi, line 4: Isn't 👇🏽 enough? i.e. Array.isArray handles null/undefined and description will always be an array if present, no?

```suggestion
  if (!Array.isArray(description)) {
```

apps/codeforafrica/src/lib/data/seo.js
```diff
@@ -0,0 +1,61 @@
...
+    return "";
+  }
+  let result = "";
+  description.forEach((item) => {
```
kilemensi, line 8: Not a fan of forEach... I think map or reduce are always better.

apps/codeforafrica/src/lib/data/seo.js
```diff
@@ -0,0 +1,61 @@
...
+
+export default function getPageSeoFromMeta(page, settings) {
+  const { title: pageTitle, meta: pageMeta } = page;
+  const {
+    title: metaTitle,
+    description: metaDescription,
+    image = {},
+  } = pageMeta;
+  const { title: siteTitle, description: siteDescription } = settings;
+  const title = metaTitle || pageTitle || siteTitle || null;
```
kilemensi, line 28: Don't we have site title also defined in env vars? APP_NAME or APP_TITLE?

koechkevin (approved):

kilemensi (commented):
👍🏽 

---

There is a way we're handling alternative hrefs or urls in SEO and I don't see it here. Is it because we don't support localisation?

apps/codeforafrica/src/lib/data/rest/index.js
kelvinkipruto: @kilemensi This is to handle an edge case where the error page has not been created in the CMS yet.

kelvinkipruto:
> There is a way we're handling alternative hrefs or urls in SEO and I don't see it here. Is it because we don't support localisation?

@kilemensi, Yes. We do not support localisation yet.

kelvinkipruto:
@kilemensi @koechkevin This is ready

kilemensi:
Can we update this PR @kelvinkipruto ?

kelvinkipruto:
@kilemensi updated

apps/codeforafrica/src/lib/data/seo.js
```diff
@@ -0,0 +1,67 @@
...
+    null;
+  const description =
+    metaDescription || stringifyDescription(siteDescription) || null;
+  const titleTemplate = siteTitle ? `%s | ${siteTitle}` : null;
+  const defaultTitle = siteTitle || null;
+  const canonical = site.url.replace(/\/+$/, "");
+  const openGraph = {
+    title,
+    description,
+    type: "website",
```
kilemensi, line 43: Add a story to track these: we may have to use different types for stories, teams, etc.

kilemensi (approved):

apps/codeforafrica/src/lib/data/seo.js
kelvinkipruto: @kilemensi Already there at #584

PR #596, opened 2023-09-19
Fix failing build if no error page

kelvinkipruto:
## Description
This PR fixes a bug which causes the build to fail if an error page has not been created in the CMS. It introduces default values for the error page.

## Type of change

## Checklist:

### DIFF
--- a/apps/codeforafrica/src/lib/data/common/index.js
+++ b/apps/codeforafrica/src/lib/data/common/index.js
@@ -44,37 +44,41 @@ function getPageSlug({ params }) {
   return params?.slugs?.[pageSlugIndex] || "index";
 }
 
+function getDefaultErrorPageProps() {
+  return {
+    title: "Page not found. ",
+    subtitle: [
+      {
+        children: [
+          {
+            text: "Visit our ",
+            children: null,
+          },
+          {
+            type: "link",
+            newTab: false,
+            url: "/",
+            children: [
+              {
+                text: "homepage",
+                children: null,
+              },
+            ],
+            href: "/",
+          },
+        ],
+      },
+    ],
+  };
+}
+
 export async function getPageProps(api, context) {
   const slug = getPageSlug(context);
   const {
     docs: [page],
   } = await api.findPage(slug);
   if (!page && (slug === "404" || slug === "500")) {
-    return {
-      title: "Page not found",
-      subtitle: [
-        {
-          children: [
-            {
-              text: "Visit our ",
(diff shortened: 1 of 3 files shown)


apps/codeforafrica/src/lib/data/common/index.js
```diff
@@ -49,6 +49,11 @@ export async function getPageProps(api, context) {
   const {
     docs: [page],
   } = await api.findPage(slug);
+  if (!page && (slug === "404" || slug === "500")) {
+    return {
+      title: "Page not found",
+    };
+  }
```
kilemensi, line 56: 1. Don't different error pages have different titles?
2. Don't they also have subtitles/descriptions?

---

PS: Doesn't 500 page have to be static?

kelvinkipruto: @kilemensi The error pages all have the same title and description.

kilemensi: Can't be @kelvinkipruto: https://github.com/CodeForAfrica/ui/blob/main/apps/codeforafrica/content/pages/error.md

apps/codeforafrica/src/lib/data/common/index.js
```diff
@@ -44,11 +44,113 @@ function getPageSlug({ params }) {
   return params?.slugs?.[pageSlugIndex] || "index";
 }
 
+function getDefaultErrorPageProps(statusCode = 404) {
+  if (statusCode === 500) {
+    return {
```
kilemensi, line 49:
```suggestion
function getDefaultErrorPageProps(slug = "404) {
  if (slug === "500") {
    return {
```

Why `slug` to `statusCode` if `slug` can do?

apps/codeforafrica/src/lib/data/common/index.js
```diff
@@ -44,11 +44,113 @@ function getPageSlug({ params }) {
...
   } = await api.findPage(slug);
+  if (!page) {
+    if (slug === "500") {
+      return getDefaultErrorPageProps(500);
+    }
+    return getDefaultErrorPageProps();
+  }
   if (!page) {
     return null;
   }
```
kilemensi, line 156:
```suggestion
    if (["404", "500"].includes(slug)) {
      return getDefaultErrorPageProps(slug);
    }
    return null;
  }
```

A single `if (!page)` should be enough.

kilemensi (approved):

PR #610, opened 2023-09-25
Project List block

koechkevin:
## Description

Fixes # (issue)

## Type of change

## Checklist:


kilemensi:
This was our last discussion on [Slack](https://code4africa.slack.com/archives/G0DNN0CN5/p1695630785677869?thread_ts=1695628448.235339&cid=G0DNN0CN5) @koechkevin:
>We can keep the current implementation i.e. separate Donors/Partners collections.
We can combine the two collections but change the collection name to something like Stakeholders. Both donors and partners fields will now link to stakeholders.

Which approach has been taken in this PR?

kelvinkipruto (approved):
LGTM.
@koechkevin Add SEO

koechkevin:
> This was our last discussion on [Slack](https://code4africa.slack.com/archives/G0DNN0CN5/p1695630785677869?thread_ts=1695628448.235339&cid=G0DNN0CN5) @koechkevin:
> 
> > We can keep the current implementation i.e. separate Donors/Partners collections.
> > We can combine the two collections but change the collection name to something like Stakeholders. Both donors and partners fields will now link to stakeholders.
> 
> Which approach has been taken in this PR?
I introduced a separate Donors block.

The first approach @kilemensi

apps/codeforafrica/src/components/Projects/Projects.js
```diff
@@ -57,7 +55,9 @@ const Projects = React.forwardRef(function Projects(
   }, [data]);
 
   useEffect(() => {
-    router.push(queryParams, undefined, {
+    const [pathname] = router.asPath.split("?");
```
kilemensi, line 58: Why are we using `asPath` here instead of `pathname`? If we need to use asPath, please see the warning [here](https://nextjs.org/docs/pages/api-reference/functions/use-router#router-object).

apps/codeforafrica/src/components/Projects/Projects.js
```diff
@@ -84,7 +84,7 @@ const Projects = React.forwardRef(function Projects(
         {projects?.length > 0 ? (
           <Stack direction="column" spacing={{ xs: 5, md: 7.5 }}>
             {projects.map((project) => (
-              <ProjectCard {...project} key={project.slug} />
+              <ProjectCard {...project} key={project.id} />
```
kilemensi, line 87: Again, I'd rather we use `slug`, `href`, etc., i.e. actual content.

apps/codeforafrica/src/payload/fields/dateTime.js
```diff
@@ -0,0 +1,17 @@
+import { deepmerge } from "@mui/utils";
```
kilemensi, line 1: Can't we reuse `publishedOn` (or at least create one reusable `dateTime` field)?

apps/codeforafrica/src/payload/collections/Projects.js
```diff
@@ -0,0 +1,122 @@
...
+      label: {
+        en: "Team",
+      },
+      required: true,
+      type: "relationship",
+      relationTo: "members",
+      hasMany: true,
+    },
+    {
+      name: "donors",
```
kilemensi, line 106: Why is team between partners and donors?

apps/codeforafrica/src/payload/collections/Projects.js
```diff
@@ -0,0 +1,122 @@
...
+    image({
+      overrides: {
+        label: {
+          en: "Thumbnail",
+        },
+        name: "thumbnail",
+        required: true,
+      },
+    }),
+    linkGroup({ overrides: { name: "externalHref" } }),
```
kilemensi, line 65:
```suggestion
    linkGroup({ overrides: { name: "externalLink" } }),
```

I think we call them links, no?

apps/codeforafrica/src/payload/collections/Projects.js
```diff
@@ -0,0 +1,122 @@
...
+    },
+    {
+      name: "tagLine",
+      label: {
+        en: "Tag Line",
+      },
+      type: "text",
+    },
+    {
+      name: "title",
```
kilemensi, line 32: What's the difference between `name` and `title`?

kilemensi (commented):
👍🏽 

---

Pay attention to how information is organised/flows

apps/codeforafrica/src/components/Projects/Projects.js
koechkevin: @kilemensi pathname exposes is `/[...slugs]` and comes with this error.
image

kilemensi: 👍🏽  @koechkevin... then if you _have_ to use `asPath`, use is as directed in the docs.

apps/codeforafrica/src/payload/collections/Projects.js
koechkevin: @Kilemensi, from the implementation [here](https://github.com/CodeForAfrica/ui/blob/develop/apps/codeforafrica/content/projects/academyafrica.md) both title and name exist and they are both different.

kilemensi: Yes @koechkevin hence me asking what's their difference?

koechkevin: This is how it is the difference from the site @kilemensi 
image

kilemensi: Very weird seeing name and title used this way @koechkevin ... feels like title should be name, subtitle should be title, etc.

... but, lets continue while we think about it some more.

apps/codeforafrica/src/components/ProjectCard/ProjectCard.js
```diff
@@ -30,8 +31,8 @@ const ProjectActionArea = styled(CardActionArea, {
 
 const ProjectCard = React.forwardRef(function ProjectCard(props, ref) {
   const {
-    externalHref,
-    href,
+    externalLink,
```
kilemensi, line 34: Where is `externalLink` being used in this file?

apps/codeforafrica/src/components/ProjectPageHeader/ProjectPageHeader.js
```diff
@@ -59,7 +59,7 @@ const ProjectTitle = styled(RichTypography, {
 const ProjectPageHeader = React.forwardRef(
   function ProjectPageHeader(props, ref) {
     const {
-      externalHref,
+      externalLink,
```
kilemensi, line 62: `externalLink` would be an object similar to `link` i.e. with href attribute.

apps/codeforafrica/src/components/Projects/Projects.js
```diff
@@ -57,10 +55,14 @@ const Projects = React.forwardRef(function Projects(
...
 
   useEffect(() => {
-    router.push(queryParams, undefined, {
-      scroll: true,
-      shallow: true,
-    });
+    if (router.isReady) {
+      const [pathname] = router.asPath.split("?");
+      const url = pathname ? `${pathname}${queryParams}` : queryParams;
+      router.push(url, undefined, {
```
kilemensi, line 61: Just for my understanding, if `url` we're passing to `router.push` doesn't start with `/`, isn't the `router` smart enough to keep us on the same page and just update the query part? For example, if my current page is `/projects/pesayetu` and I run:
```js
    router.push('?page=2', undefined, { scroll: true, shallow: true });
```
Isn't my new address going to be `/projects/pesayetu?page=2`?

Basically, do I always need to include `/projects/pesayetu` even if I'm just changing the query part?

apps/codeforafrica/src/payload/collections/Projects.js
```diff
@@ -0,0 +1,130 @@
...
+      type: "text",
+      required: true,
+    },
+    richText({
+      name: "subtitle",
+    }),
+    {
+      name: "tagLine",
+      label: {
+        en: "Tag Line",
```
kilemensi, line 44:
```suggestion
        en: "Tagline",
```

apps/codeforafrica/src/payload/collections/Projects.js
```diff
@@ -0,0 +1,130 @@
...
+      type: "text",
+    },
+    tags({
+      name: "tag",
+      hasMany: false,
+      admin: {
+        position: "sidebar",
+      },
+    }),
+    linkGroup({ overrides: { name: "externalLink", label: "External Link" } }),
```
kilemensi, line 55:
```suggestion
    linkGroup({ overrides: { name: "externalLink", label: "Link" } }),
```
Shouldn't the label just be Link? To the code, it's external because there is internal link to the page but the content editors, there is just one link, no?

apps/codeforafrica/src/components/Projects/Projects.js
koechkevin: @kilemensi no, the new address will look something like `/[...slugs]?page=2`

kilemensi: Interesting @koechkevin ... then if `queryParams` where an object, using the object form of [`push`](https://nextjs.org/docs/pages/api-reference/functions/use-router#with-url-object) would have worked best here but that ship has sailed.

Rather than doing the whole split thing, `window.location.pathname` should provide us with what we need, no?

koechkevin: That also works but we will have to disable eslint ` // eslint-disable-next-line no-undef` my preference would still be asPath and check if `router.isReady`

kilemensi: We don't need to disable eslint... you just need to tell eslint this file is running on a [browser](https://github.com/CodeForAfrica/ui/blob/40ec020db553f86d37fc2c992f0f05407574c751/apps/codeforafrica/src/pages/admin/index.page.js#L1).

>my preference would still be asPath and check if router.isReady
Why do that if browser already gives you the value?

koechkevin:
> Just run this locally @koechkevin, we seem to be using tags but tags are also used by posts. We end up with things like **Projects**, **Initiatives**, **Stories**, **Opportunities**, etc., instead of something like 👇🏽
> 
> Screenshot 2023-09-28 at 15 16 22
> I think works should use different tags or should filter tags for existing works only.

@kilemensi I have updated this to get from existing works only

kilemensi:
Anything that causes image/title to flip @koechkevin (PR vs dev)
s

kilemensi:
Also @koechkevin :

Title should be rich text
Screenshot 2023-09-29 at 09 13 32


Description text has wrong font size and links should have icons
s

koechkevin:
> Anything that causes image/title to flip @koechkevin (PR vs dev) s

@kilemensi what should be the intended outcome?
In the [main branch code](https://github.com/CodeForAfrica/ui/blob/40ec020db553f86d37fc2c992f0f05407574c751/apps/codeforafrica/src/components/ProjectPageHeader/ProjectPageHeader.js#L89C29-L89C29) image comes before content. Perhaps an issue on the dev site?

kilemensi:
> @kilemensi what should be the intended outcome? In the [main branch code](https://github.com/CodeForAfrica/ui/blob/40ec020db553f86d37fc2c992f0f05407574c751/apps/codeforafrica/src/components/ProjectPageHeader/ProjectPageHeader.js#L89C29-L89C29) image comes before content. Perhaps an issue on the dev site?
>

Could be @koechkevin, source of truth should be [Figma](https://www.figma.com/file/MbCygAaNhIjgJOHrQZpZ2U/CfA---CfA-site?type=design&node-id=5283-939&mode=design&t=LluIldmCpj2j01Ai-0). What does it say?

koechkevin:
@kilemensi

apps/codeforafrica/payload.config.ts
```diff
@@ -95,7 +99,7 @@ export default buildConfig({
         doc?.slug?.value ? `${appURL}/${doc.slug.value}` : undefined,
     } as any),
     nestedDocs({
-      collections: ["pages"],
+      collections: ["pages", "projects"],
```
kilemensi, line 102: We shouldn't need this.

apps/codeforafrica/src/components/Projects/Projects.js
```diff
@@ -57,7 +56,9 @@ const Projects = React.forwardRef(function Projects(
   }, [data]);
 
   useEffect(() => {
-    router.push(queryParams, undefined, {
+    const { pathname } = window.location;
+    const url = pathname ? `${pathname}${queryParams}` : queryParams;
```
kilemensi, line 60:
```suggestion
    const url = `${pathname}${queryParams}`;
```

`pathname` should always be available.

apps/codeforafrica/src/lib/data/blockify/ourWork.js
```diff
@@ -0,0 +1,18 @@
+import { getProjects } from "@/codeforafrica/lib/data/utils/projects";
+
+async function ourWork(block, api, context) {
+  const { query } = context;
+  const data = await getProjects(api, query);
+  const { docs: allProjects } = await api.getCollection("projects");
+  const projectTags = allProjects.map(({ tag }) => tag?.name).filter(Boolean);
```
kilemensi, line 7: Are we just taking `tag.name`? I thought we would be using `tag.slag` for filtering.

apps/codeforafrica/src/payload/collections/Projects.js
```diff
@@ -0,0 +1,150 @@
...
+          type: "text",
+          required: true,
+        },
+        {
+          name: "date",
+          type: "date",
+          required: true,
+          admin: {
+            date: {
+              pickerAppearance: "dayAndTime",
```
kilemensi, line 84:
```suggestion
              pickerAppearance: "dayOnly",
```

apps/codeforafrica/src/payload/collections/Projects.js
```diff
@@ -0,0 +1,150 @@
...
+              },
+              {
+                value: "data",
+                label: { en: "Data Icon" },
+              },
+            ],
+          },
+        ],
+      },
+    }),
```
kilemensi, line 115: This is hard to understand. Why set link icon?

I think the better approach is to have link type and we only allow one of two types: source code link (which we'll show using GitHub icon) and data link (which we'll show using database icon).

Otherwise, we might as well allow content editors to set actual icons.

apps/codeforafrica/src/payload/fields/links/link.js
```diff
@@ -31,6 +31,7 @@ const link = ({
   disableOpenInNewTab = false,
   overrides = {},
   required = true,
+  labelFields: fields = [],
```
kilemensi, line 34: Why do we need this?

koechkevin: @kilemensi we need extra field for icons in linkArray

kilemensi: We can override linkArray fields directly

kilemensi (approved):

Issue #606, opened 2023-09-22
Codeforafrica single stories page


kelvinkipruto:
Implement Codeforafrica single stories page

PR #617, opened 2023-09-27
Ft/Posts

### LINKED ISSUE #593: @codeforafrica opportunities
Implement Opportunities  and individual opportunity pages

The implementation should also keep in mind https://github.com/CodeForAfrica/ui/pull/443 since this approach promises to minimise changes should content editors move blocks around.

kelvinkipruto:
## Description
This PR implements posts pages

Fixes #593 #606

## Type of change

## Screenshots
image
image
image
image
image

## Checklist:


apps/codeforafrica/src/lib/data/utils/posts.js
```diff
@@ -0,0 +1,185 @@
...
+    limit: 0,
+    where: {
+      "tags.name": {
+        like: primaryTag,
+      },
+    },
+  });
+  const allTags = posts.reduce((acc, story) => {
+    const { tags = [] } = story;
+    return [...acc, ...tags.map(({ name, slug }) => ({ name, slug }))];
```
koechkevin, line 179: @kelvinkipruto why map to the same thing?. Also, why not just fetch all tags?

apps/codeforafrica/src/payload/blocks/Posts.js
```diff
@@ -0,0 +1,70 @@
...
+    value: "opportunities",
+  },
+];
+
+const Posts = {
+  slug: "post-list",
+  fields: [
+    {
+      name: "primaryTag",
+      label: "Primary Tag",
```
kilemensi, line 17: Label should be something that makes sense to content editors... Post type or something along those lines.

apps/codeforafrica/src/lib/data/utils/posts.js
```diff
@@ -0,0 +1,185 @@
...
+    ...meta,
+  };
+  return {
+    title,
+    blocks: [
+      {
+        authors: authors.map(({ fullName }) => {
+          return {
+            name: fullName,
+            bio: "",
```
kilemensi, line 148: Any reason why we didn't structure authors to be like this from the get go?

apps/codeforafrica/src/lib/data/utils/posts.js
```diff
@@ -0,0 +1,185 @@
...
+    id,
+    title,
+    image: { src, alt },
+    excerpt,
+    publishedOn: formatDate(publishedOn, {
+      includeTime: false,
+      month: "short",
+    }),
+    tags: formatTags(tags),
+    href: `/posts/${primaryTag}/${slug}`,
```
kilemensi, line 49: The nested afterRead hook can do this, no?

apps/codeforafrica/src/lib/data/utils/posts.js
```diff
@@ -0,0 +1,185 @@
+import formatDate from "@/codeforafrica/utils/formatDate";
+
+export function formatTags(tags) {
+  const excludedTags = new Set(["stories", "opportunities"]);
```
kilemensi, line 4: If you know the distinctive values already, why use `Set`?

apps/codeforafrica/src/lib/data/utils/posts.js
```diff
@@ -0,0 +1,185 @@
...
+
+  const tagCounts = tags.reduce((counts, { name }) => {
```
kilemensi, line 6: Remove exclude tags from this point so that we don't need to worry about them when sorting.

apps/codeforafrica/src/lib/data/utils/posts.js
```diff
@@ -0,0 +1,185 @@
...
+
+  const sortedTags = Object.keys(tagCounts)
+    .filter((tag) => !excludedTags.has(tag.toLowerCase()))
+    .sort((a, b) => tagCounts[b] - tagCounts[a])
+    .map((tagName) => {
+      const tag = tags.find(({ name }) => name === tagName);
+      return {
+        name: tag.name,
+        slug: tag.slug,
+      };
```
kilemensi, line 20: ? Why not just return the found `tag`?

apps/codeforafrica/src/lib/data/utils/posts.js
```diff
@@ -0,0 +1,185 @@
...
+  return {
+    id,
+    title,
+    image: { src, alt },
+    excerpt,
+    publishedOn: formatDate(publishedOn, {
+      includeTime: false,
+      month: "short",
+    }),
+    tags: formatTags(tags),
```
kilemensi, line 48: Do we format tags for each individual post?

apps/codeforafrica/src/lib/data/pagify/stories.js
```diff
@@ -0,0 +1,16 @@
+import { getPost } from "@/codeforafrica/lib/data/utils/posts";
+
+async function stories(api, context) {
```
kilemensi, line 3: Similar to opportunities, is this for single story or stories?

apps/codeforafrica/src/lib/data/blockify/postList.js
```diff
@@ -0,0 +1,29 @@
+import {
+  getPosts,
+  formatPost,
+  getTagsByPrimaryTag,
+} from "@/codeforafrica/lib/data/utils/posts";
+
+async function postList(block, api, context) {
```
kilemensi, line 7: Why `postList` instead of `posts`?

apps/codeforafrica/src/components/OpportunityCardList/OpportunityCardList.js
```diff
@@ -5,15 +5,19 @@ import OpportunityCard from "@/codeforafrica/components/OpportunityCard";
...
     if (!opportunities?.length) {
       return null;
     }
     return (
       <Stack spacing={{ xs: 5, md: 10 }} {...other} ref={ref}>
         {opportunities.map((opportunity) => (
-          <OpportunityCard {...opportunity} key={opportunity.id} />
+          <OpportunityCard
+            {...opportunity}
+            key={opportunity.id}
```
kilemensi, line 18:
```suggestion
            key={opportunity.slug}
```
slug, href should be unique, no?

apps/codeforafrica/src/components/Opportunities/Opportunities.js
```diff
@@ -41,22 +45,24 @@ const Opportunities = React.forwardRef(function Opportunities(
...
+      const { posts: results, pagination } = data;
       setCount(pagination.count);
       setOpportunities([...results]);
     }
   }, [data]);
 
   useEffect(() => {
-    router.push(queryParams, undefined, {
+    const [pathname] = router.asPath.split("?");
+    const url = pathname ? `${pathname}${queryParams}` : queryParams;
```
kilemensi, line 64:
```suggestion
    const [pathname] = window.location;
    const url = `${pathname}${queryParams}`;
```

apps/codeforafrica/src/lib/data/utils/posts.js
kelvinkipruto: @koechkevin We do not fetch all the tags as the tags are re-used in multiple places, for example in opportunities, stories and even projects. In the stories page, for instance, we'd want to fetch just all tags for posts tagged as stories.

apps/codeforafrica/src/components/CMSContent/CMSContent.test.js
```diff
@@ -9,7 +9,7 @@ import theme from "@/codeforafrica/theme";
 const render = createRender({ theme });
 
 const defaultProps = {
-  children: "<h1>Hello</h1><p>World</p>",
+  children: [],
```
kilemensi, line 12: Why are we testing with empty values?

apps/codeforafrica/src/lib/data/utils/posts.js
kelvinkipruto: @kilemensi Yes. We need to remove the excused tags from the list

apps/codeforafrica/src/lib/data/pagify/stories.js
kelvinkipruto: This is to handle single story and opportunity

apps/codeforafrica/src/lib/data/utils/posts.js
kilemensi: Removing primary tag is different from formatting. Removing primary tag could be a simple as skipping it in the front-end.

Formatting function sorts, etc., and if there are many posts with a lot of tags, this could slow. My initial understanding was may be only the main tags filter on the stories would need this sorting
Screenshot 2023-09-28 at 16 12 38

... thinking about it more, not sure whether sorting these tags by count vs alphabetically would make more sense to site visitors.

apps/codeforafrica/src/lib/data/pagify/stories.js
kilemensi: Then the function/file should be named `story` and not `stories`.

apps/codeforafrica/src/lib/data/utils/posts.js
kelvinkipruto: Sorting alphabetically should be the best way

apps/codeforafrica/src/lib/data/utils/posts.js
kelvinkipruto: @kilemensi This is the issue I mentioned in the meeting we need to force the link to three slugs.

apps/codeforafrica/src/payload/collections/Authors.js
```diff
@@ -14,6 +14,12 @@ const Authors = {
       localized: false,
       required: true,
     },
+    {
+      name: "bio",
+      type: "textarea",
+      label: "Bio",
+      required: true,
```
kilemensi, line 21:
```suggestion
```

Shouldn't be required.

apps/codeforafrica/src/pages/api/v1/posts.page.js
```diff
@@ -0,0 +1,14 @@
+import { getPosts } from "@/codeforafrica/lib/data/utils/posts";
+import api from "@/codeforafrica/lib/payload";
+
+export default async function handler(req, res) {
+  const { primaryTag, ...other } = req.query;
+  if (!primaryTag)
+    return res.status(500).json({ error: "Primary Tag is required" });
```
kilemensi, line 7:
```suggestion
  if (!primaryTag) {
    return res.status(400).json({ error: "Primary Tag is required" });
  }
```

apps/codeforafrica/src/lib/data/utils/posts.js
```diff
@@ -0,0 +1,180 @@
...
+    where: {
+      "tags.name": {
+        like: primaryTag,
+      },
+    },
+  });
+  const allTags = posts.reduce((acc, story) => {
+    const { tags = [] } = story;
+    return [...acc, ...tags.map(({ name, slug }) => ({ name, slug }))];
+  }, []);
```
kilemensi, line 175:
```suggestion
  const allTags = posts.flatMap((post) => post.tags ?? []);
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/flatMap

apps/codeforafrica/src/lib/data/utils/posts.js
```diff
@@ -0,0 +1,180 @@
...
+
+  const tags = formatTags(allTags);
+
+  return tags;
```
kilemensi, line 179:
```suggestion
  return formatTags(allTags);
```

apps/codeforafrica/src/lib/data/utils/posts.js
```diff
@@ -0,0 +1,180 @@
...
+    (tag) => !excludedTags.includes(tag.name.toLowerCase()),
+  );
+
+  const distinctTags = [
+    ...new Set(filteredTags.map((tag) => tag.name.toLowerCase())),
+  ];
+
+  const sortedTags = distinctTags.sort().map((tagName) => {
+    return filteredTags.find((tag) => tag.name.toLowerCase() === tagName);
+  });
```
kilemensi, line 16:
```suggestion
  const filteredTags = tags.filter(
    (tag) => !excludedTags.includes(tag.name.toLowerCase()),
  );
  const distinctTags = [
    ...new Set(filteredTags.map((tag) => tag.name.toLowerCase())),
  ];
  const sortedTags = distinctTags.sort().map((tagName) => {
    return filteredTags.find((tag) => tag.name.toLowerCase() === tagName);
  });
  
```

1. Why so many empty lines between lines of code?
2. Can you also describe what are we really trying to do here? I see we're also sorting tags in the UI so what does this sort add?

apps/codeforafrica/src/lib/data/pagify/story.js
```diff
@@ -0,0 +1,16 @@
+import { getPost } from "@/codeforafrica/lib/data/utils/posts";
+
+async function story(api, context) {
```
kilemensi, line 3: Functionally, is there any difference between this and opportunity? Can't they use the same function i.e. `post`?

apps/codeforafrica/src/lib/data/blockify/posts.js
```diff
@@ -0,0 +1,29 @@
...
+  getPosts,
+  formatPost,
+  getTagsByPrimaryTag,
+} from "@/codeforafrica/lib/data/utils/posts";
+
+async function posts(block, api, context) {
+  const { primaryTag, labels, stories = {} } = block;
+  const { query } = context;
+  const { featured, title } = stories;
+  const featuredStory = featured && (formatPost(featured, primaryTag) || null);
```
kilemensi, line 11:
```suggestion
  const { featured: featuredStory, title } = stories;
  const featured = featuredStory ? formatPost(featuredStory, primaryTag) : null;
```

`formatPost` already returns null, no?

apps/codeforafrica/src/lib/data/blockify/posts.js
```diff
@@ -0,0 +1,29 @@
...
+    ...query,
+  };
+  const { posts: list, pagination } = await getPosts(api, options, primaryTag);
+  const tags = await getTagsByPrimaryTag(api, primaryTag);
+  return {
+    title,
+    labels,
+    tags,
+    featured: featuredStory || null,
+    posts: list,
```
kilemensi, line 22:
```suggestion
  const { posts, pagination } = await getPosts(api, options, primaryTag);
  const tags = await getTagsByPrimaryTag(api, primaryTag);
  return {
    title,
    labels,
    tags,
    featured,
    posts,
```

Why rename `posts` to `list` only to rename it back to `posts`?

apps/codeforafrica/src/lib/data/blockify/index.js
```diff
@@ -3,15 +3,15 @@ import hero from "./hero";
...
+import posts from "./posts";
 
 const propsifyBlockBySlug = {
   "get-involved": getInvolved,
   hero,
   "meet-our-team": meetOurTeam,
   "our-impact": ourImpact,
-  stories,
   "our-team": ourTeam,
+  "post-list": posts,
```
kilemensi, line 14: I think I asked, why `post-list` instead of just `posts`?

apps/codeforafrica/src/components/Opportunities/Opportunities.js
```diff
@@ -41,22 +46,24 @@ const Opportunities = React.forwardRef(function Opportunities(
...
+      const { posts: results, pagination } = data;
       setCount(pagination.count);
       setOpportunities([...results]);
     }
   }, [data]);
 
   useEffect(() => {
-    router.push(queryParams, undefined, {
+    const { pathname } = window.location;
+    const url = pathname ? `${pathname}${queryParams}` : queryParams;
```
kilemensi, line 65:
```suggestion
    const url = `${pathname}${queryParams}`;
```

`pathname` should always be there, no?

apps/codeforafrica/src/components/Articles/Articles.js
```diff
@@ -39,27 +46,27 @@ const Articles = React.forwardRef(function Articles(props, ref) {
...
+      const { posts: results, pagination } = data;
       setCount(pagination.count);
       setArticles(results);
     }
   }, [data, filtering]);
 
   useEffect(() => {
-    const [pathname] = router.asPath.split("?");
+    const { pathname } = window.location;
     const url = pathname ? `${pathname}${queryParams}` : queryParams;
```
kilemensi, line 70: `pathname` should always be there, no?

apps/codeforafrica/src/components/ArticlePage/ArticlePage.js
```diff
@@ -8,24 +8,28 @@ import ArticleHeader from "@/codeforafrica/components/ArticleHeader";
...
-  html,
-  publishedAt,
+  coverImage: { src },
+  content,
+  publishedOn,
   primaryTag,
 }) {
+  const sortedTags = tags
+    .sort((a, b) => a.name.localeCompare(b.name))
+    .filter((tag) => !equalsIgnoreCase(tag.name, primaryTag));
```
kilemensi, line 25: I saw sorting in `lib` but I see it again here.

apps/codeforafrica/src/lib/data/utils/posts.js
kelvinkipruto: @kilemensi The initial plan was to remove duplicates, filter out the primary Tags and sort them. I think I can update it to remove duplicates and sort them, then remove the primary Tag in the UI

apps/codeforafrica/src/lib/data/pagify/story.js
kelvinkipruto: @kilemensi The only difference is that in `story` we have to include `Recent Stories`. I think I can set a different block for that.

apps/codeforafrica/src/lib/data/blockify/posts.js
kelvinkipruto: @kilemensi eslint complains as the function is already name posts

apps/codeforafrica/src/lib/data/blockify/index.js
kelvinkipruto: @kilemensi Because we are using the slug set in the CMS to match blocks. Payload does not allow us to use similar slugs. We already used `posts` slug for the [Posts collection](https://github.com/CodeForAfrica/ui/blob/25a20dbaffe1f026eff99bf1b519a6ccd4679ab8/apps/codeforafrica/src/payload/collections/Posts.js#L8) so we used [post-list](https://github.com/CodeForAfrica/ui/blob/25a20dbaffe1f026eff99bf1b519a6ccd4679ab8/apps/codeforafrica/src/payload/blocks/Posts.js#L13) for the block

apps/codeforafrica/src/lib/data/blockify/posts.js
kilemensi: ... cool, add a single line comment explaining why you're renaming it then.

apps/codeforafrica/src/lib/data/blockify/index.js
kilemensi: ... again, have added a comment on the block explaining this?

kelvinkipruto: @kilemensi Not sure I follow. This is the only [comment](https://github.com/CodeForAfrica/ui/pull/617#discussion_r1340047858) I can see on the related block

kilemensi: I mean: Can you add a single line comment on the block explaining why we're not naming it posts?

kilemensi (approved):

koechkevin (approved):

PR #619, opened 2023-10-03
Review of @/codeforafrica CMS

kilemensi:
## Description

As we prepare to hand over the CMS to the content team, this PR reviews the current state of the CMS and makes necessary adjustments. Things that this PR addressed include:

  - [x] Admin organisation: Moving away from non-descriptive Collections/Globals into Publication, Organisation and Website (more ideas are welcome as always),
  - [x] Custom users and emails: We should be able to define roles as well as allow users to reset their passwords etc.
  - [x] Enables localisation: While not used, it's important to enable it now when we don't have any content yet rather than in the future where we may have to create complex content migration scripts.

## Type of change

## Screenshots

Screenshot 2023-10-03 at 16-42-36 Editing - User - Payload

## Checklist:


koechkevin (approved):

kilemensi:
@kelvinkipruto ?

kelvinkipruto (approved):
Look good.
One small issue when creating a new user, we have to reselect the role as the default one throws an error `users validation failed: roles: Cast to string failed for value \"[ 'editor' ]\" (type Array) at path \"roles\""`
Another issue we can maybe look into in future is if a user has not been verified, it throws wrong credential. A better error would be for them to verify their email.

apps/codeforafrica/src/payload/collections/Users.js
```diff
@@ -0,0 +1,58 @@
...
+        },
+      ],
+    },
+    {
+      name: "roles",
+      type: "select",
+      // Save this field to JWT so we can use from `req.user`
+      saveToJWT: true,
+      required: true,
+      defaultValue: [ROLE_DEFAULT],
```
koechkevin, line 47: @kelvinkipruto This shouldn't be an array since hasMany is false. Pushing a fix.

kilemensi: I think we should turn `hasMany` back to `true` @koechkevin / @kelvinkipruto . If you look at [access methods](https://github.com/CodeForAfrica/ui/blob/6783676785229949c041355282dde1f3e25ed41e/apps/codeforafrica/src/payload/access/isAdminOrSelf.js#L7), they expect roles to be an array.

... not sure what I was testing there.

PR #620, opened 2023-10-03
Ft/Featured stories

kelvinkipruto:
## Description
This PR implements the Featured stories block

## Type of change

## Screenshots
image

image

## Checklist:


apps/codeforafrica/src/payload/blocks/PostsOverView.js
```diff
@@ -0,0 +1,66 @@
...
+      required: true,
+      type: "text",
+    },
+    {
+      name: "primaryTag",
+      label: "Post Type",
+      required: true,
+      type: "select",
+      options: primaryTags,
+      hasMany: false,
```
kilemensi, line 31: I think this block is only for stories... we don't have a featured view for opportunities in the designs.

apps/codeforafrica/src/payload/blocks/PostsOverView.js
```diff
@@ -0,0 +1,66 @@
...
+      required: true,
+    },
+    {
+      name: "posts",
+      type: "relationship",
+      relationTo: "posts",
+      hasMany: true,
+      minRows: 3,
+      maxRows: 3,
```
kilemensi, line 46: These needs to be posts of type stories only.

You can also make this a group as well
```js
stories: {
  items: // This is the relationship field
  action: // This is the action linkGroup field
}
```

kilemensi (commented):
👍🏽 

---

I think lets call the block NewsAndStories as well (`news-stories` slug).

apps/codeforafrica/src/payload/blocks/FeaturedStories.js
```diff
@@ -0,0 +1,77 @@
...
+      en: "Featured Stories",
+    },
+    plural: {
+      en: "Featured Stories",
+    },
+  },
+  fields: [
+    {
+      name: "title",
+      label: "Title",
```
kilemensi, line 18:
```suggestion
```

If label is the same as title, payload will user name. No need to add it.

kilemensi (approved):

koechkevin (approved):

PR #621, opened 2023-10-03
ft/Improve Posts

kelvinkipruto:
## Description
This PR implements some improvements on posts

1. Introduces a posts settings global to control whether to show recent posts on individual pages
2. Fixes a bug whereby you cannot sort posts by tag
3. Remove the featured post from the list of posts if it is available. 

## Type of change

## Screenshots
image

image

## Checklist:


kilemensi (commented):
👍🏽 

--

1. Is this block for [Projects](https://cfa.dev.codeforafrica.org/projects/civicsignal) or [Stories](https://cfa.dev.codeforafrica.org/stories/battle-for-gender-equality-in-african-media-continues)
2. If it's related stories, we'd need a way of relating. Once way is making sure they share tags. If it's recent, we need to fetch ordered by publication date descending or something.
3. How will the content editors add this block in the CMS? We don't have individual pages for projects or stories in the CMS.

apps/codeforafrica/src/payload/blocks/Posts.js
```diff
@@ -73,6 +73,16 @@ const Posts = {
           type: "text",
           defaultValue: "Read More",
         },
+        {
```
koechkevin, line 76: @kelvinkipruto  thinking out loud, is recent~Stories expected to be fixed by post? why not fetch from the collection ordering by publishedOn ?

apps/codeforafrica/src/lib/data/pagify/post.js
```diff
@@ -1,10 +1,39 @@
...
+    api,
+    {
+      limit: 3,
+      sort: "publishedOn",
+    },
+    "stories",
+  );
+  const block = parentPage.blocks.find(
+    ({ blockType }) => blockType === "post-list",
+  );
```
koechkevin, line 18: Let us check if block exists or else we will end up with undefined errors.

apps/codeforafrica/src/payload/blocks/Posts.js
kelvinkipruto: @koechkevin I'm already fetching [ordered by publishedOn](https://github.com/CodeForAfrica/ui/blob/95c9b955435e94ab67696b83c8e6ede4c80cc393/apps/codeforafrica/src/lib/data/pagify/post.js#L8-L15). This is just the title for the Recent Stories

kelvinkipruto:
@kilemensi @koechkevin This is now ready.

kilemensi (commented):
My quick review on this @kelvinkipruto 

We should change website/settings to be:

 - Settings
   - Site
   - Publication
   - Users

**Site**: This is the current Settings global, just a rename i.e. All site-wide settings will be here.
**Publication**: This settings will cover everything under **Publication**. For now we can have just have Posts (as a tab) and under that, we can have Recent Stores, show recent stories checkbox and if checked, allow editors to set title for recent stories. Similar for Recent Opportunities.
**Users**: Remain as is.

kilemensi:
Does my [comment](https://github.com/CodeForAfrica/ui/pull/621#pullrequestreview-1661709936) make sense @kelvinkipruto?

kelvinkipruto:
@kilemensi Yes. Currently updating it.

apps/codeforafrica/src/payload/globals/Site/index.js
```diff
@@ -2,8 +2,9 @@ import EngagementTab from "./EngagementTab";
 import GeneralTab from "./GeneralTab";
 import NavigationTab from "./NavigationTab";
 
-const Settings = {
-  slug: "settings",
+const Site = {
+  slug: "site",
```
kelvinkipruto, line 6: @kilemensi Not sure about it, but since we already have some data in prod, do we need to update this slug, or should we keep it as it was: `settings`?

kilemensi: Yes, let's update it @kelvinkipruto. I don't think the content team has had time to really dig in just yet. Keep a copy of the JSON data handy before deploying & we can update the data after deployment.

kilemensi: For all the above changes, we can use slug as is i.e. `site`, `publication`, etc., or prepend `setting-` to show these are settings related i.e. `settings-site`, `settings-publication`, etc.

I'll let you be the architect.

apps/codeforafrica/src/components/Articles/Articles.js
```diff
@@ -35,7 +35,7 @@ const Articles = React.forwardRef(function Articles(props, ref) {
   const [q, setQ] = useState();
   const [filtering, setFiltering] = useState(false);
   const [tag, setTag] = useState(allTag);
-  const queryParams = useFilterQuery({ page, q, tag: tag.slug });
+  const queryParams = useFilterQuery({ page, q, tag });
```
kilemensi, line 38: Why the change from `tag.slug` to just `tag`?

apps/codeforafrica/src/components/Articles/useArticles.js
```diff
@@ -4,11 +4,13 @@ import useFilterQuery from "@/codeforafrica/components/useFilterQuery";
...
-function useArticles(params, primaryTag) {
+function useArticles(params, { primaryTag, featured }) {
   const queryParams = useFilterQuery(params);
   const query = queryParams
-    ? `${queryParams}&primaryTag=${primaryTag}`
-    : `?primaryTag=${primaryTag}`;
+    ? `${queryParams}&primaryTag=${primaryTag}${
+        featured ? `&featured=${featured}` : ""
+      }`
+    : `?primaryTag=${primaryTag}${featured ? `&featured=${featured}` : ""}`;
```
kilemensi, line 13: This can be improved e.g.
```js
  const filterQuery = queryParams ? `?${queryParams}&` : '?';
  const primaryTagQuery = `primaryTag=${primaryTag}`;
  const featuredQuery = featured ? `&featured=${featured}&` : '';
  const query = `${filterQuery}${primaryTagQuery}${featuredQuery}`
```

apps/codeforafrica/src/lib/data/blockify/posts.js
```diff
@@ -11,6 +11,13 @@ async function posts(block, api, context) {
   const featured = featuredStory ? formatPost(featuredStory, primaryTag) : null;
   const options = {
     ...query,
+    ...(featured && {
+      where: {
+        slug: {
+          not_equals: featured.slug,
```
kilemensi, line 17: Let's make sure the `slug` field has [`index: true`](https://payloadcms.com/docs/fields/text#config) if we're going to be using it in queries.

apps/codeforafrica/src/payload/globals/Publication/PostsTab.js
```diff
@@ -0,0 +1,82 @@
+const PostsTab = {
+  label: "Posts",
```
kilemensi, line 2:
```suggestion
  label: "Post",
```

Since this is for individual posts, may be we should name it so.

apps/codeforafrica/src/payload/globals/Publication/PostsTab.js
```diff
@@ -0,0 +1,82 @@
...
+          label: "Primary Tag",
+          type: "text",
+          required: true,
+          defaultValue: "stories",
+          admin: {
+            hidden: true,
+          },
+        },
+      ],
+    },
```
kilemensi, line 40: 1. UI needs to look like the rest of the globals
2. localisation must be enabled at top levels similar to other globals
```suggestion
    {
      type: "collapsible",
      label: "Story",
      fields: [
        {
          name: "recent-stories",
          label: "Recent stories",
          type: "group",
          localized: true,
          fields: [
            {
              name: "enabled",
              label: "Show recent stories",
              type: "checkbox",
              required: true,
              defaultValue: false,
            },
            {
              name: "title",
              label: "Title",
              type: "text",
              required: true,
              defaultValue: "Recent Stories",
              admin: {
                condition: (_, data) => data.showRecent,
              },
            },
            {
              name: "primaryTag",
              label: "Primary Tag",
              type: "text",
              required: true,
              defaultValue: "stories",
              admin: {
                hidden: true,
              },
            },
          ],
          admin: {
            className: "group-field-nested",
          },
        },
      ],
    },
```
Screenshot 2023-10-09 at 09 49 58

apps/codeforafrica/src/payload/globals/Publication/index.js
```diff
@@ -0,0 +1,20 @@
+import PostsTab from "./PostsTab";
+
+const Publication = {
+  slug: "publication",
+  label: "Publication",
+  access: {
+    read: () => true,
+  },
+  admin: {
+    group: "Website",
```
kilemensi, line 10:
```suggestion
    group: "Settings",
```

`admin.group` in `Users` and `Site` needs to be updated as well.

koechkevin (commented):
@kelvinkipruto are we planning to add migrations since we are changing slug and production data already exists?

kelvinkipruto:
> are we planning to add migrations since we are changing slug and production data already exists?

@koechkevin I have backed up the JSON data currently in use. We will use it when we merge this.

apps/codeforafrica/src/components/Articles/Articles.js
kelvinkipruto: @kilemensi #610 [updated the `useFilterQuery`](https://github.com/CodeForAfrica/ui/blob/e3eb9f96e118f6cdf0dee21d7764afb0c8270b6b/apps/codeforafrica/src/components/useFilterQuery/useFilterQuery.js#L17-L19) hook to take the whole tag, rather than just the slug. That is why currently in production you can't filter stories & opportunities by Tag.

apps/codeforafrica/src/components/Articles/useArticles.js
```diff
@@ -4,11 +4,13 @@ import useFilterQuery from "@/codeforafrica/components/useFilterQuery";
...
 
-function useArticles(params, primaryTag) {
+function useArticles(params, { primaryTag, featured }) {
   const queryParams = useFilterQuery(params);
-  const query = queryParams
-    ? `${queryParams}&primaryTag=${primaryTag}`
-    : `?primaryTag=${primaryTag}`;
+  const separator = queryParams ? "" : "?";
+  const filterQuery = queryParams ? `${queryParams}&` : "";
+  const primaryTagQuery = `primaryTag=${primaryTag}`;
```
kilemensi, line 11: I think this is not right-ish. `separator` should always be `?` because there will always be `primaryTagQuery` even where there isn't `queryParams`.

apps/codeforafrica/src/lib/data/utils/posts.js
```diff
@@ -117,29 +118,54 @@ export async function getPost(api, slug, primaryTag) {
...
+      }),
+      primaryTag,
+      blockType: "article",
+      ...other,
+    },
+  ];
+
+  const globalPostSettings = await api.findGlobal("settings-publication");
+  const primaryTagGlobalSettings = globalPostSettings?.[primaryTag] ?? {};
+  const { showRecent, title: recentTitle } = primaryTagGlobalSettings;
```
kilemensi, line 145:
```suggestion
  const publicationSettings = await api.findGlobal("settings-publication");
  const primaryTagPostSettings = publicationSettings?.[primaryTag] ?? {};
  const { showRecent, title: recentTitle } = primaryTagPostSettings;
```

No need to use global.

apps/codeforafrica/src/pages/api/v1/posts.page.js
```diff
@@ -2,12 +2,25 @@ import { getPosts } from "@/codeforafrica/lib/data/utils/posts";
...
   try {
-    const data = await getPosts(api, other, primaryTag);
+    const data = await getPosts(
+      api,
+      {
+        ...other,
+        ...(featured && {
+          where: {
+            slug: {
+              not_equals: featured,
```
kilemensi, line 17: Does this work well even when `featured` is `undefined`?

apps/codeforafrica/src/payload/collections/Posts.js
```diff
@@ -39,7 +40,9 @@ const Posts = {
       },
     }),
     content({ minRows: 1, required: true, localized: true }),
-    slug(),
+    slug({
+      index: true,
```
kilemensi, line 44: Add this to `payload/fields/slug.js`. Should be beneficial to everything that's using a `slug`.

apps/codeforafrica/src/payload/collections/Posts.js
```diff
@@ -23,6 +23,7 @@ const Posts = {
     description: "Stories and Opportunities",
     group: "Publication",
     useAsTitle: "title",
+    listSearchableFields: ["content", "excerpt"],
```
kilemensi, line 26: No need to include `title`?

apps/codeforafrica/src/payload/globals/Publication/PostTab.js
```diff
@@ -0,0 +1,95 @@
...
+            {
+              name: "primaryTag",
+              label: "Primary Tag",
+              type: "text",
+              required: true,
+              defaultValue: "stories",
+              admin: {
+                hidden: true,
+              },
+            },
```
kilemensi, line 40: Thinking out loud, do we need this field if the group is already named "stories" or "opportunities" i.e. `group.name` matches `primaryTag`.

kilemensi (approved):

apps/codeforafrica/src/payload/globals/Publication/PostTab.js
kelvinkipruto: I agree. We should remove this as it is not used anywhere.

apps/codeforafrica/src/payload/collections/Posts.js
kelvinkipruto: @kilemensi You can already search by title. I think it is made possible by `useAsTitle: "title"`

apps/codeforafrica/src/payload/collections/Posts.js
kelvinkipruto: Turns out this is already indexed

kelvinkipruto: Turns out this is already indexed

apps/codeforafrica/src/pages/api/v1/posts.page.js
kelvinkipruto: @kilemensi I Just tested it out and it works. Also, we are making sure to add `featured` only if a featured article exists.

apps/codeforafrica/src/components/Articles/useArticles.js
kelvinkipruto: @kilemensi [`useFilterQuery` already](https://github.com/CodeForAfrica/ui/blob/73018e2697cd8ed2f84fc73f786c27f1a961ec56/apps/codeforafrica/src/components/useFilterQuery/useFilterQuery.js#L21) appends `?` if there are `queryParams`. If `queryParams` does not exist, then we have to use create `?` for the `primaryTagQuery`

kilemensi: If it's there, then all good @kelvinkipruto

PR #631, opened 2023-10-06
Fix go to repository broken link

koechkevin:
## Description

The "go to repo" button on the "tools individual" page of the charter site is not redirecting to GitHub, neither is the expand icon.

- nestPageUnderCollection function overrides link field in a component. This PR introduces docLink field to avoid this error.
Fixes # [issue](https://code4africa.slack.com/archives/C04AJ0124H1/p1696578745144429)

## Type of change

## Checklist:

### DIFF
--- a/apps/charterafrica/src/components/Tool/Tool.js
+++ b/apps/charterafrica/src/components/Tool/Tool.js
@@ -63,7 +63,7 @@ const Tool = React.forwardRef(function Tool(props, ref) {
                   {name}
                 </RichTypography>
                 {externalLink?.href ? (
-                  <Link href={externalLink?.href}>
+                  <Link href={externalLink.href}>
                     <SvgIcon
                       inheritViewBox
                       component={ExternalLink}
--- a/apps/charterafrica/src/components/Tool/Tool.js
+++ b/apps/charterafrica/src/components/Tool/Tool.js
@@ -63,8 +63,8 @@ const Tool = React.forwardRef(function Tool(props, ref) {
                   {name}
                 </RichTypography>
                 <Button
-                  component={externalLink.href ? Link : undefined}
-                  href={externalLink.href}
+                  component={externalLink?.href ? Link : undefined}
+                  href={externalLink?.href}
                 >
                   <SvgIcon
                     inheritViewBox
--- a/apps/charterafrica/src/components/Tool/Tool.js
+++ b/apps/charterafrica/src/components/Tool/Tool.js
@@ -18,7 +18,6 @@ const Tool = React.forwardRef(function Tool(props, ref) {
   const {
     image,
     name,
-    link,
     organisation,
     theme,
     operatingCountries,
@@ -34,6 +33,7 @@ const Tool = React.forwardRef(function Tool(props, ref) {
     partners,
     tools,
     toolsTitle,
+    externalLink,
   } = props;
   return (
     <Box bgcolor="common.white" ref={ref}>
@@ -62,7 +62,10 @@ const Tool = React.forwardRef(function Tool(props, ref) {
                 <RichTypography color="neutral.dark" variant="h2SemiBold">
                   {name}
                 </RichTypography>
-                <Link href={link.href}>
+                <Button
+                  component={externalLink.href ? Link : undefined}
+                  href={externalLink.href}
+                >
                   <SvgIcon
                     inheritViewBox
                     component={ExternalLink}
@@ -73,7 +76,7 @@ const Tool = React.forwardRef(function Tool(props, ref) {
                       width: 32,
                     }}
                   />
-                </Link>
+                </Button>
               </Box>
             </Grid>
             <Grid item>
--- a/apps/charterafrica/src/components/Tool/Tool.js
+++ b/apps/charterafrica/src/components/Tool/Tool.js
@@ -62,21 +62,20 @@ const Tool = React.forwardRef(function Tool(props, ref) {
                 <RichTypography color="neutral.dark" variant="h2SemiBold">
                   {name}
                 </RichTypography>
-                <Button
-                  component={externalLink?.href ? Link : undefined}
-                  href={externalLink?.href}
-                >
-                  <SvgIcon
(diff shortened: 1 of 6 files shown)


kelvinkipruto (approved):

apps/charterafrica/src/lib/data/common/processPageTools.js
```diff
@@ -34,9 +34,15 @@ const queryBuilder = (query) => {
 };
 
 const getRepoLink = (tool) => {
-  switch (tool.source && tool.externalId) {
+  if (!tool.externalId) {
+    return null;
+  }
```
kilemensi, line 39: Isn't this check done when we're pulling data from Airtable?

apps/charterafrica/src/lib/data/common/processPageTools.js
```diff
@@ -34,9 +34,15 @@ const queryBuilder = (query) => {
...
+  switch (tool.source) {
     case "github":
-      return `https://github.com/${tool.externalId}`;
+      return `https://github.com/${tool.externalId.replace(
+        "https://github.com/",
+        "",
+      )}`;
```
kilemensi, line 45: What is happening here?

apps/charterafrica/src/payload/utils/nestCollectionUnderPage.js
```diff
@@ -16,7 +16,7 @@ function nestCollectionUnderPage(pageSlug) {
     } catch (error) {
       // TODO(kilemensi): Add Sentry to payload & report errors
     }
-    return { ...doc, link: { href } };
+    return { ...doc, docLink: doc.link ?? null, link: { href } };
```
kilemensi, line 19: What has changed here?

apps/charterafrica/src/components/Tool/Tool.js
```diff
@@ -62,7 +62,10 @@ const Tool = React.forwardRef(function Tool(props, ref) {
                 <RichTypography color="neutral.dark" variant="h2SemiBold">
                   {name}
                 </RichTypography>
-                <Link href={link.href}>
+                <Button
+                  component={externalLink?.href ? Link : undefined}
+                  href={externalLink?.href}
+                >
```
kilemensi, line 68: Why the switch from Link to Button? If there is no external link, we'll just not show the link/icon. What is the point of showing a button that does nothing when clicked?

apps/charterafrica/src/lib/data/common/processPageTools.js
koechkevin: Most  tools have the whole url as externalID

apps/charterafrica/src/payload/utils/nestCollectionUnderPage.js
koechkevin: Here there's a field called link in tools which is overwritten by this link in afterread hook. So I'm adding this to ensure we don't lose that data

kilemensi: Again, wasn't this supposed to be handled when importing from Airtable?


> **Airtable + GitHub -> CMS**: Do all necessary pre-processing and standardisation
> **CMS -> Website**: Localisation/formatting for display only.


What am I missing?

koechkevin: You're right! My bad.

apps/charterafrica/src/components/Tool/Tool.js
```diff
@@ -62,18 +62,20 @@ const Tool = React.forwardRef(function Tool(props, ref) {
...
-                    sx={{
-                      color: "text.primary",
-                      fill: "none",
-                      height: 32,
-                      width: 32,
-                    }}
-                  />
-                </Link>
+                {externalLink?.href ? (
+                  <Link href={externalLink?.href}>
```
kilemensi, line 66:
```suggestion
                  <Link href={externalLink.href}>
```

We've just confirmed that `externalLink.href` exists on the previous line.

apps/charterafrica/src/payload/utils/nestCollectionUnderPage.js
kilemensi: I'm assuming we've tested this and all is good. If yes, please resolve the conversation.

apps/charterafrica/src/lib/data/common/processPageTools.js
```diff
@@ -113,6 +104,9 @@ async function processPageSingleTool(page, api, context) {
         commitText: filterLabels.lastCommit,
         forksText: filterLabels.forks,
         starsText: filterLabels.stars,
+        externalLink: {
+          href: tool.docLink ?? null,
```
kilemensi, line 108: Just for my understanding, where does this dockLink come from or why it doesn't behave like other links?

kilemensi (approved):

apps/charterafrica/src/lib/data/common/processPageTools.js
koechkevin: @kilemensi we are reading this field from airtable (website link).
We have this field saved in cms as `link` before we added the after read hook.

kilemensi: Can't we add afterRead hook now? The thing about afterRead hook is it runs when you read from the database so it doesn't change what's stored in the database.

koechkevin: We have doc in the database as
```js
 {
  link: "https://datareview.openownership.org/" // From airtable
}
```

after read hook added this

```js
...
 return { ...doc, link: { href: "/tools/openownership" } };
```

Clicking the icon therefore href'ed to the page itself hence no action.

koechkevin: Since this is an urgent fix, changing field in the database would maybe have required more time since involves having to wait for the cron to update the list. I will create another PR to correct this.

PR #632, opened 2023-10-09
Code for Africa Our offices

koechkevin:
## Description

- Adds our offices block.
- Adds Location collection to cms for selection on our offices

Fixes # (issue)

## Type of change

## Screenshots
image
Screenshot 2023-10-06 at 09 11 34

## Checklist:


apps/codeforafrica/src/payload/collections/Locations.js
kelvinkipruto, line 1: Perhaps a better name would be `Offices`

apps/codeforafrica/payload.config.ts
```diff
@@ -57,6 +58,7 @@ export default buildConfig({
     Donors,
     GuidingPrinciples,
     Impact,
+    Locations,
```
kelvinkipruto, line 61: Would a better option be to make it a global since it's a global setting?

kilemensi: We have more than one office so I think a `collection` `Offices` is alright.

Global would also work if we use an array of offices or something but collection is more straight forward for content editors as it works similar to pretty much every other content on the CMS.

apps/codeforafrica/src/lib/data/blockify/ourOffices.js
```diff
@@ -0,0 +1,31 @@
...
+    ...item,
+    position: {
+      lng: item?.location?.[0],
+      lat: item?.location?.[1],
+    },
+    center: {
+      lng: item?.location?.[0],
+      lat: item?.location?.[1],
+    },
+  }));
```
kilemensi, line 13:
```suggestion
  const addresses = offices?.map((item) => {
    let location = null;
    if (item?.location?.length)
      location = { lng: item.location.[0], lat: item.location.[1] };
    }
    return {
      ...item,
      position: location,
      center: location,
    };
  });
```

apps/codeforafrica/src/components/OfficeAddresses/OfficeAddresses.js
```diff
@@ -46,7 +50,7 @@ const OfficeAddresses = React.forwardRef(function OfficeAddresses(props, ref) {
           zIndex: -1,
         }}
       >
-        <GoogleMap {...map} {...activeAddress?.map} style={mapStyle} />
+        <GoogleMap {...map} {...activeAddress} style={mapStyle} />
```
kilemensi, line 53: Why passing `activeAddress`? Original code assumed `activeAddress` would have `map` & hence was passing `activeAddress.map`

apps/codeforafrica/src/components/AddressCard/AddressCard.js
```diff
@@ -38,12 +47,37 @@ const AddressCard = React.forwardRef(function AddressCard(props, ref) {
...
+            variant="body3"
+          >
+            {addressLine2}
+          </RichTypography>
+          <RichTypography
+            sx={{ color: "inherit", "& p": { m: 0 } }}
+            variant="body3"
+            component="span"
+          >
+            {zipcode},
```
kilemensi, line 67: Only include `,` if there is `zipCode`. From collection definition, `zipCode` is not required and hence may be empty.

kilemensi (approved):

apps/codeforafrica/src/payload/collections/Offices.js
```diff
@@ -0,0 +1,43 @@
+const Offices = {
+  slug: "offices",
+  admin: {
+    group: "Settings",
+    defaultColumns: ["title"],
```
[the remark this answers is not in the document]
koechkevin: @kilemensi Can't  there be a situation where we have more than 1 office in a City?

PR #634, opened 2023-10-11
Fix member tags

koechkevin:
## Description

Our team tags are not tags from  tags collection. They are string tags

Fixes # (issue)

## Type of change

## Checklist:


kilemensi:
Struggling with power this morning @koechkevin but from your description:

>Our team tags are are not tags from tags collection. They are string tags
>

What is our approach to fixing this then?

koechkevin:
> Struggling with power this morning @koechkevin but from your description:
> 
> > Our team tags are are not tags from tags collection. They are string tags
> 
> What is our approach to fixing this then?
@kilemensi 
My approach is to modify these strings into an object with `slug` and `name` properties

kelvinkipruto (approved):

apps/codeforafrica/src/lib/data/blockify/ourTeam.js
```diff
@@ -1,23 +1,32 @@
...
+function getStringTags(field, docs) {
+  const tags = sortTags(
+    docs
+      .map((item) =>
+        item[field] ? { name: item[field], slug: item[field] } : null,
+      )
+      .filter(Boolean),
+  );
+  return [{ name: "All", slug: "All" }, ...tags];
+}
```
kilemensi, line 18: What does `String` mean in getStringTags?

koechkevin: We want to get country tags for example, they exist as strings, not objects.

kilemensi: 1. Is there any other field besides country that this method is handling? If not, we should call getCountryTags
2. We were supposed to be using this [file](https://github.com/CodeForAfrica/ui/blob/main/apps/codeforafrica/src/lib/data/json/countries.js) I think to have country `name` -> `iso alpha3 code` as object. Not sure why we didn't do that.


I think we're using something like that in charterAFRICA (check somewhere around Map/Focal countries) or searching online should also give you an idea of what I'm [talking](https://github.com/lukes/ISO-3166-Countries-with-Regional-Codes) [about](https://github.com/lukes/ISO-3166-Countries-with-Regional-Codes). If it's not too late, we should switch to that format.

koechkevin: 2. We used this file as options in cms but it is stored as a string(value). We also do not want all countries listed for filtering. we only need those that exist in the team collection. @kilemensi

kilemensi: Yes @koechkevin, I full understand that. What I mean is:

1. We should switch the file used in the CMS to the one that has name and iso code instead of just name for both label and value.
2. Yes, we don't want to show all countries but once we get the string (which will be a code this time), we'll use this file to format it into object i.e {.name: "Name of the country", slug: iso code } as required by tag handling function.

apps/codeforafrica/src/lib/data/json/countries.js
```diff
@@ -1,1972 +1,4227 @@
...
+    slug: "Zimbabwe",
+    name: "Zimbabwe",
   },
 ];
 
 export const countriesByContinent = (continent) => {
   return countries
     .filter((country) => continent === country.continent)
-    .map(({ continent: cntnt, name, ...c }) => c);
+    .map(({ value, label }) => ({
```
kilemensi, line 4218: `code`, we store countries by code & not `value`. `alpha-3` to be specific

apps/codeforafrica/src/lib/data/blockify/ourTeam.js
```diff
@@ -1,23 +1,34 @@
+import { countries } from "@/codeforafrica/lib/data/json/countries";
 import { getMembers } from "@/codeforafrica/lib/data/utils/members";
+import { sortTags } from "@/codeforafrica/lib/data/utils/tags";
+import equalsIgnoreCase from "@/codeforafrica/utils/equalsIgnoreCase";
+
+const getCountryFromName = (name) =>
+  countries.find((c) => equalsIgnoreCase(c.slug, name)) ?? null;
```
kilemensi, line 7: Should be getCountryFromCode or getCountryFromAlpha3 to be specific

apps/codeforafrica/src/lib/data/blockify/ourTeam.js
```diff
@@ -1,23 +1,34 @@
...
+
+function getTeamTags(docs) {
+  const tags = sortTags(docs.map((item) => item.team).filter(Boolean));
+  return [{ name: "All", slug: "All" }, ...tags];
```
kilemensi, line 11: You can declare `const ALL_TAG = { name: "All", slug: "All" };` somewhere and reuse it... may be in `utils/tags` or something.

kilemensi (approved):
🚀 

---

Let's also quickly update the members country in CMS after deployment (I think there is about 4 members or so).

PR #670, opened 2024-01-09
Migrate payload version 2

koechkevin:
## Description

On OCTOBER 9, 2023, Payload CMS announced the release of [version 2](https://payloadcms.com/blog/payload-2-0).
Version 2 comes with a ton of features such as support for postgres, admin UI updates and support for migrations. Even though we are not going to use the postgres feature, support for migrations is a crucial for the maintenance of our app.

This PR updates payload version to 2.6.0 for charter Africa. I implemented this connected to dev database(Ideally a replica of the prod db.).

To test this PR,
This PR adds a [Makefile](https://github.com/CodeForAfrica/ui/blob/feature/upgrade-payload-to-v2/docker-compose.yml) and [charterafrica start script](https://github.com/CodeForAfrica/ui/blob/feature/upgrade-payload-to-v2/Makefile).

Execute the command below
`make charterafrica`.
The command will start charterafrica in production mode and attach docker database

⚠️To initialize the db, 
Generate keyfile by running
`openssl rand -base64 741 > ./mongo-keyfile`
`chmod 600 ./mongo-keyfile`

> Based on the [migration steps docs](https://github.com/payloadcms/payload/releases/tag/v2.0.0), there are some steps that require manual migration( [User preferences](https://payloadcms.com/docs/admin/preferences#managing-user-preferences) ). We do not store user preferences, so there is no need for manual migrations
## Type of change

## Checklist:


koechkevin:
@kelvinkipruto @kilemensi please test this to see if I have missed out anything

apps/charterafrica/server.ts
```diff
@@ -7,6 +7,9 @@ import next from "next";
 import nodemailerSendgrid from "nodemailer-sendgrid";
 import payload from "payload";
 import { Payload } from "payload/dist/payload";
+import dotenv from "dotenv";
+
+dotenv.config({ path: path.resolve(__dirname, "./.env.local") });
```
kelvinkipruto, line 12: @koechkevin Not sure why we are hardcoding `.env.local` here since its only meant for local environments.

apps/charterafrica/src/payload/fields/richText.js
```diff
@@ -43,12 +44,16 @@ async function mapLinkToHrefAfterRead({ req: { payload }, value }) {
   return insertHref(value, payload);
 }
 
-function richText(overrides) {
+function richText({ admin, ...overrides }) {
+  const editor = slateEditor({
```
kelvinkipruto, line 48: Great approach here. I had updated every file where `richText` was used 🤦🏿‍♂️ 🤦🏿‍♂️.

apps/charterafrica/payload.config.ts
```diff
@@ -73,6 +76,11 @@ const adapter = s3Adapter({
 
 export default buildConfig({
   serverURL: appURL,
+  editor: slateEditor({}),
+  db: mongooseAdapter({
+    url: process.env.MONGO_URL,
+    migrationDir: "./migrations/v2",
```
kilemensi, line 82: Shouldn't this be `src/migrations`? Or are they migrations for `just` `v1` to `v2` migration.

apps/charterafrica/server.ts
kilemensi: `loadEnvConfig` below this line should load all necessary `.env` files. We SHOULDN'T need this line.

kilemensi (commented):
👍🏽 

---

What about all the manual steps mentioned [here](https://github.com/payloadcms/payload/releases/tag/v2.0.0) e.g. drafts, preferences, etc.

apps/charterafrica/payload.config.ts
koechkevin: @kilemensi there exists a folder called migrations with a file for migrations we implemented in v1. To avoid conflicts, I thought it best to have v2 migrations in a folder called v2.

kilemensi: ... nah, all migrations should be in one folder so they get applied sequentially @koechkevin.

Unless you think the old migration won't run in the version (which would mean we need to decided what to do with them).

koechkevin: @kilemensi Yes, the old migration won't run in the new version. Meaning we have 2 options with it, delete the file if the migrations have already been applied or  rewrite the script to ensure it runs in the current version.

kilemensi:
> Payloads' database migrations rely on transactions and mongo transactions require a replicaset and not single node db.
>

@koechkevin 

1. Have we figured out how to make this work with Dokku 
2. Shouldn't we have introduce a `docker` `compose` setup for local testing/development with Mongo cluster? cc @thepsalmist

koechkevin:
> > Payloads' database migrations rely on transactions and mongo transactions require a replicaset and not single node db.
> 
> @koechkevin
> 
> 1. Have we figured out how to make this work with Dokku
> 2. Shouldn't we have introduce a `docker` `compose` setup for local testing/development with Mongo cluster? cc @thepsalmist

@kilemensi The error we were facing has nothing to do with mongodb replicaSets. It was a known [payload cms bug](https://github.com/payloadcms/payload/commit/21b9453cf4e6eebf145d89a0190942015658413d). Which was fixed in v2.7.0 released just last week.

kilemensi:
> @kilemensi The error we were facing has nothing to do with mongodb replicaSets. It was a known [payload cms bug](https://github.com/payloadcms/payload/commit/21b9453cf4e6eebf145d89a0190942015658413d). Which was fixed in v2.7.0 released just last week.

On one hand that's cool to hear but the PR still says database migrations relies on transactions which in turn relies on replica sets @koechkevin .

Does this mean Mongo migrations don't require replica sets?

koechkevin:
> > @kilemensi The error we were facing has nothing to do with mongodb replicaSets. It was a known [payload cms bug](https://github.com/payloadcms/payload/commit/21b9453cf4e6eebf145d89a0190942015658413d). Which was fixed in v2.7.0 released just last week.
> 
> On one hand that's cool to hear but the PR still says database migrations relies on transactions which in turn relies on replica sets @koechkevin .
> 
> Does this mean Mongo migrations don't require replica sets?

Yes

kilemensi:
> > Does this mean Mongo migrations don't require replica sets?
> 
> Yes

Can you link to where it confirms this @koechkevin? My quick Google-fu shows [articles](https://www.mongodb.com/community/forums/t/why-replica-set-is-mandatory-for-transactions-in-mongodb/9533/4) (oldish-yes) that say we do need replica set to enable transactions.

koechkevin:
> > > Does this mean Mongo migrations don't require replica sets?
> > 
> > 
> > Yes
> 
> Can you link to where it confirms this @koechkevin? My quick Google-fu shows [articles](https://www.mongodb.com/community/forums/t/why-replica-set-is-mandatory-for-transactions-in-mongodb/9533/4) (oldish-yes) that say we do need replica set to enable atomic transactions.

@kilemensi 
From [payload documentation on transactions](https://payloadcms.com/docs/database/transactions),

> By default, Payload will use transactions for all operations, as long as it is supported by the configured database. Database changes are contained within all Payload operations and any errors thrown will result in all changes being rolled back without being committed. When transactions are not supported by the database, Payload will continue to operate as expected without them.

kilemensi:
> @kilemensi From [payload documentation on transactions](https://payloadcms.com/docs/database/transactions),
> 
> > By default, Payload will use transactions for all operations, as long as it is supported by the configured database. Database changes are contained within all Payload operations and any errors thrown will result in all changes being rolled back without being committed. When transactions are not supported by the database, Payload will continue to operate as expected without them.

On the same doc that you've linked to @koechkevin :

>Note:
**MongoDB** _requires_ a connection to a `replicaset` in order to make use of transactions.
>

My take away is we need transaction (at least in PROD) and replicaset is the only way to achieve this with MongoDB.

koechkevin:
> > @kilemensi From [payload documentation on transactions](https://payloadcms.com/docs/database/transactions),
> > > By default, Payload will use transactions for all operations, as long as it is supported by the configured database. Database changes are contained within all Payload operations and any errors thrown will result in all changes being rolled back without being committed. When transactions are not supported by the database, Payload will continue to operate as expected without them.
> 
> On the same doc that you've linked to @koechkevin :
> 
> > Note:
> > **MongoDB** _requires_ a connection to a `replicaset` in order to make use of transactions.
> 
> My take away is we need transaction (at least in PROD) and replicaset is the only way to achieve this with MongoDB.

@kilemensi We'd be limited by dokku-mongo plugin due to [this comment on replicaSets](https://github.com/dokku/dokku-mongo/issues/110#issuecomment-428804387)

kilemensi:
> @kilemensi We'd be limited by dokku-mongo plugin due to [this comment on replicaSets](https://github.com/dokku/dokku-mongo/issues/110#issuecomment-428804387)
>

The problem of running migrations _without_ transactions is: if a migration fails halfway through, the whole database becomes unusable @koechkevin. It may be acceptable in DEV but certainly not in PROD.

apps/charterafrica/migrations/20240111_075951_update_blocks_rename_collection_to_items.ts
```diff
@@ -0,0 +1,36 @@
+import { MigrateUpArgs, MigrateDownArgs } from "@payloadcms/db-mongodb";
```
kilemensi, line 1: Isn't this the old migrations? It should run before the v1_v2 migration, right?

Dockerfile.dev.charterafrica
```diff
@@ -0,0 +1,108 @@
+FROM node:18-alpine as node-alpine
```
kilemensi, line 1: Why do we need a separate Dockerfile for DEV?

apps/charterafrica/src/lib/data/common/processPageIndex.js
```diff
@@ -48,8 +48,8 @@ const processSpotlight = (page, api, context) => {
         item: {
           ...itemData,
           image: {
-            src: itemData.image.url,
-            alt: itemData.image.alt,
+            src: itemData.image?.url ?? null,
```
kilemensi, line 51: Isn't image **required** on a `Spotlight` component?

charterafrica-dev.yml
```diff
@@ -0,0 +1,19 @@
+version: "3.8"
```
kilemensi, line 1: Any reason this file is not named `docker-compose`?

charterafrica-dev.yml
```diff
@@ -0,0 +1,19 @@
...
+services:
+  charterafrica:
```
kilemensi, line 3: charterafrica should require mongodb service. I don't see any mongodb service.

charterafrica-dev.yml
koechkevin: @kilemensi A similar file would maybe be needed when migrating codeforafrica

charterafrica-dev.yml
koechkevin: Since we are going to have database as a separate service, I thought we'd only provide it as a URL

charterafrica-dev.yml
kilemensi: Then we'll have docker-compose.codeforafrica.yml, docker-compose.charterafrica.yml, etc. i.e. Similar to how we've named Dockerfiles.

PS: Since it's all about services, we can have a single docker-compose file with all services and a dev can choose which app they want to run e.g. `docker compose up codeforafrica`, `docker compose up charterafrica`, etc.

cc @thepsalmist if what I'm saying makes sense

charterafrica-dev.yml
kilemensi: No, docker compose needs to connect to local mongo. It won't connect to remove (PROD or STAGING) databases.

charterafrica-dev.yml
thepsalmist: Yes makes sense to me @kilemensi a single docker-compose file, since the only dependency is MongoDB, the two apps can both depend on that. So we'll have two separate DBs within the clusture, and pass the app specific URL e.g `mongodb://127.0.0.1:27017/charterafrica` and `mongodb://127.0.0.1:27017/codeforafrica`

apps/charterafrica/migrations/20240111_075951_update_blocks_rename_collection_to_items.ts
koechkevin: Recreated payload migrations to ensure they run after old migrations

apps/charterafrica/package.json
```diff
@@ -95,6 +98,7 @@
     "babel-jest": "^29.7.0",
     "eslint": "^8.55.0",
     "eslint-config-commons-ui": "workspace:*",
+    "eslint-import-resolver-typescript": "^3.6.1",
```
kelvinkipruto, line 101: @koechkevin Not sure if this is necessary.

apps/charterafrica/package.json
```diff
@@ -51,9 +51,12 @@
     "@nivo/pie": "^0.84.0",
     "@nivo/tooltip": "^0.84.0",
     "@nivo/waffle": "^0.84.0",
+    "@payloadcms/bundler-webpack": "^1.0.5",
+    "@payloadcms/db-mongodb": "^1.2.0",
```
kelvinkipruto, line 55: I dont think it's a showstopper, but we could maybe use the latest `@payloadcms/db-mongodb@1.32` package. The latest bug fixes have some issues affecting transactions

apps/charterafrica/payload.config.ts
```diff
@@ -178,4 +188,4 @@ export default buildConfig({
         docs.reduce((url, doc) => `${url}/${doc.slug}`, ""),
     }),
   ],
-});
+} as any);
```
kelvinkipruto, line 191: I don't think it is necessary to cast this as `any`

.gitignore
```diff
@@ -51,4 +51,6 @@ apps/promisetracker/public/data/**
 !apps/promisetracker/public/data/README.md
 
 # Storybook
-storybook-static
\ No newline at end of file
+storybook-static
+
+mongo-keyfile
```
kilemensi, line 56:
```suggestion
mongo-keyfile

```

Is this file stored somewhere in the vault?

Dockerfile.charterafrica
```diff
@@ -59,6 +59,7 @@ ARG NEXT_TELEMETRY_DISABLED=1 \
     PAYLOAD_PUBLIC_APP_URL=http://localhost:3000 \
     PAYLOAD_SECRET_KEY
 
+RUN cd apps/charterafrica && npx payload migrate && cd ../../
```
kilemensi, line 62: 🤔 ... Should running migration be part of docker build?

apps/charterafrica/payload.config.ts
```diff
@@ -73,6 +76,11 @@ const adapter = s3Adapter({
 
 export default buildConfig({
   serverURL: appURL,
+  editor: slateEditor({}),
+  db: mongooseAdapter({
+    url: process.env.MONGO_URL,
+    migrationDir: "./migrations",
```
kilemensi, line 82:
```suggestion
    migrationDir: process.env.MIGRATIONS_DIR,
```

where `MIGRATIONS_DIR` defaults to `"./migrations"`.

scripts/charterafrica.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+docker-compose --env-file apps/charterafrica/.env.local up charterafrica --build -d
+
+sleep 5
+
+docker exec mongodb_container /scripts/db-init.sh
+
```
kilemensi, line 8: 1. Do we need this script? May be a Makefile with various commands should work?
2. Do we need to initialise the database **_every time_** we start the app?

docker-compose.yml
```diff
@@ -0,0 +1,59 @@
...
+      - 3000:3000
+    container_name: charterafrica
+    env_file:
+      - apps/charterafrica/.env.local
+    networks:
+      - net
+
+  mongodb_container:
+    container_name: mongodb_container
+    image: mongo:latest
```
kilemensi, line 29:
```suggestion
    image: mongo:6.0
```
Or `mongo:6.0.13` so that we all use the exact same version

Dockerfile.charterafrica
koechkevin: @kilemensi should this be independent of the build? e.g in the CI pipeline?. But I think they should be run before next-build.

kilemensi: Running migrations should not be part of the build process at all @koechkevin. Imagine you're building your image and migrations run but the next step after migration fails. You now can't deploy the app but at the same time, the schema has already been updated so the current running up is likely to fail.

... running migrations should be part of deployment strategy i.e. you're 100% sure you have deployable image.

docker-compose.yml
```diff
@@ -0,0 +1,59 @@
...
+services:
+  charterafrica:
+    depends_on:
+      mongodb_container:
+        condition: service_healthy
+    build:
+      context: .
+      dockerfile: Dockerfile.charterafrica
+      args:
+        MONGO_URL: mongodb://root:rootpassword@host.docker.internal:27017/charterafrica?authSource=admin
```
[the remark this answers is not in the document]
koechkevin: @thepsalmist for some reason during build, the URL by container name is inaccessible causing builds to fail.

.gitignore
koechkevin: [Yes](https://vault.codeforafrica.org/#/vault?search=mong&itemId=7d00fcf3-bc1e-469d-899a-318eea2f3600)

thepsalmist (approved):

apps/charterafrica/payload.config.ts
```diff
@@ -73,6 +76,11 @@ const adapter = s3Adapter({
 
 export default buildConfig({
   serverURL: appURL,
+  editor: slateEditor({}),
+  db: mongooseAdapter({
+    url: process.env.MONGO_URL,
+    migrationDir: process.env.MIGRATIONS_DIR ?? "./migrations",
```
kilemensi, line 82: Set this in the `apps/charterafrica/.env` since `.env` is available everywhere.

docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+        SENTRY_PROJECT: ${SENTRY_PROJECT}
+    ports:
+      - 3000:3000
+    container_name: charterafrica
+    env_file:
+      - apps/charterafrica/.env.local
+    networks:
+      - net
+
+  mongodb_container:
```
kilemensi, line 27: `mongodb_container` is a mouthful... why not just `mongodb` or `mongo` or just `db`?

docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+        SENTRY_AUTH_TOKEN: ${SENTRY_AUTH_TOKEN}
+        SENTRY_ORG: ${SENTRY_ORG}
+        NEXT_PUBLIC_SENTRY_DSN: ${CHARTERAFRICA_SENTRY_DSN}
+        SENTRY_ENV: ${SENTRY_ENV}
+        SENTRY_PROJECT: ${SENTRY_PROJECT}
+    ports:
+      - 3000:3000
+    container_name: charterafrica
+    env_file:
+      - apps/charterafrica/.env.local
```
kilemensi, line 23: Won't the app run without this? Otherwise, we need to update the documentation that there must be a `.env.local` file during development.

data_container/scripts/db-init.js
```diff
@@ -0,0 +1,19 @@
+const config = {
```
kilemensi, line 1: Move these helper files into `contrib/docker/` folder.

docker-compose.yml
koechkevin: Yes, the app won't build without env vars such as PAYLOAD_SECRET_KEY and SENTRY variables

apps/charterafrica/src/payload/globals/Helpdesk.js
```diff
@@ -42,7 +43,7 @@ const Helpdesk = {
           admin: {
             elements: ["ol", "ul", "link"],
           },
```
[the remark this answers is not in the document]
koechkevin: I handled this [here](https://github.com/CodeForAfrica/ui/blob/90bef53e1062b1d016e07f8b64ff036cd2f17dff/apps/charterafrica/src/payload/fields/richText.js#L48)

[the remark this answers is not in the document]
kilemensi: 🤔 ... cool hack but unless there is a **significant** number of places we need to change, I'd rather we move `admin` into `editor` everywhere.

Otherwise, we'd need to maintain two versions of code moving forward.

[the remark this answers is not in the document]
koechkevin: 🤔 we also do not need to be importing slate editor everywhere richText function is imported.
Perhaps maybe rename  admin to editor would be ideal.

docker-compose.yml
```diff
@@ -0,0 +1,60 @@
...
+      - net
+
+  mongodb:
+    container_name: mongodb
+    image: mongo:6.0.13
+    environment:
+      MONGO_INITDB_ROOT_USERNAME: root
+      MONGO_INITDB_ROOT_PASSWORD: rootpassword
+    ports:
+      - 27017:27017
```
kelvinkipruto, line 34: Not sure about this, but shouldn't they be in environment variables?

kelvinkipruto:
@koechkevin I think in the PR description here:

> ⚠️To initialize the db,
> 
> Run make init-db

it should be `make mongodb`

docker-compose.yml
```diff
@@ -0,0 +1,60 @@
...
+      retries: 3
+    command:
+      [
+        "mongod",
+        "--replSet",
+        "dbrs",
+        "--bind_ip_all",
+        "--keyFile",
+        "/etc/mongo-keyfile",
+      ]
```
kelvinkipruto, line 53:
```suggestion
    command: ["sh", "-c", "chmod 600 /etc/mongo-keyfile && exec mongod --replSet dbrs --bind_ip_all --keyFile /etc/mongo-keyfile"]
```
Mongo wont start if the keyfile has the wrong permission. From [here](https://stackoverflow.com/a/14791571)

kelvinkipruto: @koechkevin I see that you have updated it in the description, but I think the `chmod` command should be run inside the container.

koechkevin: Updated the description to capture this @kelvinkipruto

docker-compose.yml
```diff
@@ -0,0 +1,62 @@
...
+    image: mongo:6.0.13
+    environment:
+      MONGO_INITDB_ROOT_USERNAME: ${MONGO_INITDB_ROOT_USERNAME:-root}
+      MONGO_INITDB_ROOT_PASSWORD: ${MONGO_INITDB_ROOT_PASSWORD:-rootpassword}
+    ports:
+      - ${MONGODB_PORT:-27017}:27017
+    env_file:
+      - apps/charterafrica/.env.local
+    volumes:
+      - ./contrib/docker/db:/data/db
```
kilemensi, line 38: Why? contrib folder isn't for data. It's for additional configs e.g. contrib/docker if for additional docker/docker compose configs, etc.

kilemensi:
PS: @koechkevin / @kelvinkipruto , instead of PR description (or in addition to), lets add the documentation to the apps/charterafrica/README.md.

contrib/docker/scripts/dbInit.js
```diff
@@ -0,0 +1,19 @@
...
+      priority: 1,
+    },
+  ],
+};
+rs.initiate(config, { force: true });
+
+rs.status();
+db.createUser({
+  user: "root",
+  pwd: "rootpassword",
```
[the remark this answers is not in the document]
kilemensi: @koechkevin

contrib/docker/scripts/dbInit.js
```diff
@@ -0,0 +1,19 @@
+const config = {
```
kilemensi, line 1: Is this file still being used now that we're creating the cluster in healthcheck?

.gitignore
```diff
@@ -51,4 +51,7 @@ apps/promisetracker/public/data/**
...
+db_data
```
kilemensi, line 57:
```suggestion
```
`db_data` is a volume and not a file. We shouldn't need this.

apps/charterafrica/README.md
```diff
@@ -16,7 +16,22 @@ and modify the `.env.local` file according to your needs.
...
+For more, see NextJS env var docs [here](https://nextjs.org/docs/basic-features/environment-variables).
+
+## Database setup
+
+Generate a new MongoDB keyfile by running the following command:
+
+```bash
+openssl rand -base64 741 > ./mongo-keyfile
+chmod 600 ./mongo-keyfile
+```
```
kilemensi, line 28: Can't this be included in the Makefile? May be under `mongodb-keyfle`

Makefile
```diff
@@ -0,0 +1,13 @@
...
+
+charterafrica:
+	docker compose --env-file apps/charterafrica/.env.local up charterafrica --build -d
+
+mongodb:
+	docker compose --env-file apps/charterafrica/.env.local up --wait mongodb
+
+mongodb-keyfile:
+	openssl rand -base64 741 > ./mongo-keyfile
+	chmod 600 ./mongo-keyfile
```
kilemensi, line 13:
```suggestion
	chmod 600 ./mongo-keyfile

```
New line missing

apps/charterafrica/README.md
kilemensi: 1. If we have moved this to Makefile, shouldn't this doc be updated?
2. Shouldn't this command be run in the project root directory rather than app directory?

koechkevin:
@kilemensi Needs you approval to merge this?

kilemensi:
Not really @koechkevin but the minute it's merged, CI will kick deployment to dev no? Is everything ready for that?

koechkevin:
> Not really @koechkevin but the minute it's merged, CI will kick deployment to dev no? Is everything ready for that?

@kilemensi Yes.
Everything is in place. We'll run migrations once the deployment succeeds.

kilemensi:
Then over to you @koechkevin

PR #673, opened 2024-02-09
Create Pesayetu dockerfile

kelvinkipruto:
## Description

This PR adds a `dockerfile` to deploy Pesayetu.

## Type of change

## Checklist:


kilemensi (approved):
🚀 

----

Add a story to limit the number of pages [built](https://github.com/CodeForAfrica/ui/blob/f16ec58440651b105999bf0de9753c07136ea655/apps/pesayetu/src/pages/explore/%5B%5B...slug%5D%5D.js#L50-L52) at build-time. Creating over 100 pages every time we build a new image is a waste of resources ... May be we should limit to building the main country page at build-time i.e. `/explore/ke`. The rest can be created the first-time they're accessed.

koechkevin (commented):
@kelvinkipruto Where will this be deployed to? is it not dokku? if yes, why not add a workflow .

koechkevin (commented):
@kelvinkipruto we can add
```const next = require("next");
const express = require("express");

const dev = process.env.NODE_ENV !== 'production';
const hostname = process.env.NEXT_HOSTNAME || "localhost";
const port = Number.parseInt(process.env.PORT || "3000", 10);

const app = next({ dev, hostname, port });
const nextHandler = app.getRequestHandler();

async function start() {
  await app.prepare();
  const server = express();
  server.get("*", (req, res) => nextHandler(req, res));
    server.post("*", (req, res) => nextHandler(req, res));
    server.put("*", (req, res) => nextHandler(req, res));
  const port = process.env.PORT || 3000;
  server.listen(port, async () => {
    console.log(`> Ready on http://localhost:${port}`);
  });
}

start();
```

kilemensi:
I'm out of the @koechkevin, why do we need to add that?

koechkevin:
> I'm out of the @koechkevin, why do we need to add that?

@kilemensi if we are to avoid using `pnpm start` or `next start`  serving production static files then we can use the [custom server](https://nextjs.org/docs/pages/building-your-application/configuring/custom-server) to start.

kilemensi:
May be @koechkevin ... my main question is why or where does `pnpm` get the custom server to run but normal `npm` can not?

koechkevin:
> May be @koechkevin ... my main question is why or where does `pnpm` get the custom server to run but normal `npm` can not?

@kilemensi `npm start` and `pnpm start`  work the same way here. Perhaps your proposal was to use `npm` instead of npm?

kilemensi:
> > May be @koechkevin ... my main question is why or where does `pnpm` get the custom server to run but normal `npm` can not?
> 
> @kilemensi `npm start` and `pnpm start` work the same way here. Perhaps your proposal was to use `npm` instead of npm?

Sorry, no, I meant `node` @koechkevin ... why can't `node` find whatever `pnpm`/`npm` uses to start the app?

koechkevin:
> > > May be @koechkevin ... my main question is why or where does `pnpm` get the custom server to run but normal `npm` can not?
> > 
> > 
> > @kilemensi `npm start` and `pnpm start` work the same way here. Perhaps your proposal was to use `npm` instead of npm?
> 
> Sorry, no, I meant `node` @koechkevin ... why can't `node` find whatever `pnpm`/`npm` uses to start the app?

I believe this is the [script](https://github.com/vercel/next.js/blob/v14.2.3/packages/next/src/cli/next-start.ts) executed by next start.

kilemensi:
Please test the changes I've pushed @koechkevin

kelvinkipruto:
@kilemensi Just tested this and it works like a charm. Thank you. @koechkevin Your review before we merge it.

koechkevin (approved):

PR #686, opened 2024-04-29
Replace Keyfile with string credentials

koechkevin:
## Description

At the moment, google configs are a path to keyfile which may be set to a path not inside a docker container.

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/vpnmanager-deploy-dev.yml
+++ b/.github/workflows/vpnmanager-deploy-dev.yml
@@ -2,7 +2,7 @@ name: VPN Manager | Deploy | DEV
 
 on:
   push:
-    branches: [main]
+    branches: [feature/fix-cron-failing]
     paths:
       - "apps/vpnmanager/**"
       - "Dockerfile.vpnmanager"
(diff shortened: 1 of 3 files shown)


koechkevin:
> What inconsistencies @koechkevin ? I'm assuming a volume/mount is set when deploying.

@kilemensi Yes, it works as well but I think having key value in .env is better than file path.

kilemensi:
Key-Values of what @koechkevin? The whole contents of credentials.json?

koechkevin:
> Key-Values of what @koechkevin? The whole contents of credentials.json?

Yes

kilemensi:
Not sure we're talking about the same thing @koechkevin. To confirm, you want to add something like 👇🏽 directly to an env var?
```json
{
  "web": {
    "client_id": "YOUR_CLIENT_ID",
    "project_id": "YOUR_PROJECT_ID",
    "auth_uri": "https://accounts.google.com/o/oauth2/auth",
    "token_uri": "https://www.googleapis.com/oauth2/v3/token",
    "auth_provider_x509_cert_url": "https://www.googleapis.com/oauth2/v1/certs",
    "client_secret": "YOUR_CLIENT_SECRET"
  }
}
```

PR #688, opened 2024-05-03
Update email template

koechkevin:
## Description

Updates vpn manager email template as recommended.

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/apps/vpnmanager/sentry.server.config.ts
+++ b/apps/vpnmanager/sentry.server.config.ts
@@ -1,10 +1,10 @@
 import * as Sentry from "@sentry/nextjs";
 
-const options =  {
+const options = {
   dsn: process.env.SENTRY_DSN,
   environment: process.env.SENTRY_ENV ?? "local",
   tracesSampleRate: 1,
-}
+};
 
 Sentry.init(options);
 
--- a/apps/vpnmanager/sentry.server.config.ts
+++ b/apps/vpnmanager/sentry.server.config.ts
@@ -1,5 +1,13 @@
 import * as Sentry from "@sentry/nextjs";
 
+const options =  {
+  dsn: process.env.SENTRY_DSN,
+  environment: process.env.SENTRY_ENV ?? "local",
+  tracesSampleRate: 1,
+}
+
+Sentry.init(options);
+
 export function initSentry() {
-  Sentry.init();
+  Sentry.init(options);
 }
--- a/apps/vpnmanager/sentry.server.config.ts
+++ b/apps/vpnmanager/sentry.server.config.ts
@@ -1,11 +1,5 @@
 import * as Sentry from "@sentry/nextjs";
 
 export function initSentry() {
-  Sentry.init({
-    dsn: process.env.SENTRY_DSN,
-    environment: process.env.SENTRY_ENV ?? "local",
-    tracesSampleRate: 1,
-  });
+  Sentry.init();
 }
-
-initSentry();
--- a/apps/vpnmanager/sentry.server.config.ts
+++ b/apps/vpnmanager/sentry.server.config.ts
@@ -1,8 +1,11 @@
 import * as Sentry from "@sentry/nextjs";
 
-Sentry.init({
-  dsn: process.env.SENTRY_DSN,
-  environment: process.env.SENTRY_ENV,
-  tracesSampleRate: 1,
-  debug: false,
-});
+export function initSentry() {
+  Sentry.init({
+    dsn: process.env.SENTRY_DSN,
+    environment: process.env.SENTRY_ENV ?? "local",
+    tracesSampleRate: 1,
+  });
+}
+
+initSentry();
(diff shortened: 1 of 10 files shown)


apps/vpnmanager/src/lib/email/sender.ts
```diff
@@ -25,9 +30,13 @@ export async function sendVpnKeyEmail({ recipient: to, key }: MailSender) {
       to,
       from,
       subject,
-      html: emailKeyTemplate(key),
+      html: emailKeyTemplate(key, name),
     };
     await sgMail.send(message);
+    Sentry.captureMessage(
+      `Email sent to "${to}" ${emailKeyTemplate("*hidden*", name)}`,
```
kilemensi, line 37:
```suggestion
      `Outline key sent`
```

We can then use [local scope](https://docs.sentry.io/platforms/javascript/enriching-events/scopes/#local-scopes) to set [user](https://docs.sentry.io/platforms/javascript/enriching-events/identify-user/) using `${to}` and [attachment](https://docs.sentry.io/platforms/javascript/enriching-events/attachments/#uploading-attachments) using `${emailKeyTemplate("*hidden*", name)}`.

apps/vpnmanager/src/lib/email/templates.ts
```diff
@@ -1,14 +1,23 @@
...
+      <li>MacOS: <a href="https://itunes.apple.com/app/outline-app/id1356178125">Download from App Store</a></li>
+      <li>Windows: <a href="https://s3.amazonaws.com/outline-releases/client/windows/stable/Outline-Client.exe">Download</a></li>
+      <li>Linux: <a href="https://s3.amazonaws.com/outline-releases/client/linux/stable/Outline-Client.AppImage">Download</a></li>
+      <li>Android: <a href="https://play.google.com/store/apps/details?id=org.outline.android.client">Download from Google Play</a></li>
+      <li>Android alternative link: <a href="https://s3.amazonaws.com/outline-releases/client/android/stable/Outline-Client.apk">Download APK</a></li>
+    </ul>
+    <li>Copy your Access key ${key}</li>
+    <li>Open the Outline client app. If your access key is auto-detected, tap 'Connect'. If your access key is not auto-detected, paste it in the field, then tap 'Connect'. You are good to go!</li>
+  </ol>
+
```
kilemensi, line 21: May be add a note that the same key can be used in multiple devices?

apps/vpnmanager/sentry.server.config.ts
```diff
@@ -1,8 +1,11 @@
...
-});
+export function initSentry() {
+  Sentry.init({
+    dsn: process.env.SENTRY_DSN,
+    environment: process.env.SENTRY_ENV ?? "local",
+    tracesSampleRate: 1,
+  });
+}
+
+initSentry();
```
kilemensi, line 11: Why not leave `Sentry.init` as is and define `initSentry` as:

```suggestion
export function initSentry() {
  Sentry.init();
}
```
?

kilemensi:
Your reviews @VinneyJ / @kelvinkipruto

kelvinkipruto (approved):
LGTM

apps/vpnmanager/sentry.server.config.ts
```diff
@@ -1,8 +1,13 @@
...
-  environment: process.env.SENTRY_ENV,
+  environment: process.env.SENTRY_ENV ?? "local",
   tracesSampleRate: 1,
-  debug: false,
-});
+};
+
+Sentry.init(options);
+
+export function initSentry() {
```
VinneyJ, line 11:
```
export function initSentry() {
 try {
    Sentry.init(options);
} catch (error) {
    console.error("Failed to initialize Sentry", error);
}
}
``` 
just so to catch any errors that might occur during sentry setup.

VinneyJ (changes requested):

apps/vpnmanager/sentry.server.config.ts
koechkevin: @VinneyJ What would the relevance of this be apart from logging the error without notifying us?

I'd rather have it as is. If a critical error happens at initialization, let it terminate since it is necessary for its execution.

apps/vpnmanager/src/lib/email/sender.ts
```diff
@@ -25,9 +26,21 @@ export async function sendVpnKeyEmail({ recipient: to, key }: MailSender) {
...
       from,
       subject,
-      html: emailKeyTemplate(key),
+      html: emailKeyTemplate(key, name),
     };
     await sgMail.send(message);
+    Sentry.withScope((scope) => {
+      scope.setLevel("info");
+      scope.setUser({
+        id: to,
```
kilemensi, line 35:
```suggestion
```

I think if you have `email`, `id` is optional.

kilemensi (approved):

apps/vpnmanager/sentry.server.config.ts
VinneyJ: Yes @koechkevin makes sense. Was just looking at it as an alternative notification for the team when Sentry fails to initialize in production for some reason, ensuring we're promptly informed for example: 
```
    } catch (error) {
        console.error("Failed to initialize Sentry", error);
        notifySlack
```
However, the code looks good to me.

VinneyJ (approved):

PR #718, opened 2024-06-18
fix: destroy test review app

VinneyJ:
## Description

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/codeforafrica-deploy-review-app.yml
+++ b/.github/workflows/codeforafrica-deploy-review-app.yml
@@ -86,7 +86,7 @@ jobs:
             |-------|--------|
             | ${{ env.APP_NAME }}  | [Visit](https://${{ env.APP_NAME }}.dev.codeforafrica.org)  |
 
-  destroy_review_app4:
+  destroy_review_app:
     runs-on: ubuntu-latest
     if: github.event_name == 'pull_request' && github.event.action == 'closed'
     steps:
--- a/.github/workflows/codeforafrica-deploy-review-app.yml
+++ b/.github/workflows/codeforafrica-deploy-review-app.yml
@@ -86,17 +86,14 @@ jobs:
             |-------|--------|
             | ${{ env.APP_NAME }}  | [Visit](https://${{ env.APP_NAME }}.dev.codeforafrica.org)  |
 
-  destroy_review_app:
+  destroy_review_app4:
     runs-on: ubuntu-latest
-    # only run when a pull request is closed
-    if: github.event.action == 'closed'
+    if: github.event_name == 'pull_request' && github.event.action == 'closed'
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/codeforafrica-deploy-review-app.yml
```diff
@@ -88,15 +88,12 @@ jobs:
 
   destroy_review_app:
     runs-on: ubuntu-latest
-    # only run when a pull request is closed
-    if: github.event.action == 'closed'
+    if: github.event_name == 'pull_request' && github.event.action == 'closed'
     steps:
-      - name: Destroy the review app
+      - name: Destroy review app
```
kilemensi, line 93:
```suggestion
      - name: Destroy the review app
```
Don't see the need for this change

.github/workflows/codeforafrica-deploy-review-app.yml
```diff
@@ -88,15 +88,12 @@ jobs:
...
         uses: dokku/github-action@master
         with:
-          # destroy a review app
           command: review-apps:destroy
-          git_remote_url: ${{ env.DOKKU_REMOTE_URL }}/${{ env.APP_NAME }}
-          # specify a name for the review app
+          git_remote_url: ${{ env.DOKKU_REMOTE_URL }}/codeforafrica-ui
```
kilemensi, line 97: Have we tested this and it actually deletes the _review_ app (e.g. `codeforafrica-ui-789`) instead of the _dev_ app (`codeforafrica-ui`)?

kilemensi: ✅ ... saw that both URL and Name are [set](https://github.com/dokku/github-action/blob/master/example-workflows/review-app.yml#L40-L42).

kilemensi (approved):

.github/workflows/codeforafrica-deploy-review-app.yml
VinneyJ: I have not tested this yet because running the workflow requires executing it in the main branch. Do you have any suggestions on how we can safely test this without affecting the main environment, or should we proceed with a cautious test in the main branch?

kilemensi: There are tools @VinneyJ e.g. [act](https://github.com/nektos/act) but I'm not 100% convinced they're worth it... Code looks legit so I'm happy for us to proceed with deployment.

.github/workflows/codeforafrica-deploy-review-app.yml
VinneyJ: I will use this for testing purposes; removing the change.

koechkevin (approved):

PR #723, opened 2024-06-19
Fix Dokku Destoy on PR Close

kelvinkipruto:
## Description

This PR fixes a bug where Dokku did not destroy a preview app on PR close

## Type of change

- [x] Bug Fix

## Checklist:

### DIFF
--- a/.github/workflows/codeforafrica-deploy-review-app.yml
+++ b/.github/workflows/codeforafrica-deploy-review-app.yml
@@ -89,7 +89,7 @@ jobs:
 
   destroy_review_app:
     runs-on: ubuntu-latest
-    if: github.event_name == 'pull_request' && github.event.action == 'closed'
+    if: github.event_name == 'pull_request' && (github.event.action == 'closed' || github.event.action == 'merged')
     steps:
       - name: Destroy review app
         uses: dokku/github-action@master
--- a/.github/workflows/codeforafrica-deploy-review-app.yml
+++ b/.github/workflows/codeforafrica-deploy-review-app.yml
@@ -6,7 +6,8 @@ on:
       - "apps/codeforafrica/**"
       - "Dockerfile.codeforafrica"
       - ".github/workflows/codeforafrica-deploy-review-app.yml"
-    types: [opened, reopened, edited, closed, synchronize]
+      # By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened. We need to add closed to the list of activity types to ensure that the workflow runs when a pull request is closed.
+    types: [opened, reopened, closed, synchronize]
 concurrency:
   group: "${{ github.workflow }} @ ${{ github.ref }}"
   cancel-in-progress: false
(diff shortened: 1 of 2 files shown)


VinneyJ (approved):

.github/workflows/codeforafrica-deploy-review-app.yml
```diff
@@ -6,6 +6,7 @@ on:
       - "apps/codeforafrica/**"
       - "Dockerfile.codeforafrica"
       - ".github/workflows/codeforafrica-deploy-review.yml"
+    types: [opened, reopened, edited, closed, synchronize]
```
kilemensi, line 9: For the slow ones such as myself, what does this add? Or put another way, what are the default `types`?

kelvinkipruto: @kilemensi From the [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request)

> By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened. To trigger workflows by different activity types, use the types keyword.

.github/workflows/codeforafrica-deploy-review-app.yml
```diff
@@ -6,6 +6,7 @@ on:
       - "apps/codeforafrica/**"
       - "Dockerfile.codeforafrica"
       - ".github/workflows/codeforafrica-deploy-review.yml"
```
koechkevin, line 8:
```suggestion
      - ".github/workflows/codeforafrica-deploy-review-app.yml"
```

.github/workflows/codeforafrica-deploy-review-app.yml
kilemensi: Nice... can we add this as a comment above the types line? i.e. We need `closed` `edited` in addition to the default events; `opened`, `synchronize`, ad `reopened`.

What does edited do?

kelvinkipruto: Good catch @kilemensi. Edited runs when a pull request's description or labels are edited. I've removed it.

koechkevin:
@kelvinkipruto we might then need to update the condition for on merge a pr since we have only set for when it is closed.

Check line 92

.github/workflows/codeforafrica-deploy-review-app.yml
```diff
@@ -5,7 +5,9 @@ on:
     paths:
       - "apps/codeforafrica/**"
       - "Dockerfile.codeforafrica"
-      - ".github/workflows/codeforafrica-deploy-review.yml"
+      - ".github/workflows/codeforafrica-deploy-review-app.yml"
+      # By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened. We need to add closed to the list of activity types to ensure that the workflow runs when a pull request is closed.
```
kilemensi, line 9: Why such a long line? Limit to 80 - 100 chars (whatever makes sense).

kilemensi (approved):

kelvinkipruto:
@koechkevin Good catch Let me update it.

kilemensi:
> we might then need to update the condition for on merge a pr since we have only set for when it is closed
>

Again, for the rest of us still learning the ropes, when a PR is merged, isn't it also [`closed`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges) @koechkevin / @kelvinkipruto ?

koechkevin (approved):

PR #725, opened 2024-06-20
Single Multi-Stage Dockerfile for the repo

kilemensi:
## Description

We are currently using one `Dockerfile` per app. This means there is a lot of duplication/little reuse as well as it's quite easy to end up with each app using different base images/versions.

This PR introduces a single multi-stage `Dockerfile` that can be used to build all apps by setting build [`target`](https://docs.docker.com/build/building/multi-stage/#name-your-build-stages). To kick-off the process, this PR migrates **PesaYetu** and **VPN Manager** apps first.

In the process, the following were observed:

  - [x] New applications do not reuse existing configs and conventions: Both **PesaYetu** and **VPN Manager** did not have working `eslint`/`lintstaged` configurations despite these configs existing in the monorepo. While this PR has fixed these issues, we'd need to rethink our configurations now that we've added TypeScript into the mix
  - [ ] Preference to ignore lint/type warnings and errors: Again, we should move away from this behaviour. What's the point of adding TypeScript if we ignore type warnings? What's the point of using `eslint` to check for potential errors if we ignore them?
  
## Type of change

- [x] Chore

## Screenshots

N/A

## Checklist:


kilemensi:
Damn, no one wants to take the first crack at this @CodeForAfrica/tech ?

koechkevin:
`pnpm fetch` taking time🙆‍♀️

kilemensi:
> `pnpm fetch` taking time🙆‍♀️

Is your Internet protesting too @koechkevin?

koechkevin:
> > `pnpm fetch` taking time🙆‍♀️
> 
> Is your Internet protesting too @koechkevin?

at 18 Mbps download speed

koechkevin:
```
[+] Building 2859.2s (34/34) FINISHED 
```
😁

koechkevin:
LGTM

koechkevin (approved):

VinneyJ (approved):
LGTM

kelvinkipruto (approved):
Looks good @kilemensi. My only concern is that the `dockerfile` could get huge over time as we add new applications.

kilemensi:
>Looks good @kilemensi. My only concern is that the dockerfile could get huge over time as we add new applications.
>

It may @kelvinkipruto ... but if each new app added takes say `50` lines, we can easily have more than `20` apps and the whole `Dockerfile` will still be around `1000` lines. That's still manageable I believe given the fact we'll 100% be sure all apps are using same versions, built the same way, etc.

PR #726, opened 2024-06-20
Display all Tools

koechkevin:
## Description

There's a need to display all the tools even if external ID is missing.

Fixes # (issue)

## Type of change

## Screenshots
image

image

## Checklist:


apps/charterafrica/src/components/Tool/DescriptionAndShare.js
```diff
@@ -18,18 +18,20 @@ const DescriptionAndShare = React.forwardRef(
...
+            textAlign="left"
+            color="neutral.dark"
+            sx={{
+              mt: 3.75,
+              textTransform: "none",
+              textAlign: { xs: "center", sm: "left" },
+            }}
+            variant="captionCap"
+          >
+            {activeText} {lastActive}
```
kilemensi, line 32: We can create `lastActiveText` before `return` above
```js
  const lastActiveText = lastActive ? `${activeText} ${lastActive}` : null
```
Then, we don't need `{lastActive ? (` around the `RichTypography` since it can handle `null`/empty:
```suggestion
            {lastActiveText}
```

kilemensi (commented):
👍🏽 

---

Can you may be describe the approach we've taken? e.g. yes, we don't fetch the tools but will they still have associated contributors or organisations if this information is provided in the Airtable?

koechkevin:
> 👍🏽
> 
> Can you may be describe the approach we've taken? e.g. yes, we don't fetch the tools but will they still have associated contributors or organisations if this information is provided in the Airtable?

@kilemensi tools are a source of truth.
We do not process any contributor/organisation without tools. Since we initially did not process any tool without external ID, we only needed to reverse that and ensure we do not try to query from github if external ID does not exist.

kilemensi:
I think I get you @koechkevin but to confirm:

1. Everything starts with tools
2. We currently don't process any tool without externalId (GitHub)

This PR changes this
1. We directly show information about tools without externalId (point 2)

_My immediate questions:_

1. Does this mean we also show contributors/organisations using information directly obtained from Airtable or do we stop at tools only?
2. Do we show all tools without externalIds or do we at least ensure certain minimum information is present before displaying e.g. if I've only include tool name without description, logo, etc., will that show up on the site?

koechkevin:
> I think I get you @koechkevin but to confirm:
> 
> 1. Everything starts with tools
> 2. We currently don't process any tool without externalId (GitHub)
> 
> This PR changes this
> 
> 1. We directly show information about tools without externalId (point 2)
> 
> _My immediate questions:_
> 
> 1. Does this mean we also show contributors/organisations using information directly obtained from Airtable or do we stop at tools only?
> 2. Do we show all tools without externalIds or do we at least ensure certain minimum information is present before displaying e.g. if I've only include tool name without description, logo, etc., will that show up on the site?

1. We stop at tools only.
2. At the moment we ensure only the name is present. We could as well include the image.

kilemensi:
Cool @koechkevin 

1. Lets create an issue for us to consider organisations and contributors.
2. Lets include a bit more, otherwise the page will be quite empty: name, description, country, logo/image and theme? I'd also like to say at least a single partner or supporter but you can discuss that with Sarah/Eric if it's a good idea or no.

kilemensi (approved):

PR #745, opened 2024-06-27
Fix @/roboshield social links

kilemensi:
## Description

Create a reusable `SocialMediaLinkIcon` component that can be used in both `NavBar` and `Footer`

## Type of change

## Screenshots

S

## Checklist:

### DIFF
--- a/apps/roboshield/src/components/SocialMediaLinkIcon/SocialMediaLinkIcon.tsx
+++ b/apps/roboshield/src/components/SocialMediaLinkIcon/SocialMediaLinkIcon.tsx
@@ -1,0 +1,64 @@
+import { Link, RichTypography } from "@commons-ui/next";
+import { Grid, SvgIcon } from "@mui/material";
+import React from "react";
+
+import FacebookIcon from "@/roboshield/assets/icons/Type=facebook, Size=24, Color=CurrentColor.svg";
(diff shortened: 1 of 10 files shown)


apps/roboshield/src/components/SocialMediaLinkIcon/SocialMediaLinkIcon.tsx
```diff
@@ -0,0 +1,64 @@
...
+  | "Github"
+  | "Instagram"
+  | "LinkedIn"
+  | "Slack"
+  | "Twitter";
+
+interface SocialMediaLink {
+  platform: SocialMediaPlatform;
+  // TODO(koech): Confirm why we chose url instead of href in the CMS
+  url: string;
```
koechkevin, line 34: @kilemensi totally fine to use href here.

The only place we weren't supposed to use `href` was collections with afterReadHook mutating href field.

koechkevin (approved):

apps/roboshield/src/components/SocialMediaLinkIcon/SocialMediaLinkIcon.tsx
kilemensi: Cool, lets create an issue to track this... I'm assuming this is how it's done in `codeforafrica` and `charterafrica` apps so we may have to create a migration.

PR #747, opened 2024-06-28
Blockify CMS Blocks

kelvinkipruto:
## Description

This PR introduces a way to `blockify` the CMS blocks. Each block can be used on any page. It also includes strong type inference for each block. 
It also includes a simple `page-header` block for reference.

## Type of change

## Screenshots
image

## Checklist:


apps/roboshield/src/components/BlockRenderer/BlockRenderer.tsx
```diff
@@ -0,0 +1,29 @@
...
+type BlockType = Page["blocks"];
+interface BlockRendererProps {
+  blocks: BlockType;
+}
+
+const components = {
+  "page-header": PageHeader,
+};
+
+type BlockSlug = keyof typeof components;
```
kilemensi, line 13: Isn't `BlockSlug` always the same as `BlockType` (or should be the same)?
```ts
type BlockSlug = BlockType
```

And isn't `BlockRendererProps` just a subset of `Page`?
```ts
interface BlockRendererProps extends Pick<Page, "blocks"> {
}
```

apps/roboshield/payload-types.ts
```diff
@@ -0,0 +1,224 @@
...
+  name?: string | null;
+  batch?: number | null;
+  updatedAt: string;
+  createdAt: string;
+}
+/**
+ * This interface was referenced by `Config`'s JSON-Schema
+ * via the `definition` "settings-site".
+ */
+export interface SettingsSite {
```
kilemensi, line 136: There isn't any way of this coming out as `SiteSettings` or just `Site`?

apps/roboshield/src/pages/about.tsx
```diff
@@ -0,0 +1,26 @@
+import { getPageServerSideProps } from "@/roboshield/lib/data";
+import BlockRenderer from "@/roboshield/components/BlockRenderer";
+import { Page } from "@/root/payload-types";
+
+type BlockType = Page["blocks"];
+
+interface PageProps {
+  blocks: BlockType;
+}
```
kilemensi, line 9: Shouldn't `PageProps` or `ServerSideProps` be defined once in lib/data and reused? I'm assuming all pages will have similar structure.

apps/roboshield/src/pages/about.tsx
```diff
@@ -0,0 +1,26 @@
+import { getPageServerSideProps } from "@/roboshield/lib/data";
```
kilemensi, line 1: Shouldn't this be one of the catch all pages `....tsx` or `[...].tsx`? I'm assuming all pages will just call render blocks

kilemensi (commented):
Too tired to do a proper review but it looks good as a starting point.

koechkevin (commented):
image

kelvinkipruto:
@koechkevin Retry building with the latest changes. I have fixed the issue.

apps/roboshield/payload-types.ts
kelvinkipruto: @kilemensi Yes. There is a [way](https://payloadcms.com/docs/typescript/generating-types#custom-field-interfaces). Let me update it

kelvinkipruto: Looks like it does not work for globals

koechkevin:
> @koechkevin Retry building with the latest changes. I have fixed the issue. It is an issue with the [typescript version](https://github.com/payloadcms/payload/issues/6896)

@kelvinkipruto run `pnpm build-next`

kelvinkipruto:
@kilemensi can you build successfully, or you're facing the same issue @koechkevin has pointed out? On my end builds run successfully.

apps/roboshield/src/pages/about.tsx
kelvinkipruto: @kilemensi Since we only have 2 pages, I think there is no need for a catch-all page. Perhaps we can introduce it when there's an actual need for any extra page

kilemensi: Isn't `pages/index.tsx` going to be exact copy of this then @kelvinkipruto ?

koechkevin (approved):

m453h (approved):
🚀

Looks good to me, I hope we'll later on use Module aliasing in all of imports under the `src/payload/` directory for the sake of consistency but that shouldn't stop us from merging this to main

PR #748, opened 2024-07-02
@/Roboshield Move Hero  to Payload CMS

m453h:
## Description

This PR integrates the Hero Section of the RoboShield application with Payload CMS.

## Type of change

## Screenshots
Screenshot 2024-07-02 at 12 46 55

Screenshot 2024-07-02 at 12 47 46

## Checklist:


koechkevin (approved):

apps/roboshield/src/lib/data/payload.types.ts
```diff
@@ -0,0 +1,120 @@
+interface TextNode {
```
kelvinkipruto, line 1: I believe this should be generated based on the payload config instead of hard coding it.

apps/roboshield/src/payload/globals/Site/HeroTab.ts
```diff
@@ -0,0 +1,65 @@
+import { RowLabelArgs } from "payload/dist/admin/components/forms/RowLabel/types";
```
kelvinkipruto, line 1: I don't think this should be a global.

apps/roboshield/src/lib/data/payload.types.ts
m453h: I have auto generated the types with my latest commit, am I missing something else ?

m453h: @kelvinkipruto I have resorted to using Payload generated types in all sections of the Hero , I think we should be fine !

PR #755, opened 2024-07-04
Bugfix/charterafrica build

koechkevin:
## Description

Tools, Organisations and contributors throws 500 error page on production from the latest build. This happens because when trying to access code labels, locale is undefined. This PR ensures that locale can never be undefined when getting server side props.

This has been tested on [dev server](https://charterafrica.dev.codeforafrica.org/resources/tools).

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/charterafrica-deploy-dev.yml
+++ b/.github/workflows/charterafrica-deploy-dev.yml
@@ -2,7 +2,7 @@ name: charterAFRICA | Deploy | DEV
 
 on:
   push:
-    branches: [main]
+    branches: [bugfix/charterafrica-build]
     paths:
       - "apps/charterafrica/**"
       - "Dockerfile"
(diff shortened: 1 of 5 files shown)


kilemensi:
Why is `locale` `undefined` on this pages @koechkevin ?

koechkevin:
> Why is `locale` `undefined` on this pages @koechkevin ?

It is sometimes undefined for all the pages but it throws an error on these specific pages because of the way we used locale to get the [code-defined](https://github.com/CodeForAfrica/ui/blob/45e70e64b1a25033f6dd58c2052396ebf9474774/apps/charterafrica/src/lib/data/common/processPageTools.js#L182) labels.

Not sure if this bug was introduced by any recent version bumps.

kilemensi:
I think there is a deeper issue here @koechkevin ... lets deploy this as a quick fix but please create an issue for us to investigate this further tomorrow.

kilemensi (approved):

PR #758, opened 2024-07-08
Fixed charterafrica build. Add essential files

koechkevin:
## Description

Since the migration to new Dockerfile, Charterafrica build has not been working as expected.

Essential files need to be bundled e.g scripts, app.json .env etc for features to work.
This PR fixes that

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -312,7 +312,9 @@ COPY --from=codeforafrica-builder --chown=nextjs:nodejs /workspace/node_modules 
 
 # Since we can't use output: "standalone", copy all app's dependencies
 COPY --from=codeforafrica-builder --chown=nextjs:nodejs /workspace/apps/codeforafrica/node_modules ./apps/codeforafrica/node_modules
-
+COPY --from=codeforafrica-builder --chown=nextjs:nodejs /workspace/apps/codeforafrica/next.config.js ./apps/codeforafrica/next.config.js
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -201,7 +201,11 @@ COPY --from=charterafrica-builder --chown=nextjs:nodejs /workspace/node_modules
 
 # Since we can't use output: "standalone", copy all app's dependencies
 COPY --from=charterafrica-builder --chown=nextjs:nodejs /workspace/apps/charterafrica/node_modules ./apps/charterafrica/node_modules
-
+COPY --from=charterafrica-builder --chown=nextjs:nodejs /workspace/apps/charterafrica/next.config.js ./apps/charterafrica/next.config.js
+COPY --from=charterafrica-builder --chown=nextjs:nodejs /workspace/apps/charterafrica/.env ./apps/charterafrica/.env
+COPY --from=charterafrica-builder --chown=nextjs:nodejs /workspace/apps/charterafrica/migrations ./apps/charterafrica/migrations
```
kelvinkipruto, line 206: @koechkevin I still don't understand why we need the migrations. Isn't it only run once and not required after?

koechkevin: @kelvinkipruto existing ones are not needed anymore but what if new migrations are added?
They'll need to be run on the server.

kilemensi (approved):
👍🏽 

--

I'm assuming something like this could be happening with codeforafrica and roboshield since they both use Payload.

PR #782, opened 2024-07-16
Bump charterafrica version to 0.1.30

koechkevin:
## Description

Images under the section "Southern African Development Community Fact Sheets" section are not showing. It is a minor issue where a component is using `image.src` instead of `image.url`.
 
Also, some images have alt text missing so I added an `afterRead` hook to  default to filename.

Fixes # (issue)

## Type of change

## Screenshots
image

## Checklist:

### DIFF
--- a/apps/charterafrica/src/components/Spotlight/Spotlight.js
+++ b/apps/charterafrica/src/components/Spotlight/Spotlight.js
@@ -73,7 +73,7 @@ const Spotlight = React.forwardRef(function Spotlight(props, ref) {
                     <CardMedia
                       alt={item.title}
                       component="img"
-                      src={item.image.src}
+                      src={item.image.url}
                       sx={{
                         height: {
                           xs: "auto",
--- a/apps/charterafrica/src/payload/collections/Media.js
+++ b/apps/charterafrica/src/payload/collections/Media.js
@@ -19,6 +19,14 @@ const Media = {
       required: true,
     },
   ],
+  hooks: {
+    afterRead: [
+      ({ doc }) => ({
+        ...doc,
+        alt: doc.alt || doc.filename,
+      }),
+    ],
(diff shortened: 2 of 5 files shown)


apps/charterafrica/src/payload/collections/Media.js
```diff
@@ -19,6 +19,14 @@ const Media = {
...
     },
   ],
+  hooks: {
+    afterRead: [
+      ({ doc }) => ({
+        ...doc,
+        alt: doc.alt || doc.filename,
+      }),
+    ],
+  },
```
kilemensi, line 29: Makes sense-ish but not too sure about it... isn't it something that should be decided on front-end? e.g. no alt, use org's or tool's name instead, etc.

koechkevin: @kilemensi this approach ensures we do not get `undefined` serializing errors on development.

kilemensi: Yes @koechkevin but that means the front-end won't know if the user actually wanted to use the `alt` or not. `undefined` can't be serialized but `null` can.

apps/charterafrica/src/components/Spotlight/Spotlight.js
```diff
@@ -73,7 +73,7 @@ const Spotlight = React.forwardRef(function Spotlight(props, ref) {
                     <CardMedia
                       alt={item.title}
                       component="img"
-                      src={item.image.src}
+                      src={item.image.url}
```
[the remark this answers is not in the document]
koechkevin: Addressed [here](apps/charterafrica/src/lib/data/common/processPageIndex.js) @kilemensi

kilemensi (approved):

PR #784, opened 2024-07-16
@/RoboShield fix misc issues

m453h:
## Description

This PR fixes several errors in the RoboShield application namely:
1. Warning(s) logged in console due to incorrect props/parameters passed in `TwoToneBackground` and `Footer` components as shown below:
Screenshot 2024-07-16 at 18 42 45

2. Failing `build-payload` and `generate:types` due to missing `PAYLOAD_PUBLIC_APP_URL` variable

## Type of change

## Checklist:


kilemensi:
👍🏽 @m453h ... love the UI fixes.

Can you add a bit of context on the Payload/TypeScript fixes e.g. both `payload.config.ts` and `server.t`s loads `.env` (and `.env.local`) file so they should be able to pick up `PAYLOAD_PUBLIC_APP_URL` env var from there (I'm assuming `payload build` and `payload generate:types` loads one or both of these entry files).

koechkevin (approved):
@m453h You'll need to bump version in `package.json` to trigger production deployment

m453h:
> 👍🏽 @m453h ... love the UI fixes.
> 
> Can you add a bit of context on the Payload/TypeScript fixes e.g. both `payload.config.ts` and `server.t`s loads `.env` (and `.env.local`) file so they should be able to pick up `PAYLOAD_PUBLIC_APP_URL` env var from there (I'm assuming `payload build` and `payload generate:types` loads one or both of these entry files).

Yes they both load the `.env `and `.env.local,` however after digging around I observed that when the `payload.config.ts` loads the file(s) it fails to resolve  `PAYLOAD_PUBLIC_APP_URL`,  as we have it set to `PAYLOAD_PUBLIC_APP_URL=$NEXT_PUBLIC_APP_URL`

Two alternative approaches I thought of are :
1. To explicitly set the default URL value in the .env file
2. To discard `dotnev` and use  `loadEnvConfig` in  `payload.config.ts` to load the `.env` file(s)

kelvinkipruto (approved):

kilemensi:
Nice!!! Yes, `$NEXT_PUBLIC_APP_URL` is a Next.js extensions.

Cool, my suggestions would be:

1. Try `loadEnvConfig` in `payload.config.ts`. If it works, lets use that approach.
2. If it doesn't work, lets explicitly set default value to `PAYLOAD_PUBLIC_APP_URL` but also update docs to reflect that it's now required.

m453h:
> Nice!!! Yes, `$NEXT_PUBLIC_APP_URL` is a Next.js extensions.
> 
> Cool, my suggestions would be:
> 
> 1. Try `loadEnvConfig` in `payload.config.ts`. If it works, lets use that approach.
> 2. If it doesn't work, lets explicitly set default value to `PAYLOAD_PUBLIC_APP_URL` but also update docs to reflect that it's now required.

Cool, option 1. with `loadEnvConfig` works.

apps/roboshield/payload.config.ts
```diff
@@ -122,4 +122,7 @@ export default buildConfig({
     }),
   ] as any[],
   telemetry: process?.env?.NODE_ENV !== "production",
+  typescript: {
+    declare: false,
+  },
```
kilemensi, line 127: What are the benefits of this?

kilemensi (approved):

apps/roboshield/payload.config.ts
m453h: By default its set to true, so when we generate payload types at the end of the generated `payload-types.ts` we will get a module and an exported interface. 
```
declare module 'payload' {
  export interface GeneratedTypes extends Config {}
}
```

However when I do that , I end up getting a name collision on my end. 

Screenshot 2024-07-17 at 10 24 56

kilemensi: Good catch... it's a collision between `GeneratedTypes` from payload and what?

m453h: Hmm....the auto `GeneratedTypes` are colliding with whats in  `node_modules/payload/dist/index.d.ts`

m453h: I'll need to do further investigation on this will merge it, to main as it is

PR #786, opened 2024-07-17
@Hurumap ChartTooltip

### LINKED ISSUE #776: @hurumap/core/ChartTooltip

koechkevin:
## Description

Moves ChartTooltip component to Hurumap core.

NB: couldn't get `sx` or any other Material UI specific feature to work for Tooltip inside `hurumap/core` except style  + useTheme() properties.

Fixes #776

## Type of change

## Checklist:


apps/pesayetu/src/components/HURUmap/Chart/index.js
```diff
@@ -101,6 +101,7 @@ function Chart({
           </ThemeProvider>
         </StyledEngineProvider>,
       );
+      console.log(el);
```
kilemensi, line 104:
```suggestion
```

kilemensi (approved):

kilemensi:
>NB: couldn't get sx or any other Material UI specific feature to work for Tooltip inside hurumap/core except style + useTheme() properties.
>

Didn't see the above comment when I was review @koechkevin. What issues did you run into?

koechkevin:
> > NB: couldn't get sx or any other Material UI specific feature to work for Tooltip inside hurumap/core except style + useTheme() properties.
> 
> Didn't see the above comment when I was review @koechkevin. What issues did you run into?

@kilemensi sx styles just don't get applied when you use inside hurumap-core package.

kilemensi:
Not sure I get you @koechkevin ... but let me pull and run locally.

kilemensi:
I think style work because `style` have higher priority that classes (`sx`). For `sx` to work, we'll need to figure out how to use [custom theme](https://github.com/vega/vega-tooltip/blob/main/docs/customizing_your_tooltip.md).

koechkevin:
I think we'll go with inline style here because if we were to use custom theme as [specified here](https://github.com/vega/vega-tooltip/blob/main/docs/customizing_your_tooltip.md),  we'd have to write custom CSS. Since we may need to use theme properties, it would mean having to write Internal CSS in a JS file which is not ideal.  Style prop is the closest thing in terms of usage to sx in my view.

kilemensi:
> I think we'll go with inline ...
>

You're the closest to the code @koechkevin but **_If_** I understood that doc, no, we wouldn't need to create custom CSS file. `sx` should be all we need.
```jsx
    <ChartOrChartParentComponent
      ...
      sx={{
        "& .vg-tooltip.custom-theme": {
           color: "red",
        }
      }}
      >
```

`& .vg-tooltip.custom-theme` can be `& #vg-tooltip-element.vg-tooltip.custom-theme` if we know/can set id of `Tooltip` element.

But, at the moment, job one is to know, does it work?

koechkevin:
> > I think we'll go with inline ...
> 
> You're the closest to the code @koechkevin but **_If_** I understood that doc, no, we wouldn't need to create custom CSS file. `sx` should be all we need.
> 
> ```js
>     <ChartOrChartParentComponent
>       ...
>       sx={{
>         "& .vg-tooltip.custom-theme": {
>            color: "red",
>         }
>       }}
>       >
> ```
> 
> `& .vg-tooltip.custom-theme` can be `& #vg-tooltip-element.vg-tooltip.custom-theme` if we know/can set id of `Tooltip` element.
> 
> But, at the moment, job one is to know, does it work?

This wouldn't work on the tooltip @kilemensi the parent component of Tooltip is a div which doesn't support sx. Also, its children are  React components which would have to accept sx props independently

kilemensi:
... it doesn't need to be a direct parent @koechkevin . `& .class` would work on any parent; `&>.class`only works on direct parent (if I still remember my CSS correctly)

koechkevin:
@kelvinkipruto @kilemensi I was able to get `sx` to work with vega tooltip. Please have a re look at this PR

packages/hurumap-core/src/ChartTooltip/Tooltip.js
kelvinkipruto, line 1: @koechkevin Looks good. Shouldn't there be tests for `Tooltip`

kilemensi:
> @kelvinkipruto @kilemensi I was able to get `sx` to work with vega tooltip. Please have a re look at this PR

That's the thinking/tenacity I am about!!

kilemensi (commented):
👍🏽 

----

Another approach that we could have taken using the recommended [custom theme](https://github.com/vega/vega-tooltip/blob/main/docs/customizing_your_tooltip.md) is via MUI's [Global CSS Override](https://mui.com/material-ui/customization/how-to-customize/#4-global-css-override). It can style even elements in a portal. We wouldn't even need a standalone Tooltip component since we'll just be styling the Vega-Tooltip itself.

apps/climatemappedafrica/src/components/HURUmap/Chart/index.js
```diff
@@ -28,7 +24,7 @@ const useStyles = makeStyles(() => ({
 function Chart({
   indicator,
   indicatorTitle,
-  secondaryIndicator: { indicator: secondaryIndicator },
+  secondaryIndicator: sI,
```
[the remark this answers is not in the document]
koechkevin: This was a workaround for an error I encountered @kilemensi `Cannot read property indicator of undefined`

koechkevin:
> 👍🏽
> 
> Another approach that we could have taken using the recommended [custom theme](https://github.com/vega/vega-tooltip/blob/main/docs/customizing_your_tooltip.md) is via MUI's [Global CSS Override](https://mui.com/material-ui/customization/how-to-customize/#4-global-css-override). It can style even elements in a portal. We wouldn't even need a standalone Tooltip component since we'll just be styling the Vega-Tooltip itself.

This could work as well @kilemensi but I'd love to have this approach tried to see if there are challenges that would necessitate it. it'd also mean a major change that would best be implemented in a new iteration/version.

PR #790, opened 2024-07-19
@vpnmanager script not running

koechkevin:
## Description

VPN manager script is not running because dist is not included in build and app.json is not in the root directory of docker build.
This PR fixes that.

Fixes # (issue)

## Type of change

## Checklist:


kilemensi:
_Heita_ @koechkevin ... we seem to be copying node_modules (and other test packages like jest and playwrite) even though the application uses output: standalone. Can you give context as to why?

koechkevin:
> _Heita_ @koechkevin ... we seem to be copying node_modules (and other test packages like jest and playwrite) even though the application uses output: standalone. Can you give context as to why?

@kilemensi the scripts are running independent of next server forcing us to include node_modules.

For `other test packages like jest and playwrite`, pnpm install step was failing forcing me to include.

kilemensi:
>@kilemensi the scripts are running independent of next server forcing us to include node_modules.
>

Seems like a heavy penalty to pay for that @koechkevin ... can't we move the script into Next.js e.g. API endpoint and just call that end point periodically?

koechkevin:
> >@kilemensi the scripts are running independent of next server forcing us to include node_modules.
> 
> >
> 
> 
> 
> Seems like a heavy penalty to pay for that @koechkevin ... can't we move the script into Next.js e.g. API endpoint and just call that endpoint periodically?

Sure @kilemensi

koechkevin:
@CodeForAfrica/tech

kelvinkipruto (approved):

m453h (approved):

.github/workflows/vpnmanager-deploy-dev.yml
```diff
@@ -57,6 +57,7 @@ jobs:
             SENTRY_ENVIRONMENT=${{ env.SENTRY_ENVIRONMENT }}
             SENTRY_ORG=${{ secrets.SENTRY_ORG }}
             SENTRY_PROJECT=${{ secrets.VPNMANAGER_SENTRY_PROJECT }}
+            SECRET_TOKEN=${{ secrets.VPNMANAGER_SECRET_TOKEN }}
```
kilemensi, line 60: Secret for what? e.g. VPNMANAGER_SECRET_TOKEN, VPNMANAGER_API_SECRET/KEY, etc.

Dockerfile
```diff
@@ -599,9 +601,11 @@ COPY --from=vpnmanager-builder --chown=nextjs:nodejs /workspace/apps/vpnmanager/
 # https://nextjs.org/docs/advanced-features/output-file-tracing
 COPY --from=vpnmanager-builder --chown=nextjs:nodejs /workspace/apps/vpnmanager/.next/standalone ./apps/vpnmanager
 COPY --from=vpnmanager-builder --chown=nextjs:nodejs /workspace/apps/vpnmanager/.next/static ./apps/vpnmanager/.next/static
-
+COPY --from=vpnmanager-builder --chown=nextjs:nodejs /workspace/apps/vpnmanager/scripts ./apps/vpnmanager/scripts
+COPY --from=vpnmanager-builder --chown=nextjs:nodejs /workspace/apps/vpnmanager/app.json ./app.json
 USER nextjs
 
+ENV SECRET_TOKEN=${SECRET_TOKEN}
```
kilemensi, line 608:
```suggestion
```

`ENV` are stored in the image after build (`ARG` disappear after build)... don't store anything sensitive in `ENV` when building images.

apps/vpnmanager/scripts/processGsheet.js
```diff
@@ -0,0 +1,12 @@
+async function main() {
+  const { NEXT_PUBLIC_APP_URL, SECRET_TOKEN } = process.env;
```
kilemensi, line 2: May be it works (as this script is probably called outside of Next.js), but when it comes to Next.js apps it's best not to [destructure](https://nextjs.org/docs/app/api-reference/next-config-js/env) process.env:
>Next.js will replace process.env.customKey with 'my-value' at build time. Trying to destructure process.env variables won't work due to the nature of webpack [DefinePlugin](https://webpack.js.org/plugins/define-plugin/).

apps/vpnmanager/src/pages/api/processGsheet.ts
```diff
@@ -0,0 +1,15 @@
+import { NextApiResponse, NextApiRequest } from "next";
+import { processNewUsers } from "@/vpnmanager/lib/processUsers";
+
+export async function handler(req: NextApiRequest, res: NextApiResponse) {
+  try {
+    const key: string = req.headers["x-api-key"] as string;
+    const SECRET_TOKEN = process.env.SECRET_TOKEN;
+    if (!key || key !== SECRET_TOKEN) {
```
kilemensi, line 8:
```suggestion
    if (!(key && key !== SECRET_TOKEN)) {
```

apps/vpnmanager/scripts/processGsheet.js
```diff
@@ -0,0 +1,12 @@
...
+  const headers = {
+    "x-api-key": SECRET_TOKEN ?? "",
+  };
+  const res = await fetch(`${NEXT_PUBLIC_APP_URL}/api/processGsheet`, {
+    headers,
+  });
+  return res.json();
+}
+
+main().then(console.log);
```
kilemensi, line 12: 1. I'm assuming this is not logging anything sensitive?
2. Why not use await?

.github/workflows/vpnmanager-deploy-dev.yml
koechkevin: API secret key to ensure the script is not abused.

Dockerfile
```diff
@@ -582,6 +582,7 @@ RUN pnpm --filter "./apps/vpnmanager" build
 
 FROM base-runner as vpnmanager-runner
 
+ARG SECRET_TOKEN
```
kilemensi, line 585:
```suggestion
ARG API_SECRET_KEY
```

Dockerfile
```diff
@@ -599,7 +600,8 @@ COPY --from=vpnmanager-builder --chown=nextjs:nodejs /workspace/apps/vpnmanager/
 # https://nextjs.org/docs/advanced-features/output-file-tracing
 COPY --from=vpnmanager-builder --chown=nextjs:nodejs /workspace/apps/vpnmanager/.next/standalone ./apps/vpnmanager
 COPY --from=vpnmanager-builder --chown=nextjs:nodejs /workspace/apps/vpnmanager/.next/static ./apps/vpnmanager/.next/static
-
+COPY --from=vpnmanager-builder --chown=nextjs:nodejs /workspace/apps/vpnmanager/scripts ./apps/vpnmanager/scripts
+COPY --from=vpnmanager-builder --chown=nextjs:nodejs /workspace/apps/vpnmanager/app.json ./app.json
```
kilemensi, line 604: This should be in contrib/dokku folder (I'm assuming it's only used by Dokku)

kilemensi (approved):
🚀 

--

Doesn't turbo complain about new env vars such as `API_SECRET_KEY`?

Dockerfile
koechkevin: @kilemensi `app.json` should be at the root for the scripts to be executed

kilemensi: Location of `app.json` is [configurable](https://dokku.com/docs/advanced-usage/deployment-tasks/?h=app.json+location#changing-the-appjson-location) @koechkevin

PR #810, opened 2024-08-01
@hurumap/DonutChart

koechkevin:
## Description
This PR moves BarchartScope to Hurumap-Core.
 [Vega Embed](https://github.com/vega/vega-embed) is not a usual React Component which would have been an ideal candidate to move to @hurumap/core.

Currently, various chart(Barchart, LineChart, DonutChart etc) specs have been grouped in their corresponding scopes with common Scope separated.

This PR exports object Scope from hurumap-core with key value pairs of various charts to be accessed as 
```
import { Scope } from "@hurumap/core";
const { DonutChartScope } = Scope;
```
Fixes # 801

## Type of change

## Checklist:

### DIFF
--- a/apps/climatemappedafrica/src/components/HURUmap/Chart/configureScope.js
+++ b/apps/climatemappedafrica/src/components/HURUmap/Chart/configureScope.js
@@ -1,6 +1,5 @@
 import { Scope } from "@hurumap/core";
 
-import DonutChartScope from "./DonutChartScope";
 import LineChartScope from "./LineChartScope";
 import MultiLineChartScope from "./MultiLineChartScope";
 import StackedChartScope from "./StackedChartScope";
@@ -11,7 +10,7 @@ import VerticalStackedChartScope from "./VerticalStackedChartScope";
 import { hurumapArgs } from "@/climatemappedafrica/config";
 import theme from "@/climatemappedafrica/theme";
 
-const { BarChartScope } = Scope;
+const { BarChartScope, DonutChartScope } = Scope;
 
 export default function configureScope(
   indicator,
@@ -42,6 +41,19 @@ export default function configureScope(
     isCompare,
     isMobile,
   ];
+  const donutScopeOptions = {
+    primaryData: indicator?.data,
+    metadata: indicator?.metadata,
+    config: configuration,
+    secondaryData: secondaryIndicator?.data ?? null,
+    primaryParentData: showParent ? indicator?.parentData : [{}],
+    secondaryParentData: showParent ? secondaryIndicator?.parentData : [{}],
+    profileNames,
+    isCompare,
+    isMobile,
+    theme,
+    args: hurumapArgs,
+  };
   switch (chartType) {
     case "line":
       if (configuration?.stacked_field) {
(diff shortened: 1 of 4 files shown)


apps/climatemappedafrica/src/components/HURUmap/Chart/configureScope.js
```diff
@@ -42,6 +41,19 @@ export default function configureScope(
...
+    config: configuration,
+    secondaryData: secondaryIndicator?.data ?? null,
+    primaryParentData: showParent ? indicator?.parentData : [{}],
+    secondaryParentData: showParent ? secondaryIndicator?.parentData : [{}],
+    profileNames,
+    isCompare,
+    isMobile,
+    theme,
+    args: hurumapArgs,
+  };
```
kilemensi, line 56: May be this was discussed in the previous PR and I missed it but why are we creating a specifc `donutScopeOptions` just for `Donut` scope when the previous implementation used the [same](https://github.com/CodeForAfrica/ui/blob/f7f68dbe02bdbe252ba64ba07d9f7207b78f007c/apps/pesayetu/src/components/HURUmap/Chart/configureScope.js#L40-L52) `scopeOptions` for `Line`/`Multiline`, `Treemap` and `Donut`?

kilemensi (commented):
👍🏽

apps/climatemappedafrica/src/components/HURUmap/Chart/configureScope.js
koechkevin: @kilemensi when moving to packages, we are changing from passing multiple arguments to ChartScope functions to passing a single object. For now it is only DonutChart that uses these options hence the name.

koechkevin: The initial approach means that each scopeFunction takes 10 parameters. We are introducing 2 parameters to a function, I think the approach of passing positional arguments won't be sustainable in the long term if we intend to add more parameters @kilemensi

kilemensi: Yeah, 10 is a lot... but there are (at least) two things happening here:
1. Switching from positional to object argument passing
2. Moving away from common scopeOptions.

I think you can do `1` without doing `2`.

... but having gone through scope docs, what are the parameters that **must** be passed into a scope?

koechkevin: @kilemensi Scope is our internal function that returns an object of specifications in [this](https://vega.github.io/schema/vega/v5.json) schema.

koechkevin: 2. We are not entirely moving away from common scopeOptions, I just changed so it doesn't break those charts that we haven't moved to core packages @kilemensi

m453h: @koechkevin , I like the direction you took, I believe after we move all the charts the existing `scopeOptions` will be gone... I kept the current `scopeOptions` as depecrated on  #814 ...and decided that we should stick with just one common `scopeOption` object

koechkevin: We can go with that @m453h I have updated my branch

m453h (approved):

kelvinkipruto (approved):

kilemensi:
I hope this is not waiting for me @koechkevin  ... if you've got enough reviews from the team please proceed.

PR #840, opened 2024-08-14
Roboshield/live_preview_fix

koechkevin:
## Description

[Payloadcms React Use Live Preview](https://payloadcms.com/docs/live-preview/frontend) is a good tool for live preview and works well on the admin panel. It adds us the code snippet below.
```js
import { subscribe, unsubscribe, ready } from '@payloadcms/live-preview'
import { useCallback, useEffect, useState, useRef } from 'react'

export const useLivePreview = <T extends any>(props: {
  depth?: number
  initialData: T
  serverURL: string
}): {
  data: T
  isLoading: boolean
} => {
  const { depth = 0, initialData, serverURL } = props
  const [data, setData] = useState<T>(initialData)
  const [isLoading, setIsLoading] = useState<boolean>(true)
  const hasSentReadyMessage = useRef<boolean>(false)

  const onChange = useCallback((mergedData) => {
    // When a change is made, the `onChange` callback will be called with the merged data
    // Set this merged data into state so that React will re-render the UI
    setData(mergedData)
    setIsLoading(false)
  }, [])

  useEffect(() => {
    // Listen for `window.postMessage` events from the Admin panel
    // When a change is made, the `onChange` callback will be called with the merged data
    const subscription = subscribe({
      callback: onChange,
      depth,
      initialData,
      serverURL,
    })

    // Once subscribed, send a `ready` message back up to the Admin panel
    // This will indicate that the front-end is ready to receive messages
    if (!hasSentReadyMessage.current) {
      hasSentReadyMessage.current = true

      ready({
        serverURL
      })
    }

    // When the component unmounts, unsubscribe from the `window.postMessage` events
    return () => {
      unsubscribe(subscription)
    }
  }, [serverURL, onChange, depth, initialData])

  return {
    data,
    isLoading,
  }
}
```
However, data returned by this hook doesn't change when initialData changes e.g on soft URL navigation. We need to ensure that this doesn't break links.

This PR removes the package and adds  the packages that it uses under the hood independently and add a fix for our use case.

We are adding this snippet to the hook to ensure that input data takes precedence over changed document in case initialData changes.

```js
useEffect(() => {
    setData(initialData);
  }, [initialData]);
```
Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/apps/civicsignalblog/src/utils/useLivePreview.js
+++ b/apps/civicsignalblog/src/utils/useLivePreview.js
@@ -1,3 +1,4 @@
+// Adapted from https://github.com/payloadcms/payload/blob/main/packages/live-preview-react/src/index.ts
 import { subscribe, unsubscribe, ready } from "@payloadcms/live-preview";
 import { useCallback, useEffect, useState, useRef } from "react";
 
--- a/apps/roboshield/package.json
+++ b/apps/roboshield/package.json
@@ -1,6 +1,6 @@
 {
   "name": "roboshield",
-  "version": "0.1.9",
+  "version": "0.1.10",
   "private": true,
   "scripts": {
     "build-server": "tsc --project tsconfig.server.json",
--- a/apps/roboshield/package.json
+++ b/apps/roboshield/package.json
@@ -30,7 +30,7 @@
     "@next/third-parties": "^14.2.5",
     "@payloadcms/bundler-webpack": "^1.0.7",
     "@payloadcms/db-mongodb": "^1.7.1",
-    "@payloadcms/live-preview-react": "^0.2.0",
+    "@payloadcms/live-preview": "^0.2.2",
     "@payloadcms/plugin-cloud-storage": "^1.1.3",
     "@payloadcms/plugin-nested-docs": "^1.0.12",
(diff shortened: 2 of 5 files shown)


kelvinkipruto (approved):
@koechkevin Since it is the same code as the package but with a custom change to suit our use case, would the best approach be to [patch](https://pnpm.io/cli/patch) the original package.

kelvinkipruto (commented):
☝️

m453h (approved):

koechkevin:
> @koechkevin Since it is the same code as the package but with a custom change to suit our use case, would the best approach be to [patch](https://pnpm.io/cli/patch) the original package.

@kelvinkipruto good catch, but I think that's an overkill because it is tantamount to editing an already minified file and would be a challenge maintaining it.

kilemensi:
For those of us in the dark @koechkevin, can you show/describe the code we're changing and may be why?

And did we may be report the error back to Payload? They seem to be heavy users of Next.js so I'll be surprised if this doesn't get fixed ASAP.

kilemensi:
Fixed in https://github.com/CodeForAfrica/ui/pull/841

PR #843, opened 2024-08-20
Choropleth Map

### LINKED ISSUE #812: @hurumap/next/Map Support rendering Choropleth maps
As a user, I want to be able to view and and show filtered regions on the map in a colour-coded way. Currently, all region polygons on the map (which is based on [pesayetu](https://pesayetu.pesacheck.org/) has one uniform colour, blue, which does not allow a user to compare the regions on the map view based on the region's value from the selected filters. The intended outcome seems possible from the `Data mapper` component of this Wazimap [instance](https://beta.youthexplorer.org.za/), see screenshot attached.

Image

kelvinkipruto:
## Description
This PR introduces a choropleth map.

Fixes #812

## Type of change

## Checklist:

### DIFF
--- a/apps/climatemappedafrica/src/components/ExplorePage/index.js
+++ b/apps/climatemappedafrica/src/components/ExplorePage/index.js
@@ -107,6 +107,39 @@ function ExplorePage({ panelProps, profile: profileProp, ...props }) {
   if (secondaryTags?.length) {
     tags.push(secondaryTags[secondaryTags.length - 1]);
   }
+
+  const { locations, mapType = "choropleth" } = props;
+
+  let choropleth = null;
+  if (mapType === "choropleth") {
+    const filteredLocations = locations.filter(({ count }) => count !== null);
+    const counts = filteredLocations.map(({ count }) => count);
+    const maxCount = Math.max(...counts);
+    const minCount = Math.min(...counts);
+
+    const getClassification = (count) => {
+      const range = maxCount - minCount;
+      const veryLowThreshold = minCount + range * 0.2;
+      const lowThreshold = minCount + range * 0.4;
+      const moderateThreshold = minCount + range * 0.6;
+      const highThreshold = minCount + range * 0.8;
+
+      if (count <= veryLowThreshold) return "very low";
+      if (count <= lowThreshold) return "low";
+      if (count <= moderateThreshold) return "moderate";
+      if (count <= highThreshold) return "high";
+      return "very high";
+    };
+
+    choropleth = filteredLocations.map(({ code, count }) => {
+      return {
+        code,
+        count,
+        classification: getClassification(count),
+      };
+    });
+  }
+
   return (
     <>
       <Box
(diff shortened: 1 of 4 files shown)


kilemensi:
👍🏽 @kelvinkipruto 

1. Any reason why all these computations are not done on the `Map` component itself? Seems like everything comes from `locations`/`mapType` which are sent to `Map` via `{...props}`
2. Are clorepleth values always going to be in band of 5 i.e. very low, low, etc.?
3. Aren't we supposed to show a `Legend` to the user?

m453h:
Awesome implementation  🚀  @kelvinkipruto 

Having looked at it locally, I have two observations:

1. I believe the color assigned to a region/country should not change when we load the children i.e. after clicking the Geometry example for Kenya (Let me know if I'm missing context)

Screenshot 2024-08-20 at 11 58 51
Screenshot 2024-08-20 at 11 59 10

2. How do we style the borders? I believe the border color should be configurable as the selected color band can greatly affect how the map looks like with a certain border color

kelvinkipruto:
> 1. I believe the color assigned to a region/country should not change when we load the children i.e. after clicking the Geometry example for Kenya (Let me know if I'm missing context)

@m453h This is because the children have not been assigned a count value, hence the default color will be used.

kelvinkipruto:
> 2\. Are clorepleth values always going to be in band of 5 i.e. very low, low, etc.?

@kilemensi I took inspiration from [CodeForAfrica/HURUmap-UI](https://github.com/CodeForAfrica/HURUmap-UI/blob/main/stories/mapit.stories.js#L210). Another suggestion would be for the user to provide a list of colours, then we can use them when calculating the densities.

kilemensi:
>.. Another suggestion would be for the user to provide a list of colours, then we can use them when calculating the densities.
>

Like https://github.com/OpenUpSA/wazimap-ng-technical-gitbook/blob/master/profile-configuation.md#choropleth @kelvinkipruto ?

kelvinkipruto:
> > .. Another suggestion would be for the user to provide a list of colours, then we can use them when calculating the densities.
> 
> Like [OpenUpSA/wazimap-ng-technical-gitbook@`master`/profile-configuation.md#choropleth](https://github.com/OpenUpSA/wazimap-ng-technical-gitbook/blob/master/profile-configuation.md?rgh-link-date=2024-08-20T14%3A43%3A27Z#choropleth) @kelvinkipruto ?

Yes @kilemensi .

kelvinkipruto:
@kilemensi @m453h Please test again

kilemensi (commented):
👍🏽 ... haven't looked at the code but I think our original design and placement of Legend is better. Is the current PR what the data team specifically asked (or based on any other design)?

S

kelvinkipruto:
> I think our original design and placement of Legend is better.

I agree @kilemensi. It won't hide some potions on the map


> Is the current PR what the data team specifically asked (or based on any other design)?

I tried to match the placement on Wazimap. Let's go with your suggestion above

kilemensi (approved):
👍🏽 

--

I still think the colours are not right. Please pass the [two](https://colorbrewer2.org/#type=sequential&scheme=OrRd&n=9) [resources](https://d3js.org/d3-scale-chromatic) to the data team for their review but lets deploy this.

PR #870, opened 2024-09-04
@Outlinevpn Process Daily usage statistics

### LINKED ISSUE #875: @vpnmanager Data transfer statistics
Currently, Outline vpn manager statistics are generated monthly by running [this script](https://github.com/CodeForAfrica/api/blob/011d319a9d6153943889108233993bbf7bc5e452/cfa_vpn/py/vpn.py#L87).
```sh
dist/cfa_vpn.py/outline_manager.pex -e
```
Propose a solution that allows a user get these stats easily without always having the intervention of tech team devs.

koechkevin:
## Description
Currently, Outline vpn manager statistics are generated monthly by running [this script](https://github.com/CodeForAfrica/api/blob/011d319a9d6153943889108233993bbf7bc5e452/cfa_vpn/py/vpn.py#L87).

```sh
dist/cfa_vpn.py/outline_manager.pex -e
```
We are currently proposing a solution that allows a user get these stats easily without always having the intervention of tech team devs.

[Here](https://docs.google.com/document/d/1PEuFz31Z1ZBg_ByV2cH9mdDtxfMYWBWvOyOx7hqk6xw/edit?usp=sharing) is the proposed implementation plan

This PR introduces 2 API endpoints.

1. `POST /api/userStatistics`: To query outline vpn API for the statistics, calculate daily usage then store it in a local database for analysis. This endpoint will be executed via a cron.
2. `GET /api/userStatistics`: Gets the stored statistics in the database as per the queries sent.

### Process Daily User Statistics.
This action is triggered by  `POST /api/userStatistics`. Since The Outline VPN API returns only userId and total data transferred by the user over time, we need to find a way to calculate daily usage by finding the difference from last fetch.

1. Get data transferred from Outline VPN API.
2. Get users from Outline API to get user data from user IDs.
3. Calculate the data transferred by each user. We achieve this by querying the previously fetched and stored cumulative data transfer and subtracting it from the current cumulative obtained from 1 above. This cron will be executed at least once a day at around 11 PM so every calculation will be computed to the day it was run. 
4. Store the computed data in the database. If a computation for the current date for the user ID already exists, we simply update it else create a new entry.
5. Return the processed records.

### Get statistics.
`GET /api/userStatistics`
This API allows users to fetch stored statistics from the database based on various filters. The records can be filtered by date, email, date range, user ID, and record ID. Results can also be grouped and ordered based on specific fields.
| Parameter           | Type     | Description                                                                                             | Example                               |
|---------------------|----------|---------------------------------------------------------------------------------------------------------|---------------------------------------|
| `date`              | `string` | Filter results by a specific date in `YYYY-MM-DD` format.                                                | `?date=2024-09-03`                   |
| `email`             | `string` | Filter results by the user’s email address.                                                              | `?email=kevin.koech@codeforafrica.org` |
| `dateBetween.start` | `string` | Start of the date range for filtering. Requires `dateBetween.end`.                                       | `?dateBetween.start=2024-09-03`       |
| `dateBetween.end`   | `string` | End of the date range for filtering. Requires `dateBetween.start`.                                       | `?dateBetween.end=2024-09-04`         |
| `ID`                | `number` | Filter results by a specific record ID.                                                                  | `?ID=45`                             |
| `userId`            | `number` | Filter results by a specific user ID.                                                                    | `?userId=107`                        |
| `orderBy`           | `string` | Order results by a specific field in either ascending (`ASC`) or descending (`DESC`) order.              | `?orderBy=usage DESC`                |
| `groupBy` *          | `string` | Group results by a specific field. Common options are `email`, `date`.                                   | `?groupBy=email`                     |

The response will return records in the format
```
[{
        "ID": 103,
        "userId": "107",
        "usage": 9899671,
        "date": "2024-9-3",
        "cumulativeData": 5958400566,
        "email": "kevin.koech@codeforafrica.org",
        "createdAt": "2024-09-03T13:07:43.464Z"
    }]
```
Grouped by Email

```
[{
        "email": "kevin.koech@codeforafrica.org",
        "userId": "107",
        "totalUsage": 9985821081
    }]
```
Grouped by Date
```js
[{
        "date": "2024-9-4",
        "totalUsage": 21785853001
    }]
```
Fixes #875 

## Type of change

## Checklist:


apps/roboshield/next-env.d.ts
kelvinkipruto, line 1: I don't think this file should be part of the PR

koechkevin:
@m453h refer to this when reviewing implementation plan.

apps/vpnmanager/src/pages/api/userStatistics.ts
kelvinkipruto, line 1: Looks good @koechkevin. Is there need for authentication such that users can only fetch their stats? From the current implementation, it looks like anyone can fetch anybody's stats as long as they have their email adress.

apps/vpnmanager/src/lib/data/database.ts
```diff
@@ -0,0 +1,149 @@
...
+
+  static getAll(filters: Filters = {}) {
+    let query = "SELECT";
+    const params: any[] = [];
+    if (filters.groupBy === "email" || filters.groupBy === "date") {
+      if (filters.groupBy === "email") {
+        query +=
+          " email, userId, accessUrl, SUM(usage) as totalUsage FROM records WHERE 1=1";
+      }
+      if (filters.groupBy === "date") {
```
kilemensi, line 104:
```suggestion
      } else {
```

At this point inside the main if, `groupBy` can only be `email` or `date`, no?

apps/vpnmanager/src/lib/data/database.ts
```diff
@@ -0,0 +1,149 @@
...
+        query += " date, SUM(usage) as totalUsage FROM records WHERE 1=1";
+      }
+    } else {
+      query += " * FROM records WHERE 1=1";
```
kilemensi, line 108: Seems like `" FROM records WHERE 1=1"` is repeated unnecessarily 3 times in each if branch.

apps/vpnmanager/src/lib/data/database.ts
```diff
@@ -0,0 +1,149 @@
...
+      params.push(filters.email);
+    }
+    if (filters.date) {
+      query += " AND date = ?";
+      params.push(filters.date);
+    }
+    if (filters.dateBetween && !filters.date) {
+      query += " AND date BETWEEN ? AND ?";
+      params.push(filters.dateBetween.start, filters.dateBetween.end);
+    }
```
kilemensi, line 121: Isn't it better to have make `date = string | { start: string, end: string }` so that you only need to check if typeof date is string or object?

apps/vpnmanager/src/lib/userStatistics.ts
```diff
@@ -0,0 +1,82 @@
...
+      email: userDetails?.name || "",
+      accessUrl: userDetails?.accessUrl,
+    };
+    addUserStatsToDb({ ...newData, createdAt: new Date().toISOString() });
+    return newData;
+  });
+  return unprocessedUsers;
+}
+
+export async function getStats(filters: { [key: string]: string }) {
```
kilemensi, line 65: Why not
```suggestion
export async function getStats(filters: Filters) {
```
? Or isn't this why we declared the interface?

apps/vpnmanager/src/pages/api/userStatistics.ts
kilemensi: By default, Next.js API are only available to the front-end i.e. you can't call these APIs from different domain.

Does this end point return stats of _individual_ users or **_all_** users?

kilemensi (commented):
👍🏽 

--

Think about naming and structuring things e.g. in lib, may be there should be an `outline.js` file that handles pulling data from the VPN backend. API end points, may be they can be a bit more RESTful i.e. /api/statics with PUT or POST updating the stats and GET returning the stats, etc.

Update 1
On APIs, I think the `processGSheet` is the biggest offender ...

apps/vpnmanager/src/pages/api/userStatistics.ts
koechkevin: @kelvinkipruto the whole functionality is not to be used by individual users but rather by security team to obtain companywide statistics/reports. At the moment, we haven't handled the authentication part which would restrict who has access to the whole app.

apps/roboshield/next-env.d.ts
koechkevin: This is auto-generated @Kipruto

m453h (commented):
🚀 

Thanks @koechkevin , I've gained further context after carefully reading through the implementation plan and it looks solid.  I could be missing something but looking at the Table, I was thinking that [UPSERT](https://www.sqlite.org/lang_upsert.html) could simplify the whole operation of calculating and inserting the user stats if we keep `userId` and `date` as unique constraints.

kilemensi:
The key (`accessUrl`) you posted on above was real and working @koechkevin . I had to delete it for obvious reasons ... Hopeful you'll be able to create a new one.

I think lets remove `accessUrl` **completely** from all stats functionality. The only time we should access it is when we need to email it to the user **_AND_** it must never be stored or logged anywhere outside the VPN itself.

koechkevin:
[The data usage API](https://github.com/Jigsaw-Code/outline-server/blob/fa3d7c51fabcb8b2ede86cd59ba10b29047d0d37/src/shadowbox/server/manager_service.ts#L592) was hardcoded only to obtain data transferred in the last (30 * 24) hours to create a sliding window problem, which makes the daily computations we have proposed in this PR inaccurate.

We may have to open a PR for the above fix or fork the repository to deploy.

kilemensi:
> [The data usage API](https://github.com/Jigsaw-Code/outline-server/blob/fa3d7c51fabcb8b2ede86cd59ba10b29047d0d37/src/shadowbox/server/manager_service.ts#L592) was hardcoded only to obtain data transferred in the last (30 * 24) hours creating a sliding window problem, which makes the daily computations we have proposed in this PR inaccurate.
> 
> We may have to open a PR for the above fix or fork the repository to deploy.

Not sure I fully follow @koechkevin (or have full context) but:
1. Haven't we known this all along i.e. isn't this how the current script works @thepsalmist ?
2. Why would we need to fork the repo or another PR?

koechkevin:
@m453h / @kelvinkipruto

m453h (approved):
LGTM 🚀 

Just an additional thought,  the `Filters` interface could have stricter types for `orderBy` since we know the fields we would expect

PR #888, opened 2024-09-12
@/codeforafrica - fix: Partners form not rendering

m453h:
## Description
This PR resolves an issue in the `codeforafrica` CMS where the Partners collection form fails to render. The issue is fixed by passing an empty object `{}` as a parameter when calling `slateEditor`.

## Type of change

## Screenshots
Screenshot 2024-09-12 at 09 34 32

## Checklist:

### DIFF
--- a/apps/codeforafrica/src/payload/collections/Partners.js
+++ b/apps/codeforafrica/src/payload/collections/Partners.js
@@ -50,7 +50,7 @@ const Partners = {
       name: "description",
       required: true,
       localized: true,
-      editor: slateEditor(),
+      editor: slateEditor({}),
     }),
     socialLinks({
       name: "connect",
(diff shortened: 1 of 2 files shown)


apps/codeforafrica/src/payload/collections/Partners.js
```diff
@@ -50,7 +50,7 @@ const Partners = {
       name: "description",
       required: true,
       localized: true,
-      editor: slateEditor(),
+      editor: slateEditor({}),
```
kilemensi, line 53: Very bad API design from Payload's part!

kilemensi (approved):

apps/codeforafrica/src/payload/collections/Partners.js
m453h: Yeah, its like we are always expected to pass some configs which is far from the truth for most simple use cases

koechkevin (approved):

PR #889, opened 2024-09-16
@vpnamanager UI outline vpn user stats

koechkevin:
## Description

This PR Adds UI for VPN Manager.
Fixes # 

## Type of change

## Screenshots
image
image

## Checklist:


apps/vpnmanager/src/components/Statistics/Statistics.tsx
```diff
@@ -0,0 +1,313 @@
...
+    (async () => {
+      try {
+        const params: any = {};
+        Object.keys(filters).forEach((key) => {
+          if (filters[key as keyof typeof filters]) {
+            params[key] = filters[key as keyof typeof filters];
+          }
+        });
+        const res = await fetchJson.get("/api/statistics", {
+          params: params as Record<string, string>,
```
DavidTheProgrammer, line 67: Overriding the type here is a little tricky, let's satisfy it instead. let's declare params as type `Record<keyof typeof filters, string>` then work to make the iteration above compliant as well using `Object.fromEntries()`:

```
...
const params: Partial<Record<keyof typeof filters, string>> = Object.fromEntries(
          Object.entries(filters).filter(([_, value]) => Boolean(value)),
        );

        const res = await fetchJson.get("/api/statistics", {
          params: params,
        });
...
```

apps/vpnmanager/src/components/Statistics/Statistics.tsx
```diff
@@ -0,0 +1,313 @@
...
+
+  const handleFilterChange = (e: React.ChangeEvent<HTMLInputElement>) => {
+    setFilters({ ...filters, [e.target.name]: e.target.value });
+  };
+
+  const applyFilters = () => {
+    const params: any = {};
+    Object.keys(filters).forEach((key) => {
+      if (filters[key as keyof typeof filters]) {
+        params[key] = filters[key as keyof typeof filters];
```
DavidTheProgrammer, line 82: We can use a similar approach here to my recommendation above.

apps/vpnmanager/src/middleware.ts
DavidTheProgrammer, line 1: This looks fine, I understand what it's doing but it's a little verbose. Let's adjust it a little to make it's intention clear, firstly, we don't have to force the type with the `as string` the inferred type of `string | null` is fine. Second, we don't need that `key` check because if `null`  is the value of `key`, the equality check will fail either way so we don't have to check for the truthy-ness of the value. The following should work just as well unless I missed something:

```
 const key = req.headers.get("x-api-key");
  const API_SECRET_KEY = process.env.API_SECRET_KEY;
  if (req.method !== "GET") {
    if (key !== API_SECRET_KEY) {
      return Response.json(
        { success: false, message: "INVALID_API_KEY" },
        { status: 403 },
      );
    }
  }
  ```

File: apps/vpnmanager/src/pages/index.tsx [hunk 3]
```diff
@@ -1,10 +1,20 @@
...
+interface Props {
+  data: Data[];
+}
+export default function Home(props: Props) {
+  const { data } = props;
+  return <Statistics data={data} />;
 }
 
 export async function getStaticProps() {
+  const data = await getStats({ query: { orderBy: "date DESC" } } as any);
```
DavidTheProgrammer, line 14: `as any` is a code smell, let's not use it for areas where we have full control. I see that the issue is caused by declaring the argument in `getStats` to be of type `NextRequest`, let's instead narrow the value there to fit this invocation and get rid of the `as any`. I've left a comment there on my suggestion

apps/vpnmanager/src/lib/statistics.ts
DavidTheProgrammer, line 1: Let's narrow the input type of the `getStats` method from the current type of NextApiRequest to the type of it's query value as it seems that's what expect.  Something like the following should work as expected

```
export async function getStats(query: NextApiRequest["query"]) {
  const filters: Partial<Filters> & {
    "date.start"?: string;
    "date.end"?: string;
  } = query;
...
```

DavidTheProgrammer (commented):
Looks good. Just a few type things that we can bang out quickly.

kilemensi:
Not a review of this PR @koechkevin but isn't it better to first have authentication in place before we start showing stats in the UI?

apps/vpnmanager/src/middleware.ts
koechkevin: @DavidTheProgrammer if by any chance we don't set the API key, we shouldn't allow any API endpoint accessed unauthenticated, so we have to check that key is not null.

DavidTheProgrammer: Can the application deploy if we don't set the API_SECRET_KEY? If so, then we need to address that. In that case your point would be valid as we would have `null === null`.

m453h (commented):
Just some additional things I've noted from the UI, on the filter options what is the difference between `ID` and `UserID`, is there value of allowing someone to filter by UserID instead of just an email (I'm assuming that UserID would be an internally generated value from outline)

koechkevin:
> Just some additional things I've noted from the UI, on the filter options what is the difference between `ID` and `UserID`, is there value of allowing someone to filter by UserID instead of just an email (I'm assuming that UserID would be an internally generated value from outline)
@m453h userId is the id assigned to that user in outline VPN whereas ID is our own ID in out database.

koechkevin:
> Just tried running this branch locally and I'm getting 👇🏽 @koechkevin ... the app shouldn't completely fail, no? Screenshot 2024-09-17 at 13 05 00

@kilemensi I have updated this to ensure it doesn't fail if you do not have a data directory.

DavidTheProgrammer (approved):
Looks good. 🚀

apps/vpnmanager/src/components/Statistics/Statistics.tsx
```diff
@@ -0,0 +1,314 @@
...
+}
+
+interface Props {
+  data: Data[];
+}
+
+const Statistics: React.FC<Props> = ({ data: result }) => {
+  const router = useRouter();
+  const yesterday = new Date();
+  yesterday.setDate(yesterday.getDate() - 1);
```
kelvinkipruto, line 40: Since we already have `date-fns` [installed](https://github.com/CodeForAfrica/ui/blob/235b9a61d33bd2fca5c26af5bc2c68d78ae3da60/pnpm-workspace.yaml#L89), could we maybe use [`startOfYesterday`](https://date-fns.org/v4.1.0/docs/startOfYesterday)

apps/vpnmanager/src/utils/formatDate.ts
kelvinkipruto, line 1: Similarly here, I presume we can replace both functions with [`format`](https://date-fns.org/v4.1.0/docs/format) from `date-fns`

kelvinkipruto (approved):

pnpm-workspace.yaml
```diff
@@ -126,6 +127,7 @@ catalog:
   js-yaml: ^4.1.0
   jscodeshift: ^0.15.2
   jsdom: ^24.1.3
+  json2csv: 6.0.0-alpha.2
```
kilemensi, line 130: Why are we using alpha release? And is this better than the more [popular](https://www.oneschema.co/blog/top-5-javascript-csv-parsers) libs like papaparse?

apps/vpnmanager/src/components/Statistics/Statistics.tsx
```diff
@@ -0,0 +1,313 @@
...
+  data: Data[];
+}
+
+const Statistics: React.FC<Props> = ({ data: result }) => {
+  const router = useRouter();
+  const yesterday = startOfYesterday();
+  const [filters, setFilters] = useState({
+    ID: router.query.ID || "",
+    userId: router.query.userId || "",
+    email: router.query.email || "",
```
kilemensi, line 43: Do we need all three? Remember this is a security app so we shouldn't show anything we don't **need** to show.

My feeling is we only need to show email address to the security team both Sqlite ID and Outline ID shouldn't be visible anywhere.

kilemensi (approved):
👍🏽 

---

When you take screenshots please blur real user emails, etc., everything VPN related info is supposed to be private.

PR #900, opened 2024-09-20
@vpnmanager auth

koechkevin:
## Description

This PR introduces login with Google and adds restriction of which emails can access the APP.

Fixes # (issue)

## Type of change

## Screenshots
image
image

## Checklist:


apps/vpnmanager/src/pages/api/auth/[...nextauth].ts
```diff
@@ -0,0 +1,32 @@
...
+      },
+    }),
+  ],
+  pages: {
+    signIn: "/login",
+    error: "/login",
+  },
+  callbacks: {
+    async signIn({ profile }) {
+      const allowedEmails = (process.env.ALLOWED_EMAILS ?? "").split(",");
```
Comment (m453h) [inline apps/vpnmanager/src/pages/api/auth/[...nextauth].ts:24]: I could be asking for too much, but I think it would be better if these allowed emails were coming from a database (since we already have SQLite) instead of the .env file.

Comment (koechkevin) [reply apps/vpnmanager/src/pages/api/auth/[...nextauth].ts:24]: @m453h your proposal makes sense. However I made this decision based on
- The list of people going to access the app is not going to change frequently, it will probably be a small number.
- We do not intend to have a UI to manage people who can access.

Comment (m453h) [reply apps/vpnmanager/src/pages/api/auth/[...nextauth].ts:24]: Cool, I got you...based on the case it seems to be the quickest solution 👍

apps/vpnmanager/README.md
```diff
@@ -43,3 +43,31 @@ or
...
+```bash
+dokku git:from-image vpnmanager codeforafrica/vpnmanager:latest
+```
+
+4. Persist storage database.
+   Docker in their [best practices](https://docs.docker.com/build/building/best-practices/#containers-should-be-ephemeral) that containers be treated as ephemeral. In order to manage persistent storage for database, a directory outside the container should be mounted. Vpnmanager uses sqlite to locally store data as obtained from Outline VPN API. To persist this data, run the command below
+
+```bash
+dokku storage:mount vpnmanager /var/lib/dokku/data/storage/vpnmanager/data:/workspace/apps/vpnmanager/data
+```
```
kilemensi, line 73: I think this should come before deployment, no?

apps/vpnmanager/contrib/dokku/scripts/apiClient.mjs
```diff
@@ -0,0 +1,26 @@
+export async function apiClient(endpoint, method = "POST", body = null) {
```
kilemensi, line 1: `apiClient` (noun) implies a something to help you talk to/connect to a server. However, the code appears to use it as a verb i.e. `result = await apiClient(`

apps/vpnmanager/contrib/dokku/scripts/processStats.mjs
```diff
@@ -0,0 +1,9 @@
+import { apiClient } from "./apiClient.mjs";
+
+async function main() {
+  const response = await apiClient("/api/statistics");
+  return response;
+}
+
+const responseJson = await main();
+console.log(responseJson);
```
kilemensi, line 9: ?

apps/vpnmanager/src/assets/icons/Type=google, Size=24, Color=currentColor.svg
```diff
@@ -0,0 +1,7 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="25" height="24" viewBox="0 0 25 24" fill="currentColor"><script xmlns=""/>
```
kilemensi, line 1: Does this really change `color` depending on `currentColor`?

apps/vpnmanager/src/middleware.ts
```diff
@@ -2,7 +2,7 @@ import type { NextRequest } from "next/server";
 
 // Limit the middleware to paths starting with `/api/`
 export const config = {
-  matcher: "/api/:function*",
+  matcher: ["/api/statistics/:path*", "/api/users/:path*"],
```
kilemensi, line 5: Can we add a comment why need a middleware for this?

kilemensi (commented):
👍🏽 

--

Are we switching to server-side checks?

apps/vpnmanager/README.md
koechkevin: Yes

m453h (commented):
🚀 

@koechkevin  I have been able to run the code and test it locally, here are some few things I think can be improved:

-  I think we should not be having this link 
Screenshot 2024-09-24 at 12 54 50 (1)


- On the filters, we should at-least be having labels to indicate what exactly the inputs are, without looking at the code, its a bit confusing to know how the date filters work (the date format text overrides the placeholder text and there are three filters).

Screenshot 2024-09-24 at 13 03 53

- On the `README.md`, perhaps its best to explicitly state the [scopes](https://developers.google.com/identity/protocols/oauth2/scopes) needed when configuring the Google Provider, from the code I see we are using (openid, email, profile).  This would really help someone that needs to configure Google Auth for the first time.

-  I  have some random data in my  database, when I refresh the page, I momentarily see this data and then it displays no data found, I think this should not be happening.
ezgif-3-6f1b1ad968

koechkevin:
@m453h @kelvinkipruto

apps/vpnmanager/src/pages/api/users.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const methodToFunction: RestMethodFunctions = {
+  POST: processNewUsers,
+};
+
+export async function handler(req: NextApiRequest, res: NextApiResponse) {
+  try {
+    const statFunc = methodToFunction[req.method as RestMethods];
+    if (!statFunc) {
+      return res.status(404).json({ message: "Requested path not found" });
```
kilemensi, line 13: Is this about path or method?

apps/vpnmanager/src/pages/index.tsx
```diff
@@ -10,8 +12,23 @@ export default function Home(props: Props) {
...
+export async function getServerSideProps(
+  context: GetSessionParams | undefined,
+) {
+  const yesterday = startOfYesterday();
+
+  const data = await getStats({
+    query: { orderBy: "date DESC", date: format(yesterday, "yyyy-MM-dd") },
+  });
+  const session = await getSession(context);
+  if (!session) {
```
kilemensi, line 24: Same here, why are you fetching data _before_ you check if there is an active session?

apps/vpnmanager/src/components/UserAvatar/UserAvatar.tsx
```diff
@@ -0,0 +1,51 @@
...
+  };
+
+  const handleLogout = () => {
+    handleMenuClose();
+    signOut();
+  };
+
+  if (!session) {
+    return null;
+  }
```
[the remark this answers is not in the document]
kilemensi: Cool... but if I understand the code, the profile can come as part of server-side props that goes to the NavBar

apps/vpnmanager/src/pages/index.tsx
```diff
@@ -10,8 +12,24 @@ export default function Home(props: Props) {
...
+      redirect: {
+        destination: "/login",
+        permanent: false,
+      },
+    };
+  }
+
+  const yesterday = startOfYesterday();
+
+  const data = await getStats({
```
kilemensi, line 30:
```suggestion
  const yesterday = startOfYesterday();
  const data = await getStats({
```

kelvinkipruto (approved):

kilemensi (approved):
🚀

m453h (approved):

PR #901, opened 2024-09-20
@/CivicSignalBlog - Add ability to manage CivicSignal content

m453h:
## Description
This PR enhances the CivicSignalBlog by adding the ability to manage CivicSignal [web-tools ](https://civicsignal.africa/#/home)content.

This implementation  allows users to switch between different instances of the web-tools which include: 
- Research Blog
- CivicSignal Site
- Explorer
- Topic Mapper
- Source Manager

This initial implementation only supports:
- Pages Collection: Allows managing basic pages specific to each tool.
- Global Settings: Includes global configuration options for each of the above applications.

This PR focuses on simplicity and paves the way for future enhancements.

## Type of change

## Screenshots

Screenshot 2024-09-20 at 17 08 02

## Checklist:


koechkevin (approved):
LGTM

apps/civicsignalblog/src/lib/data/json/applications.js
```diff
@@ -0,0 +1,12 @@
...
+  "Explorer",
+  "Topic Mapper",
+  "Source Manager",
+  "Research Blog",
+].map((label) => ({
+  label,
+  value: label.toLowerCase().replace(/\s+/g, "_"),
+}));
+
+export default applications;
```
kilemensi, line 12: Not sure why this is in `lib/data/json`... it's certainly not a JSON.

apps/civicsignalblog/src/lib/data/json/applications.js
```diff
@@ -0,0 +1,12 @@
+const applications = [
+  "CivicSignal",
+  "Explorer",
+  "Topic Mapper",
+  "Source Manager",
+  "Research Blog",
+].map((label) => ({
+  label,
+  value: label.toLowerCase().replace(/\s+/g, "_"),
+}));
```
kilemensi, line 10:
```js
export const MAIN = "main";
export const EXPLORER = "explorer";
export const TOPIC_MAPPER = "topic-mapper";
export const SOURCE_MAPPER = "source-manager";
export const RESEARCH = "research",

export default applications = {
  MAIN: "CivicSignal",
  EXPLORER: "Explorer",
  TOPIC_MAPPER: "Topic Mapper",
  SOURCE_MAPPER: "Source Manager",
  RESEARCH: "Research",
}
```

1. I think the above is much cleaner and allows `import { MAIN } from "applications"` enabling the slug to be reused without being retyped over and over i.e. single place to change if we ever need to.
2. I don't see **Media Data**

apps/civicsignalblog/tsconfig.json
```diff
@@ -20,7 +20,13 @@
       "@/commons-ui/next/*": ["../../packages/commons-ui-next/src/*"]
     }
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
+  "include": [
+    "next-env.d.ts",
+    "**/*.ts",
+    "**/*.tsx",
+    "src/payload/components/BeforeDashboard/index.jsx",
+    "src/payload/components/BeforeDashboard/index.jsx"
```
kilemensi, line 28: What does including/not including them here actually do?

apps/civicsignalblog/src/payload/collections/Users.js
```diff
@@ -1,3 +1,6 @@
+import payload from "payload";
+
+import applications from "../../lib/data/json/applications";
```
kilemensi, line 3: Not sure I fully understand the naming of "managed applications"? Who is managing them and why are they managing them?

I think authorised applications/apps or just applications/apps would make a lot more sense (unless there is a reason behind the "managed" part)

apps/civicsignalblog/src/payload/collections/Users.js
```diff
@@ -53,6 +56,104 @@ const Users = {
...
     },
+    {
+      name: "defaultManagedApplication",
+      type: "select",
+      hasMany: false,
+      admin: {
+        isClearable: true,
+        isSortable: true,
+      },
+      options: applications,
```
kilemensi, line 67: Should default to `research` app as that's what is already there.

kilemensi (commented):
Looks really good... kudos

---

1. Isn't there a way of using URL for switch apps? Or at least show the current app as part of the URL i.e. `http://localhost:3000/admin/research` or `http://localhost:3000/admin?app=research` or _something_.
2. Lets move the app switch to where user/locale is.
Screenshot 2024-09-23 at 14 35 27

VinneyJ (approved):

apps/civicsignalblog/src/payload/collections/Users.js
m453h: Makes sense I have made the change

apps/civicsignalblog/src/payload/collections/Users.js
m453h: I kept "managed" since they are apps whose content is being managed by our CMS 😄 ... but come to think of it, its somewhat redundant...I think we can keep it as just "app"

apps/civicsignalblog/tsconfig.json
m453h: I don't even know how that got there 🤦 , must've been automatically added, I've removed it !

apps/civicsignalblog/src/lib/data/json/applications.js
m453h: I have refactored the app name to something very close to your suggestion, since we use the app names in the labels for the select field in the user collection I had to make a minor adjustment to get  the `{value: "", key:""}` array.

The **Media Data** is actually just a page under the MAIN (CivicSignal) its not a separate app that's why I didn't include it in the list

m453h:
> Looks really good... kudos
> 
> 1. Isn't there a way of using URL for switch apps? Or at least show the current app as part of the URL i.e. `http://localhost:3000/admin/research` or `http://localhost:3000/admin?app=research` or _something_.
> 2. Lets move the app switch to where user/locale is.
> 
> Screenshot 2024-09-23 at 14 35 27

I've made the switch, we can show the URL based on the App the user has switched to when we redirect to the homepage but  we will loose it whenever we click any other link, the same behaviour is observed for the user locale. 
ezgif-1-c7d30d5837

kilemensi:
> I've made the switch, we can show the URL based on the App the user has switched to when we redirect to the homepage but we will loose it whenever we click any other link ...
>

Cool @m453h . It's minor so we can leave it out for now since everything is CivicSignal.

apps/civicsignalblog/src/lib/data/json/applications.js
kilemensi: Has this been addressed @m453h ?

apps/civicsignalblog/src/payload/access/canAccessApplication.js
```diff
@@ -0,0 +1,10 @@
+export default function canAccessApplication(user, searchString) {
+  if (user) {
+    const app = user.currentApp || user.defaultApp;
+    if (app) {
+      return app === searchString.toLowerCase();
+    }
+    return false;
+  }
+  return false;
```
kilemensi, line 9:
```suggestion
  if (user) {
    const app = user.currentApp || user.defaultApp;
    return app === searchString.toLowerCase();
  }
  return false;
```

Don't think we need `if`/multiple `return`s

apps/civicsignalblog/src/payload/collections/CivicSignalBlog/Authors.js
```diff
@@ -1,7 +1,9 @@
+import canRead from "../../access/applications/researchBlog";
```
kilemensi, line 1:
```suggestion
import canRead from "../../access/applications/research";
```

Similarly for the other apps, make them match what we've defined in `applications.js`

apps/civicsignalblog/src/payload/collections/CivicSignalBlog/Pages.js
```diff
@@ -0,0 +1,31 @@
+import canRead from "../../access/applications/researchBlog";
```
kilemensi, line 1: Can't we get aliases to work in Payload? Not necessary in this PR but these relative paths are hard to keep up with.

apps/civicsignalblog/src/payload/collections/CivicSignalBlog/Pages.js
```diff
@@ -0,0 +1,31 @@
...
+import CustomPageHeader from "../../blocks/CustomPageHeader";
+import Error from "../../blocks/Error";
+import FeaturedStories from "../../blocks/FeaturedStories";
+import LongForm from "../../blocks/LongForm";
+import PageHeader from "../../blocks/PageHeader";
+import Posts from "../../blocks/Posts";
+import pages from "../../fields/pages";
```
kilemensi, line 8: I don't think `pages` is a field... may be just a function in `utils`?

apps/civicsignalblog/src/payload/collections/Users.js
```diff
@@ -53,6 +56,76 @@ const Users = {
...
+    },
+  ],
+  endpoints: [
+    {
+      path: "/current-app",
```
kilemensi, line 84: Minor but
```suggestion
      path: "/apps/current",
```
may be a bit more RESTy? (or `users/me/apps/current`)

apps/civicsignalblog/src/payload/collections/Users.js
```diff
@@ -53,6 +56,76 @@ const Users = {
...
+
+        const updatedUser = await payload.update({
+          collection: "users",
+          id: userId,
+          data: {
+            currentApp: selectedApp || currentUser.defaultApp,
+          },
+        });
+
+        res.status(200).send({
```
kilemensi, line 123:
```suggestion
        }

        const { selectedApp } = req.body;
        if (!selectedApp) {
          res.status(400).send({
            error: `Incorrect message format was received:${JSON.stringify(req.body)}`,
          });
        }

        const userId = req.user.id;
        const currentUser = await payload.findByID({
          collection: "users",
          id: userId,
          showHiddenFields: true,
        });
        if (!currentUser) {
          res
            .status(404)
            .send({ error: "User with specified ID was not found" });
        }

        const updatedUser = await payload.update({
          collection: "users",
          id: userId,
          data: {
            currentApp: selectedApp || currentUser.defaultApp,
          },
        });
        res.status(200).send({
```

Too many empty lines make it harder to follow which parts of code are related/dependant.

apps/civicsignalblog/src/payload/components/actions/index.tsx
```diff
@@ -0,0 +1,98 @@
+import { useAuth } from "payload/components/utilities";
+import React, { useEffect, useState } from "react";
+
+import applications from "../../../lib/data/json/applications";
+
+function BeforeDashboard() {
+  const { user } = useAuth();
+
+  const [selectedApp, setSelectedApp] = useState(
+    String(user.currentApp ?? user.defaultApp) ?? "",
```
kilemensi, line 10:
```suggestion
    user.currentApp || user.defaultApp || RESEARCH,
```

I'd rather we use really defaults whenever we can instead of `""`. Alternatively, ensure user.defaultApp is never undefined/null/empty.

apps/civicsignalblog/src/payload/globals/Site/researchBlog.js
```diff
@@ -0,0 +1,18 @@
+import canRead from "../../access/applications/researchBlog";
+import settings from "../../fields/settings";
+
+import EngagementTab from "./Components/EngagementTab";
+import GeneralTab from "./Components/GeneralTab";
+import NavigationTab from "./Components/NavigationTab";
+
+const ResearchBlog = settings({
```
kilemensi, line 8:
```suggestion
const Research = settings({
```

apps/civicsignalblog/src/lib/data/json/applications.js
m453h: @kilemensi, I think much like `lib/data/json/countries.js` (which is currently not being referenced anywhere), I propose to move them to the `lib/data/common`

apps/civicsignalblog/src/payload/collections/CivicSignalBlog/Pages.js
m453h: Yeah that was the plan (I didn't do it in this PR to reduce the number of files changed)... working with relative paths has been a pain

apps/civicsignalblog/src/payload/collections/CivicSignalBlog/Pages.js
m453h: Done I have moved it to utils

apps/civicsignalblog/src/lib/data/json/applications.js
m453h: Decided to move it to payload/lib/data/common

apps/civicsignalblog/src/payload/collections/Research/Pages.js
```diff
@@ -0,0 +1,31 @@
...
+import CustomPageHeader from "../../blocks/CustomPageHeader";
+import Error from "../../blocks/Error";
+import FeaturedStories from "../../blocks/FeaturedStories";
+import LongForm from "../../blocks/LongForm";
+import PageHeader from "../../blocks/PageHeader";
+import Posts from "../../blocks/Posts";
+import pages from "../../utils/createPagesCollection";
+
+const Pages = pages({
+  pageSlug: "pages",
```
kilemensi, line 11: Minor and hence can be ignored or implemented in another PR but I think it makes a bit more sense to migrate `pageSlug` from `pages` to `${RESEARCH}-pages` now when we have have very few pages in PROD

apps/civicsignalblog/src/payload/components/actions/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+            throw new Error("Network response was not ok");
+          }
+        })
+        .then((data) => {
+          window.location.replace(`/admin?app=${encodeURI(data.currentApp)}`);
+        })
+        .catch((error) => {
+          setLoading(false);
+          throw error;
+        });
```
kilemensi, line 43: Minor but shouldn't we be using `async`? If we **_have to_** use promise, shouldn't there just be a single `finally` that sets loading to false?

apps/civicsignalblog/src/payload/components/actions/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }, [selectedApp]);
+
+  const handleChange = (event) => {
+    const selectedValue = event.target.value;
+    setSelectedApp(selectedValue);
+  };
+
+  return (
+    <div style={{}}>
```
kilemensi, line 53:
```suggestion
    <div>
```

apps/civicsignalblog/src/payload/components/actions/index.tsx
```diff
@@ -0,0 +1,98 @@
...
+          <p
+            style={{
+              padding: "5px 10px",
+              marginTop: "30px",
+              background: "var(--theme-input-bg)",
+            }}
+          >
+            Loading...
+          </p>
+        </>
```
kilemensi, line 65:
```suggestion
        <p
          style={{
            padding: "5px 10px",
            marginTop: "30px",
            background: "var(--theme-input-bg)",
          }}
        >
          Loading...
        </p>
```

apps/civicsignalblog/src/payload/globals/Site/research.js
```diff
@@ -0,0 +1,18 @@
+import canRead from "../../access/applications/research";
+import settings from "../../utils/createGlobalSettings";
+
+import EngagementTab from "./Components/EngagementTab";
+import GeneralTab from "./Components/GeneralTab";
+import NavigationTab from "./Components/NavigationTab";
+
+const Research = settings({
+  slug: "settings-site",
```
kilemensi, line 9: Same here, might be best to migrate to `settings-${RESEARCH}-site`

kilemensi (approved):
🚀

apps/civicsignalblog/src/payload/collections/Research/Pages.js
m453h: ✅ 

I've done it now, I think we'll  have to recreate the pages when we migrate to prod, but its a very minor issue

kilemensi: ...or a [migration](https://payloadcms.com/docs/database/migrations) script instead of recreating them.

PR #902, opened 2024-09-27
@/CivicSignalBlog -  Use aliases in Payload CMS

m453h:
## Description

This pull request includes several changes to the `civicsignalblog` application, primarily focusing on using **Path aliases** for cleaner imports.

Additionally, it fixes failing `build-next` due to ESLint that were noted after merging  #901.

While trying to use path aliases, it was noted that Payload does not resolve the paths as expected, this is dicussed [here](https://github.com/payloadcms/payload/issues/3157). 

Thus, to resolve this we have used [Subpath Import Aliases](https://nodejs.org/api/packages.html#subpath-imports) with custom conditions to ensure that paths are correctly resolved when we run the application in both prod and dev mode. 

This has been achieved by:
- Adding  the `imports` field to the `package.json` and setting the alias for the `src` folder to `#civicsignalblog`. It should be noted that as per the documentation of sub path imports (Entries in the "imports" field must always start with # to ensure they are disambiguated from external package specifiers.)
- Making the following changes in `tsconfig.json` and `tsconfig.server.json`:
   * `"module": "esnext",` to   `"module": "NodeNext"` 
   * `"moduleResolution": "node"` to   `"moduleResolution": "NodeNext"`
These changes have been made because subpath imports need a moduleResolution of `Node16` or `NodeNext`.

- Adding the same alias `#civicsignalblog`  i.e.  in webpack `resolve > alias` config  `payload.config.ts`
- Using [custom conditions](https://www.typescriptlang.org/tsconfig/#customConditions) in `package.json` which allows already built code to be imported from ` dist/` instead of `src/ `. We have used two conditions where `dev` resolves to `src/` and default that resolves to `dist/` thus updated the `dev` script with  `--conditions=dev`
- Adding `eslint-import-resolver-typescript` plugin for enabling ESLint to resolve import statements.
___
This has been done as an alternative solution that used two packages `tsconfig-paths` and tsc alias which involved the following steps:
- Added [tsconfig-paths](https://www.npmjs.com/package/tsconfig-paths) library and updated the `dev` script with `-r tsconfig-paths/register` enabling ts-node to correctly resolve path aliases defined in `tsconfig.json`
 - Added a custom function  in `tsconfigPathToWebpackAlias.ts` that create path aliases that are used in `payload.config.ts`
 - Added  [tsc-alias](https://www.npmjs.com/package/tsc-alias) and updated the `build-next` script to include `tsc-alias -p tsconfig.server.json` that replaces alias paths with relative paths after typescript compilation.

___
P.S
The PR might have many modified files but they are mainly adding aliases, one caveat with the current implementation is explained [here](https://github.com/CodeForAfrica/ui/pull/902#discussion_r1781181890) where we are forced to explicitly  include the filename in our imports even for cases where we have an `index.js` thus in some cases you will see that we are using 

`import slug from "#civicsignalblog/payload/fields/slug/index";` 

Instead of

 `import slug from "#civicsignalblog/payload/fields/slug";`

An issue that closely relates to this limitation (though it mainly discusses about exports) can be seen [here](https://github.com/nodejs/node/issues/37928).  

---

## Type of change

## Checklist:


kilemensi:
Just saw [this](https://medium.com/@vitaliypotapov/setting-up-subpath-import-aliases-in-a-typescript-project-3ee027b75f1d) custom condition approach that doesn't require extra dependencies. Did you review it / won't it work for us? We should be on TS 5.4 already.

m453h:
> Just saw [this](https://medium.com/@vitaliypotapov/setting-up-subpath-import-aliases-in-a-typescript-project-3ee027b75f1d) custom condition approach that doesn't require extra dependencies. Did you review it / won't it work for us? We should be on TS 5.4 already.

Hmm interesting... I hadn't looked at this one...let me quickly test it and get back to you

apps/civicsignalblog/src/payload/access/applications/research.js
```diff
@@ -1,6 +1,7 @@
-import { RESEARCH } from "../../lib/data/common/applications";
 import canAccessApplication from "../canAccessApplication";
```
kelvinkipruto, line 1: Wgy don't we use the alias for all imports?

apps/civicsignalblog/src/payload/admin/scss/custom.scss
```diff
@@ -1,13 +0,0 @@
-@import "~payload/scss";
```
kelvinkipruto, line 1: Was this not being used before?

kelvinkipruto (commented):
LG

koechkevin (commented):
@m453h Looks good to me, I may have missed something but why are we using `#civicsignalblog/*` as opposed to `@/civicsignalblog/*` that we've been using all along

apps/civicsignalblog/src/payload/access/applications/research.js
m453h: Cool I'll rectifty that one 👍

apps/civicsignalblog/src/payload/admin/scss/custom.scss
m453h: Nope, I think I had added this one when I was experimenting with adding the custom select component for apps in the dashboard, but It's no longer relevant since I moved all the styling in the component itself.

m453h:
> @m453h Looks good to me, I may have missed something but why are we using `#civicsignalblog/*` as opposed to `@/civicsignalblog/*` that we've been using all along

@koechkevin , we have to use `"#civicsignalblog/*"` because we are using [Subpath Imports](https://nodejs.org/api/packages.html#subpath-imports) we have no option but to use `#` since  "Entries in the "imports" field must always start with # to ensure they are disambiguated from external package specifiers."

apps/civicsignalblog/.eslintrc.js
```diff
@@ -1,11 +1,21 @@
 module.exports = {
   root: true,
   extends: ["eslint-config-commons-ui/next"],
+  rules: {
+    "react/jsx-filename-extension": [
+      "warn",
+      { extensions: [".js", ".jsx", "ts", ".tsx"] },
```
kilemensi, line 7: Why? we shouldn't use `.jsx` and in TS, jsx is only allowed in `.tsx`.

apps/civicsignalblog/eslint.webpack.config.js
```diff
@@ -23,4 +23,8 @@ module.exports = {
     },
     extensions: [".js"],
   },
+  typescript: {
+    alwaysTryTypes: false,
+    project: "./",
```
kilemensi, line 28: What is the difference when specifying `project: "./"` vs project: `"./tsconfing.json"`?

apps/civicsignalblog/src/payload/collections/Research/Tags.js
```diff
@@ -1,5 +1,5 @@
-import canRead from "../../access/applications/research";
-import slug from "../../fields/slug";
+import canRead from "#civicsignalblog/payload/access/applications/research";
+import slug from "#civicsignalblog/payload/fields/slug/index";
```
kilemensi, line 2: Doesn't it work without specifying `index`?

apps/civicsignalblog/src/payload/components/actions/index.tsx
```diff
@@ -34,6 +36,7 @@ function BeforeDashboard() {
           }
 
           const data = await response.json();
+          // eslint-disable-next-line no-undef
```
kilemensi, line 39: What could potentially be undef?

kilemensi (commented):
👍🏽 

--

Can you update the PR description to match this new direction we've taken?

m453h:
> 👍🏽
> 
> --
> 
> Can you update the PR description to match this new direction we've taken?

Sure, I was planning to do that...I just need to resolve one minor issue with the imports, currently we cant implicitly import from folders (i.e. the index.js file doesn't get resolved automatically thus we are  are forced to import such files by explicitly appending `index` in our import statements

e.g. in `createPagesCollection.js` we are importing fullTiltle and slug as:

` import fullTitle from "#civicsignalblog/payload/fields/fullTitle/index";` 
`import slug from "#civicsignalblog/payload/fields/slug/index";`

apps/civicsignalblog/src/payload/collections/Research/Tags.js
m453h: @kilemensi, currently not working I'm looking and the package.json and tsconfig.json configs...should've been straight forward but its not resolving automatically...this is the major blocker, same as explained [here](https://github.com/CodeForAfrica/ui/pull/902?notification_referrer_id=NT_kwDOAGhUF7MxMjYwNTM0Njg5OTo2ODM3Mjcx#issuecomment-2382555427)

apps/civicsignalblog/src/payload/components/actions/index.tsx
m453h: @kilemensi , `window` is undefined during the build process and would fail

kilemensi: ... then should you add `/* eslint-env browser */` at the top of the file to let eslint know this file will run in `browser` env rather than silencing the issue?

apps/civicsignalblog/.eslintrc.js
m453h: I think we need to completely remove this rule we do have a lot of jsx in js files and I believe there is no need to be too strict or treat them differently...I had introduced it as I was trying to resolve some linting issues on my editor, but I've just crosschecked it has nothing to do with this config  🫠

apps/civicsignalblog/src/payload/collections/Research/Tags.js
m453h: @kilemensi, I have done a lot of digging around in this one, there is no clean way (atleast I can think of) to escape using index when we are importing...the limitations I am facing is that with subpath imports there is no fallback mechanism, It would've been ideal if there was fallback mechanism when we are resolving the paths in the array of URLs we have configured 
```
[
    "./src/*/index.js",
    "./src/*.js",
    "./src/*"
]
```
However, on first match of the string pattern if no file is found it will throw an error.

One way I have tested that could work is by explicitly configuring the full path in the package.json to match like this 👇 
```
    "#civicsignalblog/payload/fields/slug": {
      "dev": [
        "./src/payload/fields/slug/index.js"
      ],
      "default": [
        "./src/payload/fields/slug/index.js"
      ]
    },

```
But this seems a bit too much, for the problem we are trying to solve...I'm open to suggestions if we should merge it as is for now

m453h: I'll create this as a separate issue we can explore later !

apps/civicsignalblog/.eslintrc.js
```diff
@@ -6,6 +6,16 @@ module.exports = {
...
+      "eslint-import-resolver-custom-alias": {
+        alias: {
+          "#civicsignalblog": "./src",
+        },
+        extensions: [".js"],
+      },
+      typescript: {
+        alwaysTryTypes: false,
+        project: "./tsconfig.json",
+      },
```
kilemensi, line 18: Since these are JS files, can you add a comment describing why we need these changes e.g. `alwaysTryTypes: false`?

apps/civicsignalblog/.eslintrc.js
```diff
@@ -6,6 +6,16 @@ module.exports = {
       webpack: {
         config: "./eslint.webpack.config.js",
       },
+      "eslint-import-resolver-custom-alias": {
+        alias: {
+          "#civicsignalblog": "./src",
+        },
+        extensions: [".js"],
+      },
```
kilemensi, line 14: `eslint` should pick these up if defined in [`jsconfig`](https://github.com/CodeForAfrica/ui/blob/30aea2de7617e9270f212ddcbf88543eface4e51/packages/eslint-config-commons-ui/index.js#L56-L61), no?

apps/civicsignalblog/payload.config.ts
```diff
@@ -11,21 +9,19 @@ import seo from "@payloadcms/plugin-seo";
...
 import Posts from "./src/payload/collections/Research/Posts";
 import Publication from "./src/payload/globals/Publication";
 import Research from "./src/payload/globals/Site/research";
 import Main from "./src/payload/globals/Site/main";
 import Tags from "./src/payload/collections/Research/Tags";
 import Users from "./src/payload/collections/Users";
 import { defaultLocale, locales } from "./src/payload/utils/locales";
-
 import Actions from "./src/payload/components/actions";
+import path from "path";
```
kilemensi, line 24: node/external deps import should go before local imports, no?

apps/civicsignalblog/.eslintrc.js
m453h: Makes sense, should work 🚀

m453h:
@CodeForAfrica/tech is this PR a go or a no-go? 👀 🤞

koechkevin:
> @CodeForAfrica/tech is this PR a go or a no-go? 👀 🤞

_kaende_

kilemensi (approved):
LGTM

kilemensi:
@koechkevin 

Screenshot 2024-10-01 at 15 28 36

koechkevin (approved):

Issue #905, opened 2024-10-02
Implement parsing of Slate TextEditor content on CivicSignal webtools from Payload


m453h:
On the current state we have integrated the Media Page with Payload Backend however, the content coming from the [Slate Rich Text Editor](https://payloadcms.com/docs/rich-text/slate) is not parsed to generate proper HTML content. We should target an implementation that reuses most of what has been done on CodeforAfrica / RoboShield apps.![Image](https://github.com/user-attachments/assets/7fd27ae2-0f25-4fe4-b496-e5c99f64fbfc)

PR #931, opened 2024-10-03
@/CivicSignalBlog - Add Ability to Manage MediaData Page

m453h:
## Description

This PR adds the ability to Manage MediaData Page, the content managed here will be consumed by the MediaData Page proposed in this [PR](https://github.com/CodeForAfrica/web-tools/pull/20) 

Fixes:
- #905 
- #906 

## Type of change

## Screenshots
Screenshot 2024-10-03 at 14 34 07
Screenshot 2024-10-03 at 14 35 43

## Checklist:


apps/civicsignalblog/src/payload/access/canAccessApplication.js
```diff
@@ -1,7 +1,9 @@
 export default function canAccessApplication(user, searchString) {
+  // We are using this condition to only control what a user can see at a given time
   if (user) {
     const app = user.currentApp || user.defaultApp;
     return app === searchString.toLowerCase();
   }
-  return false;
+  // For APIs coming from external apps we are not going to use any restriction in reading
+  return true;
```
kilemensi, line 8: Can't we use a api key or something?

apps/civicsignalblog/src/payload/blocks/WebTools/PageHeader.js
```diff
@@ -0,0 +1,30 @@
...
+  },
+  imageURL: "/images/cms/blocks/web_tools_page_header.png",
+  imageAltText: "Header for web tools pages",
+  fields: [
+    {
+      name: "title",
+      required: true,
+      type: "text",
+    },
+    richText({
```
kilemensi, line 19: If `webtools` and `research` both have `title` and `description` for a page header, can't we make both work with the same block(s0?

apps/civicsignalblog/src/payload/blocks/WebTools/ResourceList.js
```diff
@@ -0,0 +1,24 @@
+const PageHeader = {
+  slug: "webtools-resource-list",
+  labels: {
+    singular: "Resource List",
+    plural: "Resource Lists",
+  },
+  imageURL: "/images/cms/blocks/resource_list.png",
+  imageAltText: "A list of resrouces to display",
```
kilemensi, line 8:
```suggestion
  imageAltText: "A list of resources to display",
```

kilemensi: Isn't it better to call this `MediaDataList`?

kilemensi (commented):
👍🏽

apps/civicsignalblog/src/payload/blocks/WebTools/ResourceList.js
m453h: @kilemensi My thinking was to find a generic name, since this could be used to choose if we want to display a list of items could be MediaDataList, or Application Features list etc. in other pages....but looking at the current approach which would use minimal update on the codebase, I think this block will be scrapped of...unless we still want to retain movable/dynamic blocks rendering on the web-tools frontend

apps/civicsignalblog/src/payload/blocks/WebTools/PageHeader.js
m453h: The only difference between the two is that on `research` the description is a normal text field but in `webtools` we are using `RichText` as the description

kilemensi: I noticed that:

1. Do we actually have a page header in [`research`](https://research.civicsignal.africa/)?
2. I was asking if it wouldn't make sense to make the description in research also supports `richtext` (since they're all part of the same app family)

m453h: We currently don't use it anywhere on `research` front-end, but looking at how the page header would be rendered  it might not be the best option to support `richtext` since its already pre formatted 👇 
Screenshot 2024-10-08 at 15 08 45
 
---
One Important question, on the `MediaData`, should we still  go with the dynamic rendering of blocks i.e. Payload fully controls what is being rendered on the front end or should we revert to the static structure and only parse whatever is coming from payload to the structure we already have

kilemensi: ... I'm assuming it's "**This is a subtitle**"? I think lets just give uses control over it. 👇🏽 still looks good.
Screenshot 2024-10-08 at 15 24 55

compared to

Screenshot 2024-10-08 at 15 26 08

apps/civicsignalblog/src/payload/access/canAccessApplication.js
m453h: We could use an API key, it might not add much value (since we'll be reading from purely client side) but it's going to be for `readOnly` users

apps/civicsignalblog/src/payload/blocks/WebTools/PageHeader.js
m453h: Yes that's the subtitlbe, let me add the support for richtext as well 👍

apps/civicsignalblog/src/payload/lib/data/common/applications.js
```diff
@@ -1,7 +1,7 @@
-export const MAIN = "main";
+export const MAIN = "tools";
 export const EXPLORER = "explorer";
-export const TOPIC_MAPPER = "topic-mapper";
-export const SOURCE_MAPPER = "source-manager";
+export const TOPIC_MAPPER = "topics";
+export const SOURCE_MAPPER = "sources";
```
kilemensi, line 4: I like the original names i.e. `topic-mapper`, `source-mapper` compared to `topics` and `sources`. Any reason why we're changing them?

kilemensi (commented):
👍🏽

apps/civicsignalblog/src/payload/access/canAccessApplication.js
```diff
@@ -1,7 +1,11 @@
+import { ROLE_APP } from "./roles";
+
 export default function canAccessApplication(user, searchString) {
+  // We are using this condition to only control what a user can see at a given time
   if (user) {
     const app = user.currentApp || user.defaultApp;
-    return app === searchString.toLowerCase();
+    // The assumption is that one account with ROLE_APP will be used to access all the app content
+    return app === searchString.toLowerCase() || user.roles.includes(ROLE_APP);
```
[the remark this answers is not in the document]
m453h: I was thinking in the long run we would want the `ROLE_APP` to only have full read access, but no update/create/delete privileges

apps/civicsignalblog/src/payload/lib/data/common/applications.js
m453h: I had to rename them  to match what is in the [webtools.](https://github.com/CodeForAfrica/web-tools/blob/main/src/config.js)

That way, I can easily/consistently make API request for content tied with an app name using the `AppName-collection` naming scheme like how I did it [here](https://github.com/CodeForAfrica/web-tools/blob/2e4983a36b9937d2dfab859e2d94f637c5d2d92c/src/components/tools/media/Homepage.js)

kilemensi: Good find... all good.

apps/civicsignalblog/src/payload/access/canAccessApplication.js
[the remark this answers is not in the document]
m453h: Yes, we could introduce this role; however, the limitation with the current implementation is that `canAccessApplication` only restricts what the user can view based on the application they have selected—specifically by using the `currentApp` field in their profile.  i.e. each user we create has access to all apps, `canAccessApplication` controls only what collections they can see based on the app they have selected.

We could change the `currentApp` to support the selection of multiple apps so that we can accommodate what the subscriber (API) can access.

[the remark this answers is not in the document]
m453h: I'm worried that this approach wouldn't work well if we want one Payload user to access content across all the apps, I'm not sure how well payload would handle concurrent requests from different apps (my concern is that this could result into one app being denied access to the content)...I think its best if we limit  the number of times we update the `currentApp` field  i.e. just do it on the backend during account creation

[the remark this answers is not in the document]
kilemensi: Not a bad thing to worry about but if it is that _fragile_ then the `currentApp` approach is probably not the best option for what we need.

Essentially, we should be able to make a request: I want this data from this app. These requests should be self contained i.e. one requests is not affected by the previous/next.

Authentication should be able to tell us yes/no, this use is allowed/not allowed to read/write this data to this part of the app.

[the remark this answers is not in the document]
m453h: I think the best approach would be to Add `allowed apps` field that will track what apps users can access.

We will distinguish API and normal backend UI permission requests with the header we'll be sending. i.e. With app requests we will only check if the user can access the app they want content from, with backend UI we will put an additional check on the `currentApp` selected by the user. 

This way we wont need to make any updates on the `currentApp` per each request we'll be sending via the API.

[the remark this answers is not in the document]
kilemensi: ... it's a solid approach but isn't this (the `currentApp` or `app`, etc.) the kind of thing that should always be submitted per request?

1. APIs: Header, etc.
2. Browser/Backend UI: (session/secure) cookie. You set it in the browser and the browser submit it with every request.

I'm just hesitant on approaches that require two different ways of doing basically the same thing.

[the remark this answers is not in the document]
m453h: Cool Let me explore the best way we can set session values in Payload for the browser/backend UI, I think there should be a clean way to do it...this would remove the `currentApp` field on the users collection and require to move a few parts on the app selection component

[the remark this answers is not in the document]
m453h: @kilemensi, Can you have another look, I decided to keep the `currentApp` field, as I couldn't safely set session variables for the browser/backend UI, the next best thing was to try out [user preferences](https://payloadcms.com/docs/admin/preferences) but getting the values from the access control functions would need us to query the database (which adds more complexity, while as opposed to using the `user` object which is automatically provided from `req`), 

I think with these modifications we can have a more consistent flow for permission checks and would be the same regardless of where we are getting the `currentApp` from.

I have made some assumptions on the permissions for the collections i.e. _ROLE_EDITOR_, _ROLE_ADMIN_ are the only roles that can _update_, _create_, _delete_ collections as opposed to having them open to any user.

apps/civicsignalblog/payload.config.ts
```diff
@@ -37,6 +39,8 @@ const cors =
     ?.map((d) => d.trim())
     ?.filter(Boolean) ?? [];
 
+const customHeaders: string[] = ["x-current-app"];
```
kilemensi, line 42:
```suggestion
const customHeaders: string[] = ["CS-App"];
```

1. `X-` is already [deprecated](https://www.rfc-editor.org/rfc/rfc6648) so lets just use `Cs-` for CivicSignal or something along those lines
2. `current` is cool for long session i.e. backend UI, but for a header that can change on every request, I don't think it adds any value.

apps/civicsignalblog/payload.config.ts
```diff
@@ -153,4 +158,18 @@ export default buildConfig({
     }),
   ] as any[],
   telemetry: process?.env?.NODE_ENV !== "production",
+  express: {
+    postMiddleware: [
+      (_req: Request, res: Response, next: NextFunction) => {
```
kilemensi, line 163: Add a comment above this code explaining why we have to do this.

apps/civicsignalblog/src/components/PageHeader/PageHeader.js
```diff
@@ -31,9 +32,14 @@ const PageHeader = React.forwardRef(function PageHeader(props, ref) {
         >
           {title}
         </RichTypography>
-        <RichTypography component="h2" variant="h2">
-          {subtitle}
-        </RichTypography>
+        <RichText
+          ref={ref}
```
kilemensi, line 36: Why are we putting the component `ref` on this `RichText`? Shouldn't it be on parent/root component?

apps/civicsignalblog/src/payload/access/canAccessApplication.js
```diff
@@ -1,7 +1,10 @@
-export default function canAccessApplication(user, searchString) {
+export default async function canAccessApplication(req, searchString) {
+  const { user, headers } = req;
+
   if (user) {
-    const app = user.currentApp || user.defaultApp;
-    return app === searchString.toLowerCase();
+    const currentApp = headers["x-current-app"] || user.currentApp;
+    const app = currentApp || user.defaultApp;
```
kilemensi, line 6:
```suggestion
    const app = headers["CS-App"] || user.currentApp || user.defaultApp;
```

One liner is still very legible, no?

The only thing I'm still not 100% happy here is APIs **MUST** provide `CS-App` header. You don't want a situation where we end up showing wrong data in the FE.

apps/civicsignalblog/src/payload/collections/Main/MediaData.js
```diff
@@ -0,0 +1,54 @@
...
+  fields: [
+    {
+      name: "title",
+      type: "text",
+      required: true,
+      localized: true,
+    },
+    image({
+      overrides: {
+        name: "mediaDataImage",
```
kilemensi, line 32: Why not just `image`? Collection is already named `mediaData`... may be it should be `thumbnail`? At least that's what I think we're using it for.

apps/civicsignalblog/src/payload/collections/Main/MediaData.js
```diff
@@ -0,0 +1,54 @@
...
+        required: true,
+        localized: true,
+      },
+    }),
+    document({
+      overrides: {
+        name: "mediaDataDocument",
```
kilemensi, line 39:
```suggestion
        name: "document",
```

I also think it should be: `title`, `description` and then `image`/`thumbnail`. `document` can either go first (i.e. before `title`) as it's the most important part or last as it'll just be a link on FE.

kilemensi (commented):
Almost there... 👍🏽

apps/civicsignalblog/src/payload/collections/Users.js
```diff
@@ -73,6 +74,11 @@ const Users = {
         isClearable: true,
         isSortable: true,
       },
+      access: {
+        read: isAdminOrSelfFieldLevel,
+        create: isAdminFieldLevel,
+        update: isAdminFieldLevel,
```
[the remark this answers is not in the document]
m453h: Yes, we can allow them to change the `defaultApp` and `currentApp`. First, let me ensure that we can restrict the available app options to only those they have permission to access.

apps/civicsignalblog/src/payload/collections/Main/MediaData.js
m453h: We can call it thumbnail 👍

apps/civicsignalblog/src/components/PageHeader/PageHeader.js
m453h: Cant even recall when I kept this 🫠 ...removed ✅

apps/civicsignalblog/src/components/PageHeader/PageHeader.js
```diff
@@ -31,9 +32,13 @@ const PageHeader = React.forwardRef(function PageHeader(props, ref) {
         >
           {title}
         </RichTypography>
-        <RichTypography component="h2" variant="h2">
-          {subtitle}
-        </RichTypography>
+        <RichText
+          typographyProps={{
```
kilemensi, line 36: Not sure how we missed this but lets add an issue to track this, `typographyProps` should be `TypographyProps`

kilemensi (approved):
Code LGTM!

koechkevin (approved):

PR #941, opened 2024-10-09
@/Climatemappedafrica Hero Section

### LINKED ISSUE #914: ClimateMapped Hero Block
Create a hero block for ClimateMapped. Similar to PesaYetu's hero block.

Request Link: [Feature 1](https://docs.google.com/document/d/1TUrdCRnbj6TSqiuwITKUpnDUEi_ktm_3PmgsaD0y2bw/edit#heading=h.7grsuhlop4fo)

> Preview:

Image

> Block Fields

Image

koechkevin:
## Description

This PR:

-  Adds Page Hero Block Section to Payload cms.
- Changes Header Subtitle to use RichText Component
- Updates the Hero Map to default from Africa instead of Kenya.

Fixes #914 

## Type of change

## Screenshots
image
image

## Checklist:


apps/climatemappedafrica/src/components/Header/index.stories.js
```diff
@@ -0,0 +1,20 @@
...
+  return <Header {...args}>{title}</Header>;
+}
+
+export const Default = Template.bind({});
+
+Default.args = {
+  overline: "Overline",
+  title: 'Title with <span class="highlight">Highlight</span>',
+  subtitle: "A short paragraph to describe what the header is all about.",
+};
```
kilemensi, line 20:
```suggestion
import React from "react";

import Header from ".";

const meta = {
  title: "PesaYetu/Components/Header",
};

function Template({ title, ...args }) {
  return <Header {...args}>{title}</Header>;
}

const Default = Template.bind({});

Default.args = {
  overline: "Overline",
  title: 'Title with <span class="highlight">Highlight</span>',
  subtitle: "A short paragraph to describe what the header is all about.",
};

export { Default };
export default meta;
```

Prefer fixing linting errors whenever possible instead of disabling them.

apps/climatemappedafrica/src/lib/data/blockify/hero.js
```diff
@@ -0,0 +1,20 @@
+import {
+  fetchProfile,
+  fetchProfileGeography,
+} from "@/climatemappedafrica/lib/hurumap";
+
+export default async function hero(block) {
+  const { geometries } = await fetchProfileGeography("af");
```
kilemensi, line 7: Minor but `"af"` can be set in the CMS, no?

apps/climatemappedafrica/src/payload/blocks/PageHero.js
```diff
@@ -0,0 +1,44 @@
+import richText from "../fields/richText";
+
+const PageHero = {
+  slug: "page-hero",
+  imageURL: "/images/cms/blocks/hero.png",
+  imageAltText: "Used in homepage",
+  labels: {
+    singular: "Site Hero",
+    plural: "Site Hero",
+  },
```
kilemensi, line 10: Looking at the labels, shouldn't the component be named SiteHero? I'm assuming the implication is the whole site will have just one of these.

apps/climatemappedafrica/src/payload/blocks/PageHero.js
```diff
@@ -0,0 +1,44 @@
...
+    {
+      name: "searchPlaceholder",
+      type: "text",
+      label: "Search Placeholder",
+    },
+    {
+      name: "comment",
+      type: "text",
+      label: "Comment",
+      required: true,
```
kilemensi, line 39: Comment should probably be optional as well, no?

kilemensi (commented):
Haven't run it locally but code looks alright to me

koechkevin:
> Hi @koechkevin I'm struggling to get the frontend running locally I'm getting a 404 even after creating the pages...not sure what I could be missing 🤔
> 
> Just an additional thing, if there isn't any other PR that modifies the `.env.template` it'll surely be a good time to include the `PAYLOAD_SECRET` and `MONGO_URL` in the template.

There's a PR that updates `env.template` @m453h Not sure why there's a 404 on your side though. Could be page slugs? or you still have the permanent redirects?

kelvinkipruto (commented):
Look ok @koechkevin, however it looks like the map on the hero is not interactive like the one in Pesayetu.

apps/climatemappedafrica/src/lib/data/blockify/hero.js
koechkevin: @kilemensi Having this in the CMS opens a can of worms, which may not be minor. For instance:

 - What are the options we will allow users to choose from? countries or continents.
 - If continents, then `af` is the only option and we have the centre co-ordinates.
 - If countries, then we also need to set center co-ordinates so that map shows accurately. Can we guarantee that all regions in any country are implemented?
 
 This is only the hero page and would be best to just default to `af` then `countries` become clickable IMO.

kilemensi: My understanding is we only need three things @koechkevin:

1. `geography`: We can pull this from HURUmap API and show it as a `Select` field. This way, this information only needs to be entered once (in the backend admin).
2. `centre`: We can use `Point` field (or two `Number` fields) for `lat` and `long`.

>Can we guarantee that all regions in any country are implemented?
>
Isn't the data team handling the validity/correctness of the data?

---

My thinking is a single `Hero` component, should work in any HURUmap based app and hence it should have preferably zero hard-coded props/settings; backend should drive what is shown in it.

kelvinkipruto (approved):

apps/climatemappedafrica/src/components/Header/Header.js
```diff
@@ -3,6 +3,8 @@ import { Box } from "@mui/material";
 import PropTypes from "prop-types";
 import React from "react";
 
+import RichText from "@/climatemappedafrica/components/RichText";
```
kilemensi, line 6: Hasn't this `RichText` component been moved to `@commons-ui/payload`?

apps/climatemappedafrica/src/components/Header/Header.js
```diff
@@ -50,16 +52,26 @@ const Header = React.forwardRef(function Header(props, ref) {
...
-        {...SubtitleProps}
-        sx={{
-          mt: 20,
-          ...SubtitleProps?.sx,
-        }}
-      >
-        {subtitle}
-      </RichTypography>
+      {typeof subtitle === "string" ? (
+        <RichTypography
```
kilemensi, line 56: Cool but lets leave this header as is (just string)... I've created a [`RichHeader`](https://github.com/CodeForAfrica/ui/pull/952/files#diff-ab220b51a3db21b1094cd4fbc66d9b4b821bdb22fe9059db62bcb38881c21228) component that handles rich text.

apps/climatemappedafrica/src/components/Header/index.stories.js
```diff
@@ -0,0 +1,19 @@
+import React from "react";
```
kilemensi, line 1: We don't need stories... tests are enough for apps.

apps/climatemappedafrica/src/lib/data/blockify/hero.js
```diff
@@ -0,0 +1,32 @@
...
+  };
+  const childLevel = childLevelMaps[level];
+  const { locations } = await fetchProfile();
+  const featuredLocations = locations.filter(
+    (location) => location.level === childLevel,
+  );
+
+  const { children } = geometries;
+  const boundary =
+    children[level === "country" ? "County" : childLevel] ?? null;
```
kilemensi, line 23: Why this mismatch of `country` and `County`?

apps/climatemappedafrica/src/pages/[[...slugs]].js
```diff
@@ -2,43 +2,43 @@ import React from "react";
...
-function Index({ blocks, fallback, ...props }) {
+export default function Index({ blocks, fallback }) {
   if (!blocks?.length) {
     return null;
   }
 
-  let PageConfig = React.Fragment;
-  let pageConfigProps;
+  let PageComponent = React.Fragment;
+  let pageComponentProps;
```
Comment (kilemensi) [inline apps/climatemappedafrica/src/pages/[[...slugs]].js:22]: Why are we changing this component/file?

apps/climatemappedafrica/src/pages/api/hurumap/profiles.js
```diff
@@ -0,0 +1,15 @@
...
+
+export default async function index(req, res) {
+  if (req.method === "GET") {
+    try {
+      const { geoCode } = req.query;
+      const result = await fetchProfile(geoCode);
+      return res.status(200).json(result);
+    } catch (err) {
+      return res.status(500).json(err.message);
+    }
```
kilemensi, line 11: 🤔 ... shouldn't `404` be thrown when we have invalid `geoCode`? `500` means server error (which can happen but not always).

kilemensi (commented):
👍🏽 

---

Lets ensure the components follow the standard `Folder/{index.js},{Folder.js},{Folder.test.js}`

apps/climatemappedafrica/src/lib/data/blockify/hero.js
koechkevin: @kilemensi because this is how children data looks like when the selected level is Country
```
{ County: { type: 'FeatureCollection', features: [] } }
```
at least for Kenya

kilemensi: Other countries don't have counties @koechkevin

apps/climatemappedafrica/media/pesayetu_logo.svg
kelvinkipruto, line 1: @koechkevinWe do not need Pesayetu's logo here

apps/climatemappedafrica/src/components/Hero/Map.js
```diff
@@ -0,0 +1,134 @@
...
+      });
+      layer.on("mouseout", () => {
+        setHoverGeo(null);
+        layer.setStyle({
+          fillOpacity: 1,
+          fillColor: theme.palette.background.default,
+        });
+      });
+      layer.on("click", () => {
+        router.push(`/explore/${feature.properties.code.toLowerCase()}`);
```
kelvinkipruto, line 83: This should not be hardcoded. It will be fixed with https://github.com/CodeForAfrica/ui/pull/945. We can add an issue to track it later

apps/climatemappedafrica/src/components/RichText/RichText.js
kelvinkipruto, line 1: We already have a shared richtext from [`@commons-ui/payload`](https://github.com/CodeForAfrica/ui/blob/main/packages/commons-ui-payload/src/RichText/RichText.js)

kelvinkipruto (commented):
Looks good @koechkevin. However one thing I noticed is when adding the block, it takes some time dur to the API request made, and a user can click it multiple times. Perhaps we can cache the response to make it faster

apps/climatemappedafrica/src/lib/data/blockify/hero.js
```diff
@@ -0,0 +1,34 @@
...
+
+export default async function hero(block) {
+  const { geometries } = await fetchProfileGeography(
+    block.location?.name?.toLowerCase() ?? "af",
+  );
+  const { level } = geometries.boundary?.properties ?? {};
+  const childLevelMaps = {
+    continent: "country",
+    country: "region",
+  };
```
kilemensi, line 14: Just thinking out loud, doesn't the geographies/geometries we get from HURUmap define hierarchy we can use rather than manually computing it.

apps/climatemappedafrica/src/pages/[[...slugs]].js
```diff
@@ -2,16 +2,18 @@ import React from "react";
...
 import { getPageServerSideProps } from "@/climatemappedafrica/lib/data";
 
 const componentsBySlugs = {
+  hero: Hero,
   summary: Summary,
   team: AboutTeam,
 };
 
-function Index({ blocks, fallback, ...props }) {
+export default function Index({ blocks, fallback }) {
```
Comment (kilemensi) [inline apps/climatemappedafrica/src/pages/[[...slugs]].js:16]: Why?

apps/climatemappedafrica/src/pages/[[...slugs]].js
```diff
@@ -23,7 +25,7 @@ function Index({ blocks, fallback, ...props }) {
     pageConfigProps = { value: { fallback } };
   }
   return (
-    <Page {...props}>
+    <Page {...pageConfigProps}>
       <PageConfig {...pageConfigProps}>
```
Comment (kilemensi) [inline apps/climatemappedafrica/src/pages/[[...slugs]].js:29]: `pageConfigProps` sounds like something that should go into `PageConfig` rather than `Page`. Are we naming them wrong?

kilemensi (commented):
👍🏽

apps/climatemappedafrica/src/lib/data/blockify/hero.js
koechkevin: @kilemensi for KE this is the object response from hurumap
image
Any other country, it uses the Region key, see below (TZ, DJ, etc)
image
for continent,
image
hence the decision

kelvinkipruto: @koechkevin I agree with @kilemensi that we do not need to compute it manually, other countries may have `Province`, for example, Angola. We had a similar problem with @thepsalmist, and we resolved it by adding every possible country-level subdivision to the `preferred_children` array in the profile config on HURUMap, then access it as follows: 

[`const preferredChildrenPerLevel = preferredChildren[level];`](https://github.com/CodeForAfrica/ui/blob/62f1e858338f21e6643a8a8e60fc0eeade2a5ea8/packages/hurumap-next/src/Map/LazyMap.js#L41). 

We can have a call to clarify it if need be

apps/climatemappedafrica/src/components/Hero/Hero.js
```diff
@@ -0,0 +1,186 @@
...
+  featuredLocations,
+  searchPlaceholder,
+  properties,
+  location: { center },
+  level,
+  ...props
+}) {
+  const theme = useTheme();
+  const { typography, breakpoints } = theme;
+  const isUpLg = useMediaQuery(theme.breakpoints.up("lg"));
```
kilemensi, line 30: As we've already injected `theme` via `ThemeProvider`, we can just:
```suggestion
  const isUpLg = useMediaQuery((theme) => theme.breakpoints.up("lg"));
```

apps/climatemappedafrica/src/components/Hero/Hero.js
```diff
@@ -0,0 +1,186 @@
...
+        sx={{
+          position: "absolute",
+          zIndex: -1,
+          height: typography.pxToRem(468),
+          width: "100%",
+          [breakpoints.up("md")]: {
+            height: typography.pxToRem(456),
+          },
+          [breakpoints.up("lg")]: {
+            height: typography.pxToRem(600),
```
kilemensi, line 54: `sx` can handle breakpoints without the explicit use of `theme.breakpoints`
```suggestion
          width: "100%",
          height: { xs: "468px", md: "456px", lg: "600px" },
```

apps/climatemappedafrica/src/components/Hero/Hero.js
```diff
@@ -0,0 +1,186 @@
...
+          <Box
+            sx={{
+              display: {
+                xs: "none",
+                md: "block",
+              },
+            }}
+          >
+            <Grid item lg={1} />
+          </Box>
```
kilemensi, line 81: Why not just
```suggestion
          <Grid item lg={1} sx={{ display: {  xs: ....} }}  />
```
? Everything inside `Grid container` must be `Grid item ...`

apps/climatemappedafrica/src/components/Hero/Map.js
```diff
@@ -0,0 +1,134 @@
...
+    marginTop: typography.pxToRem(55),
+    [breakpoints.up("lg")]: {
+      height: typography.pxToRem(471),
+      marginTop: typography.pxToRem(42),
+      width: typography.pxToRem(371),
+    },
+    "& .leaflet-container": {
+      background: "transparent",
+    },
+  },
```
kilemensi, line 24: Doesn't this work as `sx` on component?

apps/climatemappedafrica/src/components/Hero/Map.js
```diff
@@ -0,0 +1,134 @@
...
+  tooltip: {
```
kilemensi, line 25: Is this used anywhere?

apps/climatemappedafrica/src/payload/blocks/Hero.js
```diff
@@ -0,0 +1,67 @@
...
+    richText({
+      name: "subtitle",
+      required: true,
+      label: "Sub Title",
+    }),
+    {
+      name: "searchLabel",
+      type: "text",
+      label: "Search Label",
+      required: true,
```
kilemensi, line 27: Aren't these fields missing `localized: true`?

apps/climatemappedafrica/src/payload/fields/LocationSelect.js
```diff
@@ -0,0 +1,43 @@
...
+}
+
+function LocationSelect(props) {
+  const url = `${apiUrl}/api/hurumap/profiles`;
+  const { data } = useSWR(url, fetcher, {
+    dedupingInterval: 60000,
+    revalidateOnFocus: false,
+  });
+  const { locations } = data ?? {};
+  const memoOptions = () =>
```
kilemensi, line 30:
```suggestion
  const locationOptions = () =>
```

apps/climatemappedafrica/src/payload/fields/LocationSelect.js
```diff
@@ -0,0 +1,43 @@
...
+    (
+      locations
+        ?.filter(({ level }) => level !== "region")
```
kilemensi, line 33: Does this work across counties? i.e. won't we end up with counties or provinces as valid options?

kilemensi (commented):
👍🏽

apps/climatemappedafrica/src/payload/fields/LocationSelect.js
koechkevin: Yes @kilemensi all locations under country have `level: region` key value pair  whether province county etc.

apps/climatemappedafrica/src/pages/api/hurumap/profiles.js
```diff
@@ -0,0 +1,25 @@
+import { fetchProfile } from "@/climatemappedafrica/lib/hurumap";
+
+let cache = null;
+let cacheExpiry = 0;
```
kilemensi, line 4: Not sure that this kind of cache actually works/helps

apps/climatemappedafrica/src/payload/blocks/Hero.js
```diff
@@ -0,0 +1,73 @@
...
+    richText({
+      name: "title",
+      required: true,
+      label: "Title",
+      localized: true,
+    }),
+    richText({
+      name: "subtitle",
+      required: true,
+      label: "Sub Title",
```
kilemensi, line 22:
```suggestion
      label: "Subtitle",
```

Or just remove `label` if it's the same as `name`.

kilemensi (approved):
Code LGTM!

apps/climatemappedafrica/src/components/LineClampedRichTypography/LineClampedRichTypography.js
```diff
@@ -0,0 +1,54 @@
+import { RichTypography } from "@commons-ui/core";
```
kilemensi, line 1: Forgot to ask, shouldn't this be `@commons-ui/next`? I believe `climatemappedafrica` (and `pesayetu`) are still using the old @commons-ui/core. Not sure if that one had `RichTypography` and if it did, not sure if it works like the current `RichTypography`.

```suggestion
import { RichTypography } from "@commons-ui/next";
```

PR #945, opened 2024-10-11
ClimateMapped Navigation

### LINKED ISSUE #915: ClimateMapped Navigation
Create a navigation bar for ClimateMapped similar to Pesayetu's

Request Link: [Feature 1](https://docs.google.com/document/d/1TUrdCRnbj6TSqiuwITKUpnDUEi_ktm_3PmgsaD0y2bw/edit#heading=h.7grsuhlop4fo)

> Preview
Image
Image

> Structure
Image

kelvinkipruto:
## Description

This PR setups the main navigation in ClimateMapped.
It also introduces `HURUMap` global settings to manage any HURUMap-related configuration

Fixes #915

## Type of change

## Screenshots
image

image

image

## Checklist:


kilemensi:
>It also introduces page variants, to differentiate between normal and /explore pages
>

Can you expound on this?

kelvinkipruto:
> Can you expound on this @kelvinkipruto?

@kilemensi There are two different types of pages in Pesayetu, the default pages like `/home`, `/about`, etc, and the `explore` page. They have different features, for example, the navbar is different for each page type, and the `explore` page does not have a footer while the rest have.

kilemensi:
Cool @kelvinkipruto 

1. Are we going to have explore pages in the CMS? My understanding is all content for these pages comes from the Django backend... adding them to Payload feels like duplication.
2. Weren't there a talk of adding footer to explore pages as well?

kelvinkipruto:
@kilemensi We're going to have just the main `explore` page in the CMS to configure components like `panels`, and `tutorial`

kilemensi:
🤔 @kelvinkipruto 

1. Sounds like we only need one page to be different so a `variant` field on **all** pages sounds like an over kill.
2. Tutorials, etc., do not change from page to page so they sound like something that should be site-wide configurable rather than at page level i.e. may be we should have settings -> HURUmap, or something.

Similarly, may be in these HURUmap settings (or any other global), we can select which page should be treated as explore page. This ensures we only have one page per site.

It's possible to also have a HURUmap block that when added to a page, we treat that page as `explore` in the front end... but this approach mean there could be more than one explore page in the same site (similar issue with `variant` approach but at least it doesn't impact any other page with additional fields).

kelvinkipruto:
@kilemensi I think the global HURUMap settings is the best approach. I have updated the PR to take that into account

kilemensi:
Can we update this to main @kelvinkipruto? I think we'd need to test the explore page functionality, no?

kelvinkipruto:
@koechkevin @kilemensi This is also ready for review

apps/climatemappedafrica/src/lib/data/common/index.js
```diff
@@ -33,6 +33,24 @@ function getFooter(siteSettings) {
   };
 }
 
+function getMenus(siteSettings) {
```
kilemensi, line 36: This should be `getNavigation` or `getNavbar` or something along those lines... menus are just part of it. What do we call it in **charterafrica** or **codeforafrica**?

apps/climatemappedafrica/src/lib/data/common/index.js
```diff
@@ -33,6 +33,24 @@ function getFooter(siteSettings) {
...
+  const {
+    connect: { links = [] },
+    primaryNavigation: { menus = [], connect = [] },
+    primaryLogo,
+    drawerLogo,
```
kilemensi, line 41: drawerLogo? Not sure why this is from the CMS. Is it something customisable? Otherwise this is just another icon (part of the theme).

apps/climatemappedafrica/src/lib/data/common/index.js
```diff
@@ -49,13 +67,21 @@ export async function getPageProps(api, context) {
...
 
   const siteSettings = await api.findGlobal("settings-site");
   const footer = getFooter(siteSettings);
+  const menus = getMenus(siteSettings);
 
   return {
     blocks,
     footer,
+    menus,
+    variant: page.slug === explorePage.slug ? "explore" : "default",
```
kilemensi, line 85: I'm still not sure why we need variant on page. My understanding is:

If this page is explore page (or child of explore page):

1. Set variant on the Navigation (should show location search)
2. Set variant on the Footer (current we shouldn't display but if we're switching to always showing a footer, we won't need variant on footer)
3. Load HURUmap data and convert that into hurumap block that we can use to render the data exploration components on the FE.

What am I missing?

kilemensi (commented):
👍🏽 

--

I expect Navigation and Footer to show on any page even without any blocks defined. Homepage still renders black in this PR.

koechkevin (approved):

apps/climatemappedafrica/src/lib/data/common/index.js
kelvinkipruto: @kilemensi Yes. We need a different logo because the background of the drawer is blue, and the main logo would not look well
image
Sakina provided a new one that works very well as shown in the screenshots above

koechkevin:
@kelvinkipruto this should be go?

m453h (approved):

apps/climatemappedafrica/src/components/Footer/index.js
```diff
@@ -1,101 +1,115 @@
...
-        }}
-      >
-        <Grid container direction="row" justifyContent="space-between">
-          <Grid item xs={12} container>
-            {logoProps && (
-              <LogoButton
-                {...logoProps}
-                component="a"
+    !(variant === "explore" && isDesktop) && (
+      <div className={classes.root}>
```
kilemensi, line 28: Lets use `sx` more...
```suggestion
    <Box sx={{ display: { xs: "none", lg: variant !== "explore" ? "flex" : "none" } }}>
```

Instead of
```jsx
  const theme = useTheme();

  const isDesktop = useMediaQuery(theme.breakpoints.up("lg"));
  !(variant === "explore" && isDesktop) && (
```

apps/climatemappedafrica/src/lib/data/common/index.js
kilemensi: Damn, that's one logo too many (not your fault)... lets add a story to track it but it shouldn't block this PR

apps/climatemappedafrica/src/components/Footer/index.js
```diff
@@ -1,101 +1,115 @@
...
-                      textDecoration: "underline",
-                    },
-                  },
+            </Grid>
+            <Grid item xs={12} lg={4}>
+              <Grid
+                container
+                classes={{
+                  root: classes.allLinks,
                 }}
```
kilemensi, line 76: Shouldn't need `classes` here.

apps/climatemappedafrica/src/lib/data/common/index.js
```diff
@@ -4,7 +4,7 @@ export function imageFromMedia(alt, url) {
   return { alt, src: url };
 }
 
-function getFooter(siteSettings) {
+function getFooter(siteSettings, variant) {
```
kilemensi, line 7: Minor but taking a `page` is probably better i.e. settings and page in question.
```suggestion
function getFooter(siteSettings, page) {
```

Or `function getFooter(siteSettings, { variant })`;

apps/climatemappedafrica/src/pages/[[...slug]].js
```diff
@@ -11,9 +13,24 @@ const componentsBySlugs = {
   team: AboutTeam,
 };
 
-function Index({ blocks, fallback, ...props }) {
-  if (!blocks?.length) {
-    return null;
+function Index({ blocks, menus, footer: footerProps, seo = {}, fallback }) {
+  const pageSeo = {};
+  pageSeo.title = seo?.title || undefined;
```
Comment (kilemensi) [inline apps/climatemappedafrica/src/pages/[[...slug]].js:18]:
```suggestion
  pageSeo.title = seo?.title || null;
```

`seo?.title` is already `undefined` if `seo` is `undefined`. Why would we need the `extra || undefined`?

apps/climatemappedafrica/src/components/DropdownSearch/index.js
```diff
@@ -112,16 +112,34 @@ function DropdownSearch({
...
+          component={SearchIcon}
+          viewBox="0 0 48 48"
+          sx={{
+            width: 48,
+            height: 48,
+            borderRadius: "50%",
+            border: "2px solid #fff",
+          }}
+        />
+      )}
```
kilemensi, line 142: Without any duplicates:

```js
  let iconComponent = SearchIcon;
  let iconBorder = undefined;
  if (variant === explore) {
    iconComponent = IconProp;
    iconBorder = {
      borderRadius: "50%",
      border: "2px solid #fff",
    };
  }
```
Then
```suggestion
      <SvgIcon
        component={iconComponent}
        viewBox="0 0 48 48"
        sx={{
          width: 48,
          height: 48,
          ...iconBorder,
        }}
      />
```

apps/climatemappedafrica/src/lib/data/common/index.js
kelvinkipruto: Tracking in: #961

apps/climatemappedafrica/src/lib/data/common/index.js
kelvinkipruto: @kilemensi This won't work as we're fetching the variant from `globals` not the page itself.

PR #947, opened 2024-10-11
@/civicsignalblog - fix: Implement payload migrations to rename existing pages to research-pages collections

m453h:
## Description
The `civicsignalblog` Payload CMS has been updated to support managing content for multiple applications. As part of this change, the generic pages collection has been replaced by application-specific collections, where each follows a naming convention of `<APPLICATION_NAME>-pages.` Consequently, the original `pages` collection, which used the [slug](https://payloadcms.com/docs/configuration/collections#options) of `pages`, is no longer visible in the CMS and is not rendered on the frontend.

This PR fixes this problem by adding database [migrations](https://payloadcms.com/docs/database/migrations) to rename existing Pages in the database with the slug `pages` to  pages with a slug `research-pages`.

Since we no longer have a collection with `pages` slug we can no longer use `payload`'s internal API to query for the `pages` and migrate them to `research-pages`, thus we have resorted to using a direct connection to the database through the `mongoose-adapter`. 

The migration flow is as follows:
- Find all collections with the name `pages`
- For each `pages` collection found insert it to the `research-pages`, if a collection with the same slug exists in the `research-pages` log an error.
- Repeat the same procedure for migrating `_pages_versions` to  `_research-pages_versions`

This approach ensures that no data loss occurs when we run the migration since we are only copying data from one collection to another.

The following additional fixes have been added to ensure everything works accordingly:
  * Updated `package.json` to include new migration scripts for Payload CMS. (`apps/civicsignalblog/package.json`)
  * Modified `payload.config.ts` to use `applicationPages` for collections in SEO and nestedDocs plugins. 

Fixes # (issue)

## Type of change

## Checklist:


apps/civicsignalblog/migrations/20241011_104214_migration.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  // We will need the admin database for rename operation see https://www.mongodb.com/docs/manual/reference/command/renameCollection/
+  const adminDb = mongoose.connection.getClient().db("admin");
+
+  const oldCollectionVersionName = `_${oldCollectionName}_versions`;
+  const newCollectionVersionName = `_${newCollectionName}_versions`;
+
+  // Drop the new collection, since it has nothing and we cant rename the old collection to new name when this exists
+  await db.command({ drop: newCollectionName });
+  payload.logger.info(`Dropped the existing ${newCollectionName} collection.`);
```
kilemensi, line 21: Can we do this **without** dropping _any_ db? May be warn and exit if new collection already exists?

kilemensi (commented):
👍🏽 

---

Not sure if this was discussed before but shouldn't all collections be renamed to `app-collection`? e.g. `posts`.

apps/civicsignalblog/migrations/20241011_104214_migration.ts
m453h: mmh there's no clean solution that I can quickly think of...we'll always have to drop a collection from the db since we are trying to avoid name collision when we do the rename operation.

If we warn and exit it would not solve our problem (currently we already have `research-pages` which have no data, and we also have `pages` which have data and cant be viewed in payload cms/front-end, thus we don't really need `research-pages` and we can safely drop them.

Alternatively, we could rename the collections to something else like `_{$newCollectionName}` if we really want to avoid dropping collections

m453h:
> 👍🏽
> 
> Not sure if this was discussed before but shouldn't all collections be renamed to `app-collection`? e.g. `posts`.

I believe we discussed this before, and if my memory serves me right, it was revolving around using the naming convention for collections that would have the same slugs, which at the time only applied to `pages`. That's why posts wasn't renamed.

apps/civicsignalblog/migrations/20241011_104214_migration.ts
kilemensi: Yes it won't **solve** the problem but it will tell us that _something_ went wrong. Basically, when it comes to managing data, you don't want automatic data deletion anywhere in the process, **especially** in `PROD`. Someone must 👁️ and give the 👍🏽 to delete (preferably after a careful and thorough review).

... but to make sure we're all on the same page, First time migrations are run:
1. the script will **find** old collection **but not** new collection (no need to delete).
2. If successful, at the end of migration, old collection would have been renamed to new collection.

If someone runs the migration again:
1. the script **will not find** old collection **but** there will be new collection.
2. the script will delete this new collection.
3. since there is no old collection, there is nothing to rename.


I'll sleep better an night if we issue warning and exit in ☝🏽 situation without deleting anything.

kilemensi:
> I believe we discussed this before, and if my memory serves me right, it was revolving around using the naming convention for collections that would have the same slugs, which at the time only applied to `pages`. That's why posts wasn't renamed.
>

Sounds like a good short-term solution but not sure how it will play long term i.e. when all apps have been migrated. All new collections will have `app-collection` naming but `posts` will stick out and someone new will have to have institutional knowledge above.

Cool, if we can't do it now, lets create an issue to track this and implement it when time permits.

apps/civicsignalblog/migrations/20241011_104214_migration.ts
m453h: It is slightly different:

For the first time the migration is run:
1. The script will find both the old and new collection, the reason being that Payload has already automatically created a new collection after it found the new slug i.e. `research-pages` but didn't delete the old collection which has `pages`.

1. So running the rename the operation without deleting the new-collection will always fail due to name collision.

Additionally, Payload keeps a [record](https://payloadcms.com/docs/database/migrations#status) of executed migrations and will not execute them when someone re-runs the migrations unless they first intentionally rollback the execution status

Screenshot 2024-10-15 at 14 59 37

I believe the best approach that will improve the migration is to prompt the user for confirmation that the two collections will be dropped before proceeding with the execution, (we can also include the number of records found in each of the collections that will be dropped, so that someone can make an informed decision while knowing the potential consequences of running the migration).

m453h:
> > I believe we discussed this before, and if my memory serves me right, it was revolving around using the naming convention for collections that would have the same slugs, which at the time only applied to `pages`. That's why posts wasn't renamed.
> 
> Sounds like a good short-term solution but not sure how it will play long term i.e. when all apps have been migrated. All new collections will have `app-collection` naming but `posts` will stick out and someone new will have to have institutional knowledge above.
> 
> Cool, if we can't do it now, lets create an issue to track this and implement it when time permits.

Sure, I will create an issue to track this...at least I will be able to create another PR following a similar approach to this once we have the pages migrations working correctly

apps/civicsignalblog/migrations/20241011_104214_migration.ts
kilemensi: I hear you but still no... I'm not comfortable with deleting data in a script like this.

If the new collection is already created, the best approach would be to copy items from old collection into new collection using item `slug` as unique key for detecting duplicates (if Mongo supports that).

This means all documents in old collection that are not in the new collection will be copied over. If there is a `slug` clash, the script should log a warning but not override what's in the new collection already.

Someone can review and delete the old collection when they're happy that everything has been migrated into the new collection.

m453h: Cool, let me go with this approach !

m453h: @kilemensi, implementation using the new approach is done, luckily Payload does mark the `slug` as a unique field thus we can easily detect duplicates and log an error. I believe this should be in good order !

apps/civicsignalblog/migrations/20241011_104214_migration.ts
```diff
@@ -0,0 +1,51 @@
...
+      payload.logger.info(
+        `Successfully migrated version collection with slug: ${record.slug} from ${oldCollectionVersionName} to ${newCollectionVersionName}`,
+      );
+    } catch (error) {
+      payload.logger.error(
+        `Error migrating version collection with slug: ${record.slug} from ${oldCollectionVersionName} to ${newCollectionVersionName}`,
+      );
+    }
+  }
+  payload.logger.info(`✓ Successfully completed migration`);
```
kilemensi, line 48: We can't really say "successfully completed migration" here, what if there were errors?

kilemensi (approved):
👍🏽 

----

Lets investigate [`insertMany`](https://www.mongodb.com/docs/manual/reference/method/db.collection.insertMany/) for future improvements

apps/civicsignalblog/migrations/20241011_104214_migration.ts
m453h: 👍  

Now tracking `failedInserts` and displaying an appropriate message

m453h:
> 👍🏽
> 
> Lets investigate [`insertMany`](https://www.mongodb.com/docs/manual/reference/method/db.collection.insertMany/) for future improvements

I actually wanted to start my implementation with `insertMany` one thing I didn't like about it was the lack of flexibility on logging successful inserts, will look more into it for future migrations.

PR #959, opened 2024-10-16
Fix ClimateMapped build public URL

kelvinkipruto:
## Description
The `NEXT_PUBLIC_URL` wad defined but was not used when building the image. This causes a CORS error when accessing PayloadCMS. This issue fixes this bug

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/apps/climatemappedafrica/package.json
+++ b/apps/climatemappedafrica/package.json
@@ -1,6 +1,6 @@
 {
   "name": "climatemappedafrica",
-  "version": "0.0.2",
+  "version": "0.0.3",
   "private": true,
   "author": "Code for Africa <tech@codeforafrica.org>",
   "description": "Climate Mapped Africa ",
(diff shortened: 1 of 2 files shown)


apps/climatemappedafrica/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "climatemappedafrica",
-  "version": "0.0.2",
+  "version": "0.0.3",
```
VinneyJ, line 3: The latest image in dockerhub is at 0.0.3 already. Plus the branch won't build the ARM image since you do not have that functionality in your branch/PR. You can make the env change I will pull and make the version change if need be.

VinneyJ (changes requested):

apps/climatemappedafrica/package.json
kelvinkipruto: @VinneyJ I do not see version 3 in dockerhub. This is also to ensure pushing to dokku will work.

VinneyJ: Awesome then. We can merge.

VinneyJ (approved):

VinneyJ (approved):

Issue #930, opened 2024-10-03
CLimateMapped Tutorial panel


kelvinkipruto:
Create a tutorial panel showing users how to use the site

Request Link: [Feature 9](https://docs.google.com/document/d/1TUrdCRnbj6TSqiuwITKUpnDUEi_ktm_3PmgsaD0y2bw/edit#heading=h.ubjp6vs5uy89)

PR #963, opened 2024-10-18
ClimateMapped: Explore Page

kelvinkipruto:
## Description
PR #945 removed the previous `explore` page
This PR makes the `explore` page work once again.

## Type of change

## Screenshots
image
image
image

## Checklist:


apps/climatemappedafrica/src/payload/fields/LocationSelect.js
```diff
@@ -0,0 +1,40 @@
+import { Select } from "payload/components/forms";
```
koechkevin, line 1: We already merged a PR that included this @kelvinkipruto

apps/climatemappedafrica/src/lib/data/common/index.js
```diff
@@ -74,12 +95,16 @@ export async function getPageProps(api, context) {
...
+  let blocks = await blockify(page.blocks, api, context);
   const variant = page.slug === explorePage.slug ? "explore" : "default";
 
   const siteSettings = await api.findGlobal("settings-site");
   const footer = getFooter(siteSettings, variant);
-  const menus = getNavBar(siteSettings, variant);
+  const menus = getNavBar(siteSettings, variant, explorePage);
+
+  if (slug === explorePage.slug) {
+    blocks = await processExplorePage(slugs.slice(1), hurumap, explorePage);
```
koechkevin, line 106: I see slugs can be undefined in line 80 above

apps/climatemappedafrica/src/lib/data/common/index.js
```diff
@@ -74,12 +95,20 @@ export async function getPageProps(api, context) {
...
-  const blocks = await blockify(page.blocks, api, context);
+  let blocks = await blockify(page.blocks, api, context);
   const variant = page.slug === explorePage.slug ? "explore" : "default";
 
   const siteSettings = await api.findGlobal("settings-site");
   const footer = getFooter(siteSettings, variant);
-  const menus = getNavBar(siteSettings, variant);
+  const menus = getNavBar(siteSettings, variant, explorePage);
+
+  if (slug === explorePage.slug) {
```
koechkevin, line 105: I think we changed params from slug to `slugs` in line 80 above. We can fix it in this PR.

apps/climatemappedafrica/src/lib/data/blockify/tutorial.js
```diff
@@ -0,0 +1,88 @@
+async function tutorial() {
+  // TODO: Move this to Payload CMS
```
koechkevin, line 2: Do we already have an issue for this @kelvinkipruto

koechkevin (approved):
LGTM @kelvinkipruto left a small comment we may have missed in previous PRs

apps/climatemappedafrica/src/lib/data/blockify/tutorial.js
kelvinkipruto: Yes @koechkevin #930

apps/climatemappedafrica/src/payload/globals/HURUMap/index.js
```diff
@@ -15,6 +19,36 @@ const HURUMap = {
       relationTo: ["pages"],
       maxDepth: 1,
       required: true,
+      admin: {
+        description:
+          "The page to use as the Explore page. It will contain the interactive map.",
+      },
+    },
+    {
```
kilemensi, line 27: 1. If this is here (which I think is correct), shouldn't we remove these from Hero block? Otherwise, editors have to set this twice.
2. Should't you include an option to indicate whether pinning should be enabled or not at initial location? I think in PesaYetu, pinning/comparison is enabled for KE.

apps/climatemappedafrica/src/payload/globals/HURUMap/index.js
```diff
@@ -15,6 +19,36 @@ const HURUMap = {
...
+      name: "initialLocation",
+      label: "Initial Location",
+      type: "group",
+      fields: [
+        {
+          name: "name",
+          type: "text",
```
kilemensi, line 34: Again, lets not forget about localized.

apps/climatemappedafrica/src/lib/data/common/index.js
```diff
@@ -47,17 +47,39 @@ function getNavBar(siteSettings, variant) {
   return {
     logo: imageFromMedia(title, primaryLogo.url),
     drawerLogo: imageFromMedia(title, drawerLogo.url),
+    explorePageUrl: slug,
```
kilemensi, line 50: 1. Isn't the selected explore page in `siteSettings`?
2. I don't think `explorePageUrl` is a correct name. It's not a URL; may be just a path or path segment. And explore is a UI name, I think as far as backend/HURUmap is concerned this is related to profile(s), no?

apps/climatemappedafrica/src/lib/data/common/index.js
```diff
@@ -47,17 +47,39 @@ function getNavBar(siteSettings, variant) {
...
     menus,
     socialLinks,
     variant,
   };
 }
 
+async function processExplorePage(slugs, hurumap, explorePage) {
```
kilemensi, line 57: Why is this being handled different from other blocks?

kilemensi (commented):
👍🏽

apps/climatemappedafrica/src/payload/globals/HURUMap/index.js
kelvinkipruto: > If this is here (which I think is correct), shouldn't we remove these from Hero block? Otherwise, editors have to set this twice.

I also agree with this @kilemensi. Initially, we thought the map in the hero could be different, but thinking about it further, it should be the same

apps/climatemappedafrica/src/lib/data/common/index.js
kelvinkipruto: @kilemensi Similar to the above, based on the conversation [here](https://github.com/CodeForAfrica/ui/pull/945#issuecomment-2410564857), the explore page can never have any other block/component other than the interactive map. Also, we don't expect the map to be used on more than 1 page, so we don't need to turn it into a block.

apps/climatemappedafrica/src/lib/data/blockify/explore-page.js
```diff
@@ -0,0 +1,97 @@
...
+    locations,
+    mapType,
+    panel,
+    profile,
+    variant: "explore",
+    preferredChildren,
+  };
+
+  return res;
+  // return {}
```
koechkevin, line 94:
```suggestion
```

apps/climatemappedafrica/src/lib/data/blockify/index.js
```diff
@@ -1,20 +1,24 @@
...
 
-export const blockify = async (blocks, api, context) => {
+export const blockify = async (blocks, api, context, hurumap) => {
   const promises = blocks?.map(async (block) => {
     const slug = block.blockType;
     const propsifyBlock = propsifyBlockBySlug[slug];
 
     if (propsifyBlock) {
-      return propsifyBlock(block, api, context);
+      return propsifyBlock(block, api, context, hurumap);
```
koechkevin, line 21: This "overload" is risky I think perhaps we should use an object instead to avoid what we have done in `blockify/hero.js`

kelvinkipruto: @koechkevin can you clarify more on: 

> we should use an object instead to avoid what we have done in blockify/hero.js

kilemensi (approved):
👍🏽 ... as we're pressed for time, lets get this merged but the approach could be a bit better.

apps/climatemappedafrica/src/lib/data/blockify/index.js
koechkevin: I meant
```suggestion
      return propsifyBlock({ block, api, context, hurumap });
```

kelvinkipruto: Aaah, named params. Great idea.

kelvinkipruto:
> approach could be a bit better

@kilemensi Could you provide a few pointers on how to make it much better?

PR #976, opened 2024-10-25
Climatemapped Tutorial Panel

### LINKED ISSUE #930: CLimateMapped Tutorial panel

kelvinkipruto:
## Description

This PR transforms the HURUMap setting into tabs for easier navigation. It also setups the TutorialPanel for ClimateMappedAfrica HURUMap page.

Fixes #930 

## Type of change

## Screenshots
image

image
image
image

## Checklist:


kilemensi:
Good pull.

A few ideas to think over and lets chat on Monday (_DO NOT WORK ON THIS OVER THE WEEKEND!_):

1. Lets convert HURUmap settings into Tab view with two tabs: **Profile** (or something along those likes) and **Tutorial**
2. **Profile**
   - Lets change the "Explore Page" to something like "Show on page".
   - Lets add "None/none" as the first item in the page select and this should be the default. If this is set, HURUmap functionality is disabled. If any other page is selected, then HURUmap functionality is enabled i.e. everything from this point on.
   - Lets change "Initial Location" to "Root Geography" or something that would be familiar to someone who has used the Django admin (don't remember the name there but I think this is what it is called in the code)
   - Lets change "Allow pinning of initial location" to something along the lines "Root geography has data" (I'm sure you can come up with a better label or again find something that's familiar to users of the Django admin). The idea here is to ask if the root location itself has data or only children have data. If root geography has data, pinning would work; if it doesn't, pinning should be disabled. Asking about pinning directly doesn't get this message across I believe. PS: Have we actually tested pinning after all these changes we've made and confirm that it's still working?

3. **Tutorial**
   - Again, first thing should be to show/not show (or enable/disable) Tutorial... but still remember, this tab should only be active **if** HURUmap functionality is enabled.
   - Instead of "items", I think "steps" is a better name/label?
   - Minor but a list of selectable `selectors` may be better since we don't expect editors to inspect html content of the explore page to figure out the ids.

----

Enjoy your weekend.

koechkevin (approved):

m453h (approved):
Code Looks great 🚀 

Just a minor thing I observed when I tried running the code and the HuruMap global settings couldn't be rendered on the admin UI, I figured it's because of wrong configurations on my end as the `LocationSelect`  component does fetch data from an external API and I haven't set any of those configs.

Screenshot 2024-10-29 at 10 55 33
 
It would be nicer if we could at-least display some message when the fetch fails for this component instead of having a blank page.

kilemensi:
>Just a minor thing I observed when I tried running the code and the HURUMap global settings couldn't be rendered on the admin UI, I figured it's because of wrong configurations on my end as the LocationSelect component does fetch data from an external API and I haven't set any of those configs.
>

I'd go even further and suggest we create a story:

1. By default HURUmap should disabled hence no error/warning needed (done in this PR does?)
2. If enabled, first field should be URL to backend API server, with validation to check that it returns expected data (this will replace the current `HURUMAP_API_URL` ENV var)
3. Everything else in this PR should remain as is.

*_This is not a priority for tomorrow/handover to data team. We can implement it at our own pace_*

apps/climatemappedafrica/src/lib/data/blockify/explore-page.js
```diff
@@ -5,10 +5,10 @@ import {
 
 async function explorePage({ block: { slugs }, hurumap }) {
   const {
-    initialLocation,
+    rootGeography,
     page: { value },
   } = hurumap;
-  const { name } = initialLocation;
+  const { code: name, rootGeographyHasData: pinRootGeography } = rootGeography;
```
kilemensi, line 11: I would have kept `rootGeography` and `rootGeographyHasData` names instead of `initialLocation` and `pinRootGeography` or `pinInitialLocation`.

apps/climatemappedafrica/src/payload/globals/HURUMap/Geography.js
```diff
@@ -0,0 +1,55 @@
+import LocationSelect, { validateLocation } from "../../fields/LocationSelect";
+
+const Geography = {
```
kilemensi, line 3: Shouldn't this be named `RootGeography`?

apps/climatemappedafrica/src/payload/globals/HURUMap/Geography.js
```diff
@@ -0,0 +1,55 @@
...
+          name: "rootGeographyHasData",
+          type: "checkbox",
+          localized: true,
+          label: {
+            en: "Root geography has data",
+          },
+          defaultValue: false,
+          admin: {
+            description:
+              "Indicates whether the root geography itself has data. If checked, pinning will be enabled; if unchecked, only child locations will have data and pinning will be disabled.",
```
kilemensi, line 47:
```suggestion
              "Indicates whether the root geography itself has data that can be used for comparison with its children.",
```

apps/climatemappedafrica/src/payload/globals/HURUMap/Profile.js
```diff
@@ -0,0 +1,21 @@
...
+      label: {
+        en: "Show on Page",
+      },
+      localized: true,
+      type: "relationship",
+      relationTo: ["pages"],
+      maxDepth: 1,
+      required: true,
+      admin: {
+        description: "The page to show the interactive map on.",
```
kilemensi, line 15:
```suggestion
        description: "The page to show the HURUmap profile on.",
```

apps/climatemappedafrica/src/payload/globals/HURUMap/Tutorial.js
```diff
@@ -0,0 +1,122 @@
...
+    value: "pin",
+    label: "Pin",
+  },
+];
+
+const Tutorial = {
+  label: "Tutorial",
+  fields: [
+    {
+      name: "enableTutorial",
```
kilemensi, line 29: Unless Payload rejects them, I think simpler names are better. The global is already called Tutorial
```suggestion
      name: "enabled",
```

apps/climatemappedafrica/src/payload/globals/HURUMap/Tutorial.js
```diff
@@ -0,0 +1,122 @@
...
+      label: {
+        en: "Enable Tutorial",
+      },
+      type: "checkbox",
+      defaultValue: true,
+      localized: true,
+    },
+    {
+      name: "tutorialPanel",
```
kilemensi, line 38:
```suggestion
      name: "panel",
```

apps/climatemappedafrica/src/payload/globals/HURUMap/Tutorial.js
```diff
@@ -0,0 +1,122 @@
...
+            en: "Steps",
+          },
+          type: "array",
+          localized: true,
+          required: true,
+          admin: {
+            initCollapsed: true,
+            components: {
+              RowLabel: ({ data, index }) => {
+                return data?.title || `Item ${String(index).padStart(2, "0")}`;
```
kilemensi, line 60:
```suggestion
                return data?.title || `Step ${String(index).padStart(2, "0")}`;
```

apps/climatemappedafrica/src/payload/globals/HURUMap/index.js
```diff
@@ -13,58 +15,17 @@ const HURUMap = {
...
-      label: "Explore Page",
-      type: "relationship",
-      relationTo: ["pages"],
-      maxDepth: 1,
-      required: true,
-      admin: {
-        description:
-          "The page to use as the Explore page. It will contain the interactive map.",
-      },
+      name: "enableHURUMap",
```
kilemensi, line 18:
```suggestion
      name: "enabled",
```

kilemensi (approved):

PR #977, opened 2024-10-25
Fix @/climatemappedafrica images on Dev

kilemensi:
## Description

Main fix is adding a custom loader that will delegate loading of images coming from CMS (`/media`) to payload directly rather than via Next.js which may fail in weird ways (e.g. Next.js tries to load them via `/_next/image?url=`). The PR also:

  - [x] Improves the local/S3 image handling: If S3 key and secret are found, S3 is used. Otherwise, local media folder is used
  - [x] Fixes the print icon on explore page
  - [x] Ensure the panel in desktop mode scales well for screens that are lg+ i.e. the rich data/pin icons move accordingly.
  - [x] Ensure `TotorialComponentProps` exist when there is an `tutorial` block only.

PS: Lets test the image loading S3/local well to make sure we've fixed it once and for all.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Screenshots

Screenshot 2024-10-25 at 14 55 32

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation

PR #978, opened 2024-10-25
ClimateMappedAfrica Default Variables

kelvinkipruto:
## Description

This PR setups some default environment variables on ClimateMappedAfrica.

## Type of change

## Checklist:


VinneyJ (approved):

kilemensi:
Can you may be describe what issue is this PR fixing? For example, I don't think `NEXT_PUBLIC_SEO_DISABLED` is even supported in the `climatemappedafrica` app.

kilemensi:
Isn't that same issue fixed in https://github.com/CodeForAfrica/ui/pull/977 @kelvinkipruto ?

kelvinkipruto:
@kilemensi Yes, it can fix it. I hadn't looked at your PR before working on this.

apps/climatemappedafrica/.env
```diff
@@ -1,3 +1,7 @@
+NEXT_PUBLIC_IMAGE_UNOPTIMIZED=true
+NEXT_PUBLIC_SEO_DISABLED=true
+PAYLOAD_PUBLIC_DEFAULT_LOCALE=en
+PAYLOAD_PUBLIC_LOCALES="en"
```
koechkevin, line 4: Are they needed at buildtime?

kelvinkipruto: looks like we do not need them

kelvinkipruto:
This PR is not needed as the issues will be fixed in #977

PR #983, opened 2024-10-29
@/CivicSignal: Fix previews on research blog content

### LINKED ISSUE #907: Fix CivicSignal preview issues on ResearchBlog

### LINKED ISSUE #999: Previews not working when draft mode is enable on the CivicSignal Research blog
Currently previews are not working when draft mode is enabled in the CivicSignal Research blog.

m453h:
## Description
This PR introduces several fixes to the CivicSignal Research blog to enable previews and address some issues on the blog's UI.

Fixes #999 

## Type of change

## Checklist:


apps/civicsignalblog/src/components/ArticleGrid/ArticleGrid.js
```diff
@@ -26,12 +26,12 @@ const ArticleGrid = React.forwardRef(function ArticleGrid(props, ref) {
   return (
     <Section sx={{ px: { xs: "20px", sm: 0 }, ...sx }} {...other} ref={ref}>
       <Grid container sx={{ py: { xs: "28px", md: 8, lg: "105.29px" } }}>
-        <Grid
+        <Box
           item
           xs={12}
```
kilemensi, line 31: Box doesn't support `item` or `xs` props.

apps/civicsignalblog/src/payload/access/canAccessApplication.js
```diff
@@ -4,7 +4,7 @@ export default async function canAccessApplication(req, searchString) {
   if (user) {
     const app = headers["CS-App"] || user.currentApp || user.defaultApp;
     return (
-      app.toLowerCase() === searchString.toLowerCase() &&
+      app?.toLowerCase() === searchString.toLowerCase() &&
```
kilemensi, line 7: In what scenario will all of `headers["CS-App"]`, `user.currentApp`, and `user.defaultAp` be `undefined`?

apps/civicsignalblog/src/payload/components/allowedAppSelect/index.tsx
```diff
@@ -32,7 +32,7 @@ function CustomSelectComponent({ path, label }) {
         path={path}
         name={path}
         options={options}
-        value={String(value)}
+        value={value ? String(value) : ``}
```
kilemensi, line 35: Why do we need `String(value)` instead of just `value`?

kilemensi (approved):
LGTM!

apps/civicsignalblog/src/payload/access/canAccessApplication.js
m453h: This only occurs when we login  on the admin UI after we have created the first account on a fresh install.
Because at the time of the account creation we can not render the  options for `user.CurrentApp` or user.defaultApp since these depend on the user's `allowedApps` which will always be undefined.

kilemensi: ... what do we do in that case? Is the user allowed to access any app?

m453h: If the user had selected Research app among their `allowedApps`, then we would default `user.CurrentApp `to `Research`, otherwise we would just allow them to manage collections that are not tied to any specific application i.e. Media and Users as  👇 

Screenshot 2024-11-14 at 13 21 50

apps/civicsignalblog/src/payload/components/allowedAppSelect/index.tsx
m453h: It was more of trying to overcome Typescript not knowing if value is really a string but I have added proper typing on the useField hook

kilemensi (approved):

PR #1001, opened 2024-11-14
Ft/upgrade next to 15

koechkevin:
## Description

This PR upgrades the Next.js framework from the current version to v15.0.3 to leverage on the latest features, performance improvements, and bug fixes introduced in this release.

### Key Changes
- Upgrade Next.js Version to 15.0.3.
- Upgrade Eslint to v9 since next js 15 added support for eslint 9.
- Updated codebase's eslint configuration to [flatConfigs](https://eslint.org/blog/2024/04/eslint-v9.0.0-released/#removed-all-formatters-except-stylish%2C-html%2C-json%2C-and-json-with-meta).
- Removed `.eslintignore` and `.eslintrc` since with flatConfigs all can be defined in eslint.config.js.

You may notice many changed files in component files, but this is due to a change in how to disable a rule in a file with flat configs.
- We do not need to replicate `// eslint-disable-next-line testing-library/render-result-naming-convention` inside the test file since this disable command has been specified in common eslint config `"testing-library/render-result-naming-convention": "off",`.

Review of this PR should be around `*/**/eslint.config.js` and `packages/eslint-config-commons-ui`.

Find a write-up of how these configs work [here](https://github.com/CodeForAfrica/ui/blob/9a76ce44332a48f7e2c87b7dcc8ea2cb8cfe670d/packages/eslint-config-commons-ui/README.md)

Fixes # (issue)

## Type of change

## Checklist:


packages/eslint-config-commons-ui/next.js
```diff
@@ -1,3 +1,11 @@
...
-};
+const { fixupConfigRules } = require("@eslint/compat");
+const { FlatCompat } = require("@eslint/eslintrc");
+
+const commonConfig = require("./index");
+
+const flatCompat = new FlatCompat();
+
+module.exports = [
+  ...fixupConfigRules(flatCompat.extends("plugin:@next/next/core-web-vitals")),
```
kilemensi, line 9: Why do we need this? Aren't we [going](https://nextjs.org/blog/next-15#eslint-9-support) all in [Eslint 9](https://eslint.org/blog/2024/04/eslint-v9.0.0-released/#flat-config-is-now-the-default-and-has-some-changes)?

packages/hurumap-next/eslint.config.js
```diff
@@ -0,0 +1,34 @@
...
+      "**/npm-debug.log*",
+      "**/yarn-debug.log*",
+      "**/yarn-error.log*",
+      "**/.vercel",
+      "**/.now",
+      "**/.turbo",
+      "**/test-results/",
+      "**/playwright-report/",
+    ],
+  },
```
kilemensi, line 26: Can't we defined all these ignores in either `eslint-config-commons-ui/core` or `eslint-config-commons-ui/next` and just reuse them in other packages/apps?

apps/vpnmanager/src/middleware.ts
```diff
@@ -7,7 +7,7 @@ export const config = {
 
 export function middleware(req: NextRequest) {
   const key: string | null = req.headers.get("x-api-key");
-  const API_SECRET_KEY = process.env.API_SECRET_KEY;
+  const { API_SECRET_KEY } = process.env;
```
kilemensi, line 10: Don't [destructure](https://nextjs.org/docs/app/api-reference/next-config-js/env) env vars.
>Next.js will replace process.env.customKey with 'my-value' at build time. Trying to destructure process.env variables won't work due to the nature of webpack [DefinePlugin](https://webpack.js.org/plugins/define-plugin/).

kilemensi (commented):
👍🏽 

---

Still going through it but quick feedback on issues that kind of stood out to me.

packages/eslint-config-commons-ui/next.js
koechkevin: @kilemensi I came across [this issue](https://github.com/vercel/next.js/issues/64114) and the only workaround I had was to use [eslint compatibility utilities](https://eslint.org/blog/2024/05/eslint-compatibility-utilities/) to fix up things.

apps/vpnmanager/src/lib/outline.ts
```diff
@@ -9,8 +9,11 @@ import { fetchJson } from "@/vpnmanager/utils";
 
 export class OutlineVPN {
   apiUrl: string;
+
   fingerprint?: string;
+
   timeout?: number;
+
```
kilemensi, line 16: Why so many empty lines?

kilemensi (commented):
👍🏽 

---

Are we using [this comment](https://github.com/eslint/eslint/discussions/16960#discussioncomment-10985494), should we be using [the experimental file configuration](https://eslint.org/docs/latest/use/configure/configuration-files#experimental-configuration-file-resolution)?

koechkevin:
> 👍🏽
> 
> Based on [this comment](https://github.com/eslint/eslint/discussions/16960#discussioncomment-10985494), should we be using [the experimental file configuration](https://eslint.org/docs/latest/use/configure/configuration-files#experimental-configuration-file-resolution)?

@kilemensi this seems to work for js apps if we put eslint configs at root but since ts apps configurations have some differences, we may have to retain them as is.

kilemensi:
>@kilemensi this seems to work for js apps if we put eslint configs at root but since ts apps configurations have some differences, we may have to retain them as is.
>

There is a slight chance I've misread it @koechkevin so please read the docs again but my understanding is as follow:
Given the following directory tree:
```
  <root>
  ├── eslint.config.js
  ├── apps
  │   ├── centralcms
  │   ├── charterafrica
  │   │   ├── src
  │   │   ├── eslint.config.json
  │   ├── civicsignalblog
  ├── packages
  │   ├── commons-ui-core
  │   ├── commons-ui-next
  │   ├── commons-ui-payload
```

Without the flag, `eslint` looks for config from current directory moving up. That is, if you run the command at `<root>`, if there is a `eslint.config.js` at root, that config will be used regardless of any other config at `apps` level. If it doesn't find one at `<root>`, then it will travel `<root>` -> `apps` -> `charterafrica` and use the first found `eslint.config.js` along that path.

With the flag on, the process is actually reversed. `eslint` starts looking at the file being linted regardless of where the command was executed. So if we're linting a file in `charterafrica/src`, eslint will traverse backwards and find `eslint.config.json` in `charterafrica` first, or along the path `charterafrica` -> `apps` -> `<root>` until it finds one.

In a monorepo, the second approach sounds like the right approach to me.

kilemensi:
I just saw

> `chore: Remove obsolete ESLint configuration files across multiple app`
>

fly by my screen @koechkevin. Are we changing the approach?

koechkevin:
> I just saw
> 
> > `chore: Remove obsolete ESLint configuration files across multiple app`
> 
> fly by my screen @koechkevin. Are we changing the approach?

@kilemensi been playing around with this and it looks like even without a flag 🤔 eslint traverses backwards. So I just removed eslint configurations in apps and retained one in the root. I had to push it just to be sure that things aren't cached on my end.

koechkevin:
@kelvinkipruto @m453h spare some time here

kelvinkipruto:
@koechkevin Just tested it on ClimateMapped, and it works fantastic. Perhaps we can now use `use cache` to improve its speed😅.
Let me still review the other apps

apps/promisetracker/eslint.config.js
```diff
@@ -0,0 +1,14 @@
...
+    settings: {
+      "import/resolver": {
+        webpack: {
+          config: "./eslint.webpack.config.js",
+        },
+      },
+    },
+  },
+  ...eslintConfig,
+];
```
m453h, line 14: since this is almost identical to what we have on [PesaYetu](https://github.com/CodeForAfrica/ui/blob/d9bbe3f518a102b384b6859e163a1a6f19c64cd4/apps/pesayetu/eslint.config.js) I was just wondering if there was any specific reason of not starting with  spreading eslintConfig and then keep any custom configs

m453h (commented):
🚀 

 I have tested `CivicSignalBlog`, `ClimateMapped` and `VPNManager` and they have worked , I'll need to test the remaining apps as well.

---
Just a thought I had as I was reading through this [discussion](https://github.com/eslint/eslint/discussions/16960)... would it make sense to have a root-level `eslint.config.js` and extend it per specific app's needs ?  Or would that lead to unnecessary complexity?

koechkevin:
> 🚀
> 
> I have tested `CivicSignalBlog`, `ClimateMapped` and `VPNManager` and they have worked , I'll need to test the remaining apps as well.
> 
> Just a thought I had as I was reading through this [discussion](https://github.com/eslint/eslint/discussions/16960)... would it make sense to have a root-level `eslint.config.js` and extend it per specific app's needs ? Or would that lead to unnecessary complexity?

@m453h The configs for all apps are similar, would it make sense to have as many eslint configs when only 1 at the root can work?

m453h:
> > 🚀
> > I have tested `CivicSignalBlog`, `ClimateMapped` and `VPNManager` and they have worked , I'll need to test the remaining apps as well.
> > Just a thought I had as I was reading through this [discussion](https://github.com/eslint/eslint/discussions/16960)... would it make sense to have a root-level `eslint.config.js` and extend it per specific app's needs ? Or would that lead to unnecessary complexity?
> 
> @m453h The configs for all apps are similar, would it make sense to have as many eslint configs when only 1 at the root can work?

Yes, I was thinking that just one `eslint.config.js` file at the root level would make more sense,  though I had seen we still had separate files for some apps like `civicsignalblog`, `vpnmanager`, `techlab`, `uibook`, ~~`pesayetu`~~, ~~`promisetracker`~~

koechkevin:
> > > 🚀
> > > I have tested `CivicSignalBlog`, `ClimateMapped` and `VPNManager` and they have worked , I'll need to test the remaining apps as well.
> > > Just a thought I had as I was reading through this [discussion](https://github.com/eslint/eslint/discussions/16960)... would it make sense to have a root-level `eslint.config.js` and extend it per specific app's needs ? Or would that lead to unnecessary complexity?
> > 
> > 
> > @m453h The configs for all apps are similar, would it make sense to have as many eslint configs when only 1 at the root can work?
> 
> Yes, I was thinking that just one `eslint.config.js` file at the root level would make more sense, though I had seen we still had separate files for some apps like `civicsignalblog`, `vpnmanager`, `techlab`, `uibook`, ~`pesayetu`~, ~`promisetracker`~

@m453h typescript apps' configurations are slightly different from those for JS apps so we'd have to retain app-level configurations for the TS apps.

koechkevin:
@CodeForAfrica/tech ?

m453h:
> @CodeForAfrica/tech ?

Hi @koechkevin just early feedback tested running several apps noticed there is an issue on [CivicSignal Research](https://github.com/CodeForAfrica/ui/blob/main/apps/civicsignalblog), its failing to resolve this custom component

```
src/payload/components/allowedAppSelect/index.js
```


I did a quick test by renaming the component to `.tsx` and importing it as shown below 👇  and it resolves this issue.

```
import CustomSelectComponent from "#civicsignalblog/payload/components/allowedAppSelect/index"
```

I recall keeping this component with `.js` extension in order to auto import this component without needing to use the `index` name e.g. in [defaultApp.js](https://github.com/CodeForAfrica/ui/blob/main/apps/civicsignalblog/src/payload/fields/customSelect/defaultApp.js), and [currentApp.js](https://github.com/CodeForAfrica/ui/blob/main/apps/civicsignalblog/src/payload/fields/customSelect/currentApp.js) haven't figured out yet why moving to the `eslint.config.js` is causing this issue though

kelvinkipruto:
@koechkevin A few issues I have encountered.
1. In NextJs 15, [Dynamic APIs are Asynchronous](https://nextjs.org/docs/messages/sync-dynamic-apis). Some props must be awaited. `techlabblog` throws errors regarding this. Running the recommended [codemod](https://nextjs.org/docs/messages/sync-dynamic-apis#possible-ways-to-fix-it) fixes it.
2. `"target": "ES2017"` is missing in `climatemappedfrica`'s [`tsconfig.json`](https://github.com/CodeForAfrica/ui/blob/ft/upgrade_next_to_15/apps/climatemappedafrica/tsconfig.json)
3. Looking at the `eslint.config.js` for the typescript apps, they are all similar. Is it possible to have one configuration that can be reused for all the apps?

m453h:
> > import CustomSelectComponent
> 
> Renamed to `index.tsx` and I don't seem to find any issue @m453h

Great we can keep the custom components in `tsx`, have you changed the import ? Because build would fail to resolve the module unless  we explicitly import it as:

```
import CustomSelectComponent from "#civicsignalblog/payload/components/allowedAppSelect/index";
```

kilemensi:
@koechkevin , to move things along, this is what I raised on Friday:

Before this upgrade, there were 2 eslint configs in[eslint-config-commons-ui](https://github.com/CodeForAfrica/ui/tree/main/packages/eslint-config-commons-ui):
1. index.js: This is the default config and it's the React only config. It is useful in other packages e.g. commons-ui/core, hurumap/core, etc., that do not use Next.js
2. next.js: This is the Next.js config (JavaScript only). It is useful in both packages and apps that use Next.js.

Each package/app chose which of the above config to use and import it via package/app-level .eslintrc.js

Moving to eslint 9, my questions are:
1. How are we introducing TypeScript support (initially just Next.js)?
2. Based on `1`, how many "reusable" configs will we now have in the `eslint-config-commons-ui`?
4. Using flatConfig, where and how are the various `eslint.config.js` applied to cover all packages and apps in the repo?

koechkevin:
> @koechkevin , to move things along, this is what I raised on Friday:
> 
> Before this upgrade, there were 2 eslint configs in [eslint-config-commons-ui](https://github.com/CodeForAfrica/ui/tree/main/packages/eslint-config-commons-ui):
> 
> 1. `index.js`: This is the default config and it's the React only config (JavaScript only). It is useful in other packages e.g. commons-ui/core, hurumap/core, etc., that do not use Next.js
> 2. `next.js`: This is the Next.js config (JavaScript only). It is useful in both packages and apps that use Next.js.
> 
> Each package/app chose which of the above config to use and import it via package/app-level `.eslintrc.js`
> 
> Moving to eslint 9, my questions are:
> 
> 1. How are we introducing TypeScript support (initially just Next.js)?
> 2. Based on `1`, how many "reusable" configs will we now have in the `eslint-config-commons-ui`?
> 3. Using flatConfig, where and how are the various `eslint.config.js` applied to cover all packages and apps in the repo?

@kilemensi to answer your questions,
Since we are now using `eslint --flag unstable_config_lookup_from_file --fix './'` command, we have 1 `eslint.config.js` that will be used by all next js apps which do not have their own `eslint.config.js`.
1. We are introducing common configs for typescript-next apps i.e `typescript.js`  in `eslint-common-config`. This basically extends `./next` configs. Individual apps will now have their own `eslint.config.js` implementing these configs. 
```
const eslintConfig = require("eslint-config-commons-ui/typescript");

module.exports = eslintConfig;
```
2. We now have 3 reusable configs i.e `index.js`, `next.js` and `typescript.js`
3. For the packages, we only have one config file under packages directory which implements `index.js`

kilemensi:
> 3. For the packages, we only have one config file under packages directory which implements index.js
>

Some packages such as `commons-ui/next`, `hurumap-next`, etc. should be checked with `next.js` instead of `index.js` @koechkevin.

koechkevin:
> > 3. For the packages, we only have one config file under packages directory which implements index.js
> 
> Some packages such as `commons-ui/next`, `hurumap-next`, etc. should be checked with `next.js` instead of `index.js` @koechkevin.

@kilemensi 
For the packages that use the `next` configs, we use the same approach as we used in typescript-next. They have their own config files extending `./next`. [commons-ui-next](https://github.com/CodeForAfrica/ui/blob/main/packages/commons-ui-next/.eslintrc.js) extended `index.js` was it a mistake?

kilemensi:
> ... [commons-ui-next](https://github.com/CodeForAfrica/ui/blob/main/packages/commons-ui-next/.eslintrc.js) extended `index.js` was it a mistake?
>

Most likely @koechkevin; see [`hurumap-next`](https://github.com/CodeForAfrica/ui/blob/41f5cdc0cb3e7849457bdfc297b47a2c91493ee3/packages/hurumap-next/.eslintrc.js#L3)

koechkevin:
closes #1023 #1024

koechkevin:
@m453h @kelvinkipruto ?

kelvinkipruto (approved):
LGTM

kilemensi:
I take @CodeForAfrica/tech are happy with this PR @koechkevin ? If yes, lets merge it. Any outstanding issue, will be fix in a separate PR.

PR #1006, opened 2024-11-22
@/CivicSignalBlog: Misc fixes on CivicSignal blog

m453h:
## Description

This PR includes fixes for two issues on the CivicSignal blog;

1. The `FilterBar` for stories is being rendered even when there is just one featured story, this bug was introduced by #983 when resolving the issue of  the `FilterBar` being incorrectly rendered too close to the `FeaturedArticle`. 

2. Site Header and Footer content not being loaded due to the changed name of the globals configuration slug from `settings site` to `settings-research-site`

## Type of change

## Screenshots
Before the fixes
Screenshot 2024-11-22 at 12 31 18

After the fixes
Screenshot 2024-11-22 at 12 28 37

## Checklist:

### DIFF
--- a/apps/civicsignalblog/migrations/20241122_054517_migration.ts
+++ b/apps/civicsignalblog/migrations/20241122_054517_migration.ts
@@ -39,7 +39,7 @@ export async function up({ payload }: MigrateUpArgs): Promise<void> {
       `✓ Successfully migrated global config [${oldGlobalConfigType}] to [${newGlobalConfigType}]`,
     );
   } catch (error) {
-    payload.logger.error("Error: Migration failed.");
+    payload.logger.error("Error: Migration 20241122_054517 failed.");
     throw error;
   }
 }
--- a/apps/civicsignalblog/migrations/20241122_054517_migration.ts
+++ b/apps/civicsignalblog/migrations/20241122_054517_migration.ts
@@ -9,11 +9,11 @@ export async function up({ payload }: MigrateUpArgs): Promise<void> {
   await mongoose.connect(mongoURL);
   const db = mongoose.connection.db;
 
-  const existingGlobalConfig = await db.collection("globals").findOne({
+  const existingOldGlobalConfig = await db.collection("globals").findOne({
     globalType: oldGlobalConfigType,
   });
 
-  if (!existingGlobalConfig) {
+  if (!existingOldGlobalConfig) {
     payload.logger.error(
       `No existing global settings [${oldGlobalConfigType}] config found, exiting...`,
     );
--- a/apps/civicsignalblog/migrations/20241122_054517_migration.ts
+++ b/apps/civicsignalblog/migrations/20241122_054517_migration.ts
@@ -32,12 +32,8 @@ export async function up({ payload }: MigrateUpArgs): Promise<void> {
   }
 
   try {
-    newGlobalConfig = {
-      ...existingGlobalConfig,
-      globalType: newGlobalConfigType,
-    };
-
-    delete newGlobalConfig._id;
+    const { _id, globalType, ...newGlobalConfig } = existingOldConfig;
+    newGlobalConfig.globalType = newGlobalConfigType;
     await db.collection("globals").insertOne(newGlobalConfig);
     payload.logger.info(
       `✓ Successfully migrated global config [${oldGlobalConfigType}] to [${newGlobalConfigType}]`,
--- a/apps/civicsignalblog/migrations/20241122_054517_migration.ts
+++ b/apps/civicsignalblog/migrations/20241122_054517_migration.ts
@@ -14,7 +14,7 @@ export async function up({ payload }: MigrateUpArgs): Promise<void> {
   });
 
   if (!existingOldGlobalConfig) {
-    payload.logger.error(
+    payload.logger.info(
       `No existing global settings [${oldGlobalConfigType}] config found, exiting...`,
     );
     return;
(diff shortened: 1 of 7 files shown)


apps/civicsignalblog/migrations/20241122_054517_migration.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export async function up({ payload }: MigrateUpArgs): Promise<void> {
+  await mongoose.connect(mongoURL);
+  const db = mongoose.connection.db;
+
+  const existingGlobalConfig = await db.collection("globals").findOne({
+    globalType: oldGlobalConfigType,
+  });
+
+  if (!existingGlobalConfig) {
```
kilemensi, line 16:
```suggestion
  const existingOldGlobalConfig = await db.collection("globals").findOne({
    globalType: oldGlobalConfigType,
  });

  if (!existingOldGlobalConfig) {
```

apps/civicsignalblog/migrations/20241122_054517_migration.ts
```diff
@@ -0,0 +1,50 @@
...
+    payload.logger.error(
+      `No existing global settings [${oldGlobalConfigType}] config found, exiting...`,
+    );
+    return;
+  }
```
kilemensi, line 21: Should this **really** be an error (`payload.logger.error`)?

apps/civicsignalblog/migrations/20241122_054517_migration.ts
```diff
@@ -0,0 +1,50 @@
...
+  let newGlobalConfig = await db.collection("globals").findOne({
+    globalType: newGlobalConfigType,
+  });
+
+  if (newGlobalConfig) {
+    payload.logger.error(
+      `Migration failed: Unable to migrate [${oldGlobalConfigType}] to [${newGlobalConfigType}] because a global configuration with this name already exists.`,
+    );
+    return;
+  }
```
kilemensi, line 32:
```suggestion
  const existingNewGlobalConfig = await db.collection("globals").findOne({
    globalType: newGlobalConfigType,
  });

  if (existingNewGlobalConfig) {
    payload.logger.error(
      `Migration 20241122_054517 failed: Global type [${newGlobalConfigType}] already exists.`,
    );
    return;
  }
```

Just used `20241122_054517` from filename as migration name but if there is a better name, lets use that.

apps/civicsignalblog/migrations/20241122_054517_migration.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  try {
+    newGlobalConfig = {
+      ...existingGlobalConfig,
+      globalType: newGlobalConfigType,
+    };
+
+    delete newGlobalConfig._id;
```
kilemensi, line 40:
```suggestion
    const { _id, globalType, ...newGlobalConfig } = existingOldConfig;
    newGlobalConfig.globalType = newGlobalConfigType;
```

kilemensi (commented):
👍🏽 

---

1. Lets prefer using `const` over `let`.
2. Is the `ArticleGrid` component in this app different from the one in [`codeforafrica`](https://github.com/CodeForAfrica/ui/blob/main/apps/codeforafrica/src/components/ArticleGrid/ArticleGrid.js) visible [here](https://cfa.dev.codeforafrica.org/stories)?

m453h:
> 👍🏽
> 
> 1. Lets prefer using `const` over `let`.
> 2. Is the `ArticleGrid` component in this app different from the one in [`codeforafrica`](https://github.com/CodeForAfrica/ui/blob/main/apps/codeforafrica/src/components/ArticleGrid/ArticleGrid.js) visible [here](https://cfa.dev.codeforafrica.org/stories)?

Yes this is exactly the same component, I've replaced the current implementation in this PR with what we have on `codeforafrica` it does solve the original problem we had in the `civicsignalblog` which was being rendered as 👇 
Screenshot 2024-11-22 at 13 26 49

Instead of:
Screenshot 2024-11-22 at 13 28 33


Only one problem that I will push a fix for both `codeforafrica` and `codeforafrica` is the case of when we are only having one story, we shouldn't render the filter bar 👇 
Screenshot 2024-11-22 at 13 30 49

apps/civicsignalblog/migrations/20241122_054517_migration.ts
m453h: Filename looks good, we can easily trace which migration failed👍

apps/civicsignalblog/migrations/20241122_054517_migration.ts
m453h: Hmm, we can keep it as `info` 🤔

apps/civicsignalblog/migrations/20241122_054517_migration.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  try {
+    const { _id, globalType, ...newGlobalConfig } = existingOldGlobalConfig;
+    newGlobalConfig.globalType = newGlobalConfigType;
+    await db.collection("globals").insertOne(newGlobalConfig);
+    payload.logger.info(
+      `✓ Successfully migrated global config [${oldGlobalConfigType}] to [${newGlobalConfigType}]`,
+    );
+  } catch (error) {
+    payload.logger.error("Error: Migration failed.");
```
kilemensi, line 42:
```suggestion
    payload.logger.error("Error: Migration 20241122_054517 failed.");
```

kilemensi (approved):

PR #1008, opened 2024-11-27
Remove useStyles from HURUMap

kelvinkipruto:
## Description
This PR removes `useStyles` from HURUMap

## Type of change

## Screenshots
image 
image
image
image
iPhone 12 Pro-1732875510535
iPhone 12 Pro-1732875577791

## Checklist:


kilemensi (commented):
👍🏽 

---

Haven't checked the code but you've changed heading spacing and things look unbalanced.
S

kilemensi:
Headings look good now @kelvinkipruto but Select looks off.
S

m453h (commented):
Mostly looks good to me  🚀 I only have a few very minor issues 

1. There’s an issue with the Pin and Compare component when focused, I can see the shadow/tint slightly overlaps the border radius. Although this is visible in the deployed app, it’s more noticeable in the current version 👇

Screenshot 2024-11-29 at 20 21 41
 I played a bit with the CSS on the browser but explicitly setting the borderRadius on focus state kinda solves the problem:

```
 "& .MuiSelect-filled:focus": {
            borderRadius: theme.typography.pxToRem(6),
 },
```

Screenshot 2024-11-29 at 20 37 41

2. Some of these icons look like they are cut at the bottom,  Not sure why this is happening though might be a good idea to fix them on the same PR, and I was just thinking out loud since we are using  𝕏 instead of the bird as one of the social media icons on the footer, don't we need to update this one as well 🤔 
Screenshot 2024-11-29 at 21 08 33

kelvinkipruto:
@kilemensi @m453h @koechkevin Thi si ready fo another round of reviews.

m453h:
We are almost there, I'm not sure why this is happening but the dropdown items seem a bit off in terms of size (probably missing some padding) and there is some extra margin on the left.

Screenshot 2024-12-03

apps/climatemappedafrica/src/components/HURUmap/CategoryHeader/index.js
```diff
@@ -1,31 +1,53 @@
...
 function CategoryHeader({ title, description, icon, ...props }) {
-  const classes = useStyles();
-
   if (!title) {
     return null;
   }
   return (
-    <div className={classes.root}>
-      <Typography {...props} variant="h3" className={classes.title}>
+    <div>
```
koechkevin, line 10: Why not use Box instead

koechkevin (approved):

apps/climatemappedafrica/src/components/HURUmap/ChartFilter/index.js
```diff
@@ -86,44 +83,97 @@ function ChartFilter({
...
+              }}
+              MenuProps={{
+                "& .MuiMenu-paper": {
+                  height: theme.typography.pxToRem(310),
+                  boxShadow: "none",
+                },
+                "& .MuiMenu-list": {
+                  padding: `${theme.typography.pxToRem(10)} 0 ${theme.typography.pxToRem(
+                    10,
+                  )} ${theme.typography.pxToRem(15)} !important`,
```
m453h, line 122: I think this one should be set to 0, because if we set any value it would look a bit awkward with some extra padding on the left
Screenshot 2024-12-04 at 14 09 10

m453h (approved):
Just (2) minor issues but looks good to me

apps/climatemappedafrica/src/components/HURUmap/CategoryHeader/index.js
kelvinkipruto: @koechkevin A `<Box/>` is rendered as a <`div>`. Given that we do not need to style it, is there need to use it?

PR #1009, opened 2024-11-28
@/CivicSignalBlog: Add Web Tools authentication form configuration

m453h:
## Description

This PR adds configuration globals for authentication forms used in the CivicSignal CMS. 

Partially fixes [this issue](https://github.com/CodeForAfrica/web-tools/issues/21) opened in the CivicSignal web-tools 
## Type of change

## Screenshots
Screenshot 2024-11-28 at 09 08 16

## Checklist:


kilemensi:
Looks good @m453h (haven't checked the code yet). A few quick thoughts (totally open for discussion)

1. Why not have `Forms` as a separate group (rather than settings) where we can have Login form, Registration form, etc.
2. Instead of `Input Elements` and `Form Messages` as separate groups, why not combine them into one group `Fields`? Then for each field, you'll have two entries: `label` and `errorMessage` e.g. `email.label`, and `email.errorMessage`. Additionally there can be a third entry `helpMessage` (or `placeholder`) if a field needs to be extra descriptive.

Of course, we may need a common place to put those common message that are the same across all forms, etc., but may be we don't need it now? If we do, then this can go into `Settings`.

m453h:
> * Why not have `Forms` as a separate group (rather than settings) where we can have Login form, Registration form, etc.
Hmmm 🤔 I like the idea of splitting the forms... let me move the components as per this suggestion and group the inputs as well, I think it  would greatly improve the user experience

m453h:
> > * Why not have `Forms` as a separate group (rather than settings) where we can have Login form, Registration form, etc.

Hmmm 🤔 I like the idea of splitting the forms... let me move the components as per this suggestion and group the inputs as well, I think it  would greatly improve the user experience

m453h:
> > > * Why not have `Forms` as a separate group (rather than settings) where we can have Login form, Registration form, etc.
> 
> Hmmm 🤔 I like the idea of splitting the forms... let me move the components as per this suggestion and group the inputs as well, I think it would greatly improve the user experience

@kilemensi, let me know if this arrangement for the login form makes more sense

https://github.com/user-attachments/assets/d2b69ae8-fa8f-4da9-8f84-79daa7aa85a4

kilemensi (commented):
👍🏽 ... Definitely on the right track @m453h .

---

Not sure about "dynamic" fields. Don't we need to match names to what APIs expect e.g. `email` field must be called `email`, etc. If content editor renames it to emailAddress, will the form still work?

Besides that:
Screenshot 2024-12-02 at 17 10 30
Lets arrange the sections as **Publication**, **Forms** and then **Settings** last.

Screenshot 2024-12-02 at 17 10 55

Lets arrange forms as **Title**/**Description**, **Fields**, **Actions** (instead of _Buttons_ as such actions could just be/look like links), and **Messages** (instead of _Form Messages_) last.

m453h:
> 👍🏽 ... Definitely on the right track @m453h .
> 
> Not sure about "dynamic" fields. Don't we need to match names to what APIs expect e.g. `email` field must be called `email`, etc. If content editor renames it to emailAddress, will the form still work?
> 


Yes we need to match the names, I had conflicting thoughts about dynamic fields (I opted to use them since they reduce repetitions in the code) and had kept a validator to ensure all the expected `names` from the front-end are included, since I have a function that creates these fields, I can quickly repurpose it to have pre-defined fields and remove this validator

koechkevin (commented):
＋

koechkevin (approved):

m453h:
@kilemensi to add a bit of context, I am using [TextField](https://payloadcms.com/docs/fields/text) in the [Login](https://github.com/CodeForAfrica/ui/blob/b7a29111061ae8a477fc65cb88372baf10078e4f/apps/civicsignalblog/src/payload/globals/Forms/login/LoginTab.js), [PasswordReset](https://github.com/CodeForAfrica/ui/blob/b7a29111061ae8a477fc65cb88372baf10078e4f/apps/civicsignalblog/src/payload/globals/Forms/resetPassword/ResetPasswordTab.js) and [Registration](https://github.com/CodeForAfrica/ui/blob/b7a29111061ae8a477fc65cb88372baf10078e4f/apps/civicsignalblog/src/payload/globals/Forms/registration/RegistrationTab.js) Forms instead of [E-mail](https://payloadcms.com/docs/fields/email) because, within the CMS, we expect the user to set what would appear as labels, error messages and hints on the font-end forms thus the user would always need to enter free text, using the E-mail field wouldn't work in this scenario.

kilemensi:
Cool, cool @m453h ... got you now. The forms here are just for collecting messages (error, hints, etc.)

kilemensi:
Don't wait for me on this PR @m453h but did you [fix](https://github.com/CodeForAfrica/ui/pull/1009#pullrequestreview-2472821979) the titles e.g. Form Messages to just messages, Buttons to Actions, etc.

Screenshot 2024-12-10 at 14 02 22

Screenshot 2024-12-10 at 14 02 58

m453h:
> Don't wait for me on this PR @m453h but did you [fix](https://github.com/CodeForAfrica/ui/pull/1009#pullrequestreview-2472821979) the titles e.g. Form Messages to just messages, Buttons to Actions, etc.
> 
> Screenshot 2024-12-10 at 14 02 22 Screenshot 2024-12-10 at 14 02 58

Sure, just cross-checked, I had one commit I didn't push for the labels...I think everything else should be alright

PR #1019, opened 2024-12-14
Remove Old Commons-ui section

kelvinkipruto:
## Description

This PR remove Legacy `common-ui` dependency

## Type of change

## Screenshots
image
image

## Checklist:


apps/climatemappedafrica/src/components/Footer/index.js
```diff
@@ -18,21 +16,23 @@ function Footer(props) {
     links,
     variant,
   } = props;
-  const classes = useStyles(props);
+  const theme = useTheme();
 
   return (
     <Box
-      sx={(theme) => ({
+      sx={({ palette, typography }) => ({
```
kilemensi, line 23: Just prefer function form of `sx` instead of `useTheme`;
```jsx
sx={(theme)=>({ ... })}
```

apps/climatemappedafrica/src/components/Section/index.js
```diff
@@ -1,46 +1,30 @@
...
-      };
-    }
-    return acc;
-  }, {}),
-}));
-
-function Section({ className, fixed = true, ...props }) {
-  const classes = useStyles(props);
-
+function Section({ fixed = true, sx, ...props }) {
```
kilemensi, line 5: Do we still need custom `Section`? Ideally the `commons/ui/Section` component should be flexible/reusable enough.

packages/commons-ui-core/src/Copyright/Copyright.js
```diff
@@ -0,0 +1,44 @@
...
+    sx,
+  } = props;
+
+  return (
+    <Box sx={sx} ref={ref}>
+      {copyright && (
+        <Typography variant={variant} {...CopyrightProps}>
+          {copyright}
+        </Typography>
+      )}
```
kilemensi, line 23: 1. Use `RichTypography` (not only because it can handle HTML, but also there is no need of doing the whole `copyright &&` check.
2. I'm assuming copyright and year use the same props; Wouldn't `TypographyProps` do do the job instead of two separate `CopyrightProps` and `YearProps` props?

packages/commons-ui-core/src/Copyright/Copyright.test.js
```diff
@@ -0,0 +1,17 @@
+import { render } from "@commons-ui/testing-library";
+import React from "react";
+
+import Copyright from "./Copyright";
+
+describe("Copyright", () => {
```
kilemensi, line 6:
```suggestion
describe("<Copyright />", () => {
```

packages/commons-ui-core/src/Copyright/Copyright.js
```diff
@@ -0,0 +1,44 @@
...
+      {icon && (
+        <a href={url} target="_blank" rel="noopener noreferrer">
```
kilemensi, line 25: 1. Shouldn't you check for presence of `icon` **and** `url`?
2. Don't we already have `ImageButton` (or similiar-ish components) to handle displaying clickable images?

packages/commons-ui-core/src/QuickLinks/QuickLinks.js
```diff
@@ -0,0 +1,58 @@
+import { Box, Link, Typography } from "@mui/material";
+import React from "react";
+
+const QuickLinks = React.forwardRef(function QuickLinks(props, ref) {
+  const { linkComponent, links, LinksProps, LinkProps, title, TitleProps, sx } =
+    props;
+  const LinkComponent = linkComponent || Link;
```
kilemensi, line 7: Since you have `LinksProps`, shouldn't `linkComponent` be `LinksProps?.component`?

packages/commons-ui-core/src/QuickLinks/QuickLinks.js
```diff
@@ -0,0 +1,58 @@
...
+
+  return (
+    <Box
+      sx={{
```
kilemensi, line 11:
```suggestion
    <Box
      component="nav"
      {...props}
      sx={{
```

I think [`nav`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav) makes sense here.

packages/commons-ui-core/src/QuickLinks/QuickLinks.js
```diff
@@ -0,0 +1,58 @@
...
+      <Typography
+        {...LinksProps}
+        sx={{
+          listStyle: "none",
+          padding: 0,
+          ...LinksProps?.sx,
+        }}
+        component="ul"
+      >
+        {links.map(({ label, ...others }) => (
```
kilemensi, line 38: Have you checks that `links` is not `undefined`/`null`?

packages/commons-ui-core/src/RichTypography/RichTypography.js
```diff
@@ -9,9 +9,15 @@ const RichTypographyRoot = styled(Typography, {
   shouldForwardProp: (prop) => prop !== "LinkProps",
 })(({ LinkProps, theme }) => {
   const color = LinkProps?.color || "primary.main";
+  const linkDecoration = LinkProps?.textDecoration || "none";
+  const linkDecorationColor = LinkProps?.textDecorationColor || color;
```
kilemensi, line 13: 💯

kilemensi (commented):
👍🏽 ... haven't run it locally yet but code looks good.

---

No test for `QuickLinks`?

koechkevin (approved):

m453h (commented):
I tested locally and it looks good, there is one minor issue regarding the legend, it does slightly overlap the map, It would be better if we add a bit of space between it and the map (perhaps increase the map height and / hero component height to match the changes)

Screenshot 2025-01-10 at 10 57 19

apps/climatemappedafrica/src/components/Section/index.js
kelvinkipruto: @kilemensi yes. We have to give each `Section` a fixed width. Rather than do so at each place it is used, having a custom one is much better and cleaner.

kilemensi: 🤔 ... I just glanced at the code @kelvinkipruto and could be wrong but isn't the [code](https://github.com/CodeForAfrica/ui/blob/9c25d295feac00d325473af7d6d9b4347bef879e/apps/climatemappedafrica/src/components/Section/index.js#L11-L20) in this PR exactly like the one in [`commons-ui/core/Section`](https://github.com/CodeForAfrica/ui/blob/41f5cdc0cb3e7849457bdfc297b47a2c91493ee3/packages/commons-ui-core/src/Section/Section.js#L16-L28) already? Of course, `theme.widths` was renamed to `theme.contentWidths`.

kelvinkipruto: @kilemensi The main issue is the width. If we use the `commons-ui/core` `section`, it is wider than the original.

after

kilemensi: Cool @kelvinkipruto ... to confirm did you update the [theme](https://github.com/CodeForAfrica/ui/blob/41f5cdc0cb3e7849457bdfc297b47a2c91493ee3/apps/climatemappedafrica/src/theme/index.js#L133-L138) to:
```js
  contentWidths: {
    values: {
      md: 608, // 0, 80, 0, 80 margin
      lg: 1160, // 0, 140, 0, 140 margin
    },
  },
```

kelvinkipruto: This works. Thanks @kilemensi.

kilemensi (approved):
LGTM (provided tests are fixed).

PR #1035, opened 2025-01-24
Unify Google Analytics

### LINKED ISSUE #1034: Correct PesaYetu and ClimateMapped Google Analytics ID
**Describe the bug**
In PesaYetu and ClimateMapped, Google Analytics is configured to use the environment variable [`NEXT_PUBLIC_GOOGLE_ANALYTICS_ID`](https://github.com/CodeForAfrica/ui/blob/8c57d15fdd9edd511266224ac88e29c1e93ed1e5/apps/pesayetu/src/lib/ga/index.js#L4) while in the deployment configuration, the environment variable is set as `NEXT_PUBLIC_GOOGLE_ANALYTICS`.
 
We need to update the code to reference the correct environment variable.

kelvinkipruto:
## Description
This PR unifies how we're handling Google Analytics in all apps

1. For applications that have cms, the Google Analytics ID will be set in the CMS under the `Engagement` tab.
2. For those that do not, we will use an env variable `NEXT_PUBLIC_GOOGLE_ANALYTICS_ID`

Fixes #1034

## Type of change

## Checklist:


apps/charterafrica/.env
```diff
@@ -6,3 +6,4 @@ PAYLOAD_PUBLIC_LOCALES="en, fr, pt"
 SENTRY_ENVIRONMENT=local
 SENTRY_ORG=code-for-africa
 SENTRY_PROJECT=charterafrica
+NEXT_PUBLIC_GOOGLE_ANALYTICS_ID="G-5ZPEFLT0NJ"
```
kilemensi, line 9: 🤔 ... If we set the id here in `.env`, doesn't it mean even when running in dev/local, analytics events will be sent to Google?

We only want to track PROD events.

kilemensi (commented):
👍🏽 

---

1. In docker, if we don't include GA id at build time, can we change it at run time e.g. Dokku config?
2. Do we need to update app scripts/config in vault?

koechkevin:
@kelvinkipruto I found this error trying to run charterafrica
image

koechkevin:
> @kelvinkipruto I found this error trying to run charterafrica image

Ignore this, I hadn't rerun pnpm i

apps/charterafrica/.env
kelvinkipruto: > If we set the id here in `.env`, doesn't it mean even when running in dev/local, analytics events will be sent to Google?

I have been testing it locally over the past few days, and it looks like it does not track localhost.

kilemensi: Really now? Just tested it locally now and in 2 minutes I see my events in the realtime dashboard.
Screenshot 2025-01-29 at 09 20 44

koechkevin (approved):

apps/charterafrica/.env
kilemensi: I still see these GA in .env @kelvinkipruto . Why are you saying you've fixed it?

kelvinkipruto: @kilemensi I created a reusable [Google Analytics](https://github.com/CodeForAfrica/ui/blob/46a47fc263813e5303bca6fee99e1d73bcf3b5b6/packages/commons-ui-next/src/GoogleAnalytics/index.js) component that only works in prod environment

kelvinkipruto: Removed the hardcoded GA env variable

kilemensi (approved):
👍🏽 ... haven't run it locally but code looks solid.

PR #1043, opened 2025-02-04
Fix Jest Memory leaks

kelvinkipruto:
## Description

Our current Jest setup has a memory leak. You can check it using `pnpm run jest -- --detectLeaks`. 
This is not an isolated case based on this open [issue](https://github.com/jestjs/jest/issues/7874)
This causes performance issues. in my case, running the `jest` command has used up to 20GB of RAM and even caused my machine to shut down when generating new snapshots.

This PR attempts to fix that by doing garbage collection after each test run. It also limits the max RAM usage to 4GB among other improvements.

## Type of change

## Checklist:

### DIFF
--- a/packages/commons-ui-testing-library/force-gc.js
+++ b/packages/commons-ui-testing-library/force-gc.js
@@ -1,0 +1,6 @@
+/* eslint-env jest */
+/* eslint-disable no-undef */
+
+afterEach(() => {
+  if (global.gc) global.gc();
+});
--- a/apps/charterafrica/package.json
+++ b/apps/charterafrica/package.json
@@ -29,7 +29,7 @@
     "start": "PAYLOAD_CONFIG_PATH=${PAYLOAD_CONFIG_PATH:-dist/payload.config.js} NODE_ENV=${NODE_ENV:-production} node dist/server.js",
     "lint-check": "TIMING=1 eslint './'",
     "lint": "TIMING=1 eslint --flag unstable_config_lookup_from_file --fix './'",
-    "jest": "jest",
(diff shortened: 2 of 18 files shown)


kilemensi:
1. Instead of `node --expose-gc --max-old-space-size=4096 node_modules/jest/bin/jest.js`, doesn't `NODE_OPTIONS="--expose-gc --max-old-space-size=4096" jest` work? I think we should avoid hard-coding paths in `node_modules`, etc.
2. Can't we add this as as script in the project root dir i.e.`<project root>/scripts/jest.sh` that every package.json in packages/apps refers to rather than copy/pasting the full command in every package/app?

packages/commons-ui-testing-library/force-gc.js
```diff
@@ -0,0 +1,6 @@
+/* eslint-env jest */
+/* eslint-disable no-undef */
```
koechkevin, line 2: We can just ignore this file in eslint

m453h:
🚀 

I tested locally and it does significantly reduce the load on my machine I did notice this warning related to  [--forceExit](https://jestjs.io/docs/cli#--forceexit), 
```
Force exiting Jest: Have you considered using `--detectOpenHandles` to detect async operations that kept running after all tests finished?
```
I haven't played much with these cli options but I was a bit concerned with using `--forceexit`,  I saw these [combinations](https://github.com/jestjs/jest/issues/7874#issuecomment-2312758436) of flags which are pretty close to what we currently have, I did notice when we use [--runInBand](https://jestjs.io/docs/cli#--runinband) the maximum heap does tend to increase significantly

kilemensi:
My comment from [Feb 4](https://github.com/CodeForAfrica/ui/pull/1043#issuecomment-2633937511) still applies @kelvinkipruto .

PR #1039, opened 2025-02-03
ClimateMappedAfrica: User unable to click the search field

koechkevin:
## Description

currently, a user is unable to click input to search a location. This is caused by drawer backdrop which prevents actions on other visible elements even after setting  a greater zIndex.
This PR changes drawer to persistent and updates styles to maintain how it was.

Fixes #1038

## Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation

PR #1051, opened 2025-02-21
ClimateMapped Africa Fixes

kelvinkipruto:
## Description

This PR fixes several bugs in ClimateMapped.
1. Search not working due to each list item being rendered as a link instead of a button.
2. Revert the chart tooltip to its original implementation
3. Fix a bug where you could not navigate using the Location Tags
4. Fix an issue where the explore page drawer was always persistent, hence "covering" other elements.

Fixes # (issue)

## Type of change

## Screenshots
image

## Checklist:

### DIFF
--- a/apps/climatemappedafrica/src/components/DropdownSearch/DownloadSearch.js
+++ b/apps/climatemappedafrica/src/components/DropdownSearch/DownloadSearch.js
@@ -6,13 +6,13 @@ import {
   ListItem,
   SvgIcon,
   Box,
+  Button,
 } from "@mui/material";
 import { useRouter } from "next/router";
 import PropTypes from "prop-types";
 import React, { useEffect, useState } from "react";
 
 import SearchIcon from "@/climatemappedafrica/assets/icons/search.svg";
-import Link from "@/climatemappedafrica/components/Link";
 
 function DropdownSearch({
   IconButtonProps,
@@ -174,7 +174,7 @@ function DropdownSearch({
           >
             {suggestions.map(({ name, code }) => (
               <ListItem
-                component={Link}
+                component={Button}
                 variant="subtitle1"
                 underline="none"
                 onClick={() => handleSelect(code, name)}
--- a/apps/climatemappedafrica/src/components/DropdownSearch/DownloadSearch.js
+++ b/apps/climatemappedafrica/src/components/DropdownSearch/DownloadSearch.js
@@ -6,13 +6,13 @@ import {
   ListItem,
   SvgIcon,
   Box,
-  Button,
 } from "@mui/material";
 import { useRouter } from "next/router";
 import PropTypes from "prop-types";
 import React, { useEffect, useState } from "react";
 
 import SearchIcon from "@/climatemappedafrica/assets/icons/search.svg";
+import Link from "@/climatemappedafrica/components/Link";
 
 function DropdownSearch({
   IconButtonProps,
@@ -174,7 +174,8 @@ function DropdownSearch({
           >
             {suggestions.map(({ name, code }) => (
               <ListItem
-                component={Button}
+                component={Link}
+                href={`${hrefProp}/${code}`}
                 variant="subtitle1"
                 underline="none"
                 onClick={() => handleSelect(code, name)}
--- a/apps/climatemappedafrica/src/components/HURUmap/Panel/DesktopPanel/DesktopPanel.js
+++ b/apps/climatemappedafrica/src/components/HURUmap/Panel/DesktopPanel/DesktopPanel.js
@@ -47,7 +47,7 @@ function DesktopPanel({ sx, ...props }) {
             overscrollBehaviorBlock: "none",
           },
         }}
-        variant="persistent"
+        {...(open && { variant: "persistent" })}
         PaperProps={{
           elevation: 0,
           square: true,
--- a/apps/climatemappedafrica/src/components/ExplorePage/index.js
+++ b/apps/climatemappedafrica/src/components/ExplorePage/index.js
@@ -82,7 +82,7 @@ function ExplorePage({
     if (geoCode) {
       dispatch({
         type: "fetch",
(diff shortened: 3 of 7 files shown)


koechkevin (commented):
@kelvinkipruto Good work.
Not sure if it is related to my setup here but I encountered this error on clicking Nairobi from Kenya map.
I don't think this is related to this bug fix. I belive this will be resolved once we move to server rendering all pages.

apps/climatemappedafrica/src/components/HURUmap/Panel/DesktopPanel/DesktopPanel.js
```diff
@@ -47,7 +47,7 @@ function DesktopPanel({ sx, ...props }) {
             overscrollBehaviorBlock: "none",
           },
         }}
-        variant="persistent"
+        {...(open && { variant: "persistent" })}
```
kilemensi, line 50: I remember asking about the [this](https://github.com/CodeForAfrica/ui/pull/1039#discussion_r1940563033) change?

kilemensi (commented):
Haven't tried running the code locally @kelvinkipruto  but:

1. Same of these code changes seem to be different from PesaYetu code e.g. The `LinkItem` and `handleClickMap` changes. Are they also bugs that need fixing in PesaYetu.
2. Search and clicking on the map are also broken on the homepage. Are they also addressed in this PR?

apps/climatemappedafrica/src/components/HURUmap/Panel/DesktopPanel/DesktopPanel.js
kelvinkipruto: @kilemensi 
1. We only have to update `handleClickMap` in Pesayetu. The reason is that Pesayetu uses the legacy `commons-ui`, hence no need to update the `LinkItem`  
2. It looks like the reason searching and clicking on the map is not working is a data issue: looking into it.

kilemensi: And isn't link [`Link`](https://github.com/CodeForAfrica/ui/blob/main/apps/pesayetu/src/components/Link/index.js) included @kelvinkipruto ?

kelvinkipruto: > I remember asking about the [this](https://github.com/CodeForAfrica/ui/pull/1039#discussion_r1940563033) change?

@kilemensi Yes. This approach allows the search to work when the drawer is open, but we only need it to be persistent if the drawer is open. If we set it to persistent all the time, it will overlay over some elements, like the location tags, making them unclickable.

kelvinkipruto: > And isn't link [`Link`](https://github.com/CodeForAfrica/ui/blob/main/apps/pesayetu/src/components/Link/index.js?rgh-link-date=2025-02-24T13%3A26%3A08Z) included @kelvinkipruto ?

@kilemensi Yes. Pesayetu's `Link` uses the legacy [`commons-ui`](https://github.com/CodeForAfrica/ui/blob/1e0748c05b3d357ca95481349d984fc8274ddc2e/apps/pesayetu/src/components/Link/index.js#L1) which has no issue. ClimateMapped uses the new [`Link`](https://github.com/CodeForAfrica/ui/blob/1e0748c05b3d357ca95481349d984fc8274ddc2e/apps/climatemappedafrica/src/components/Link/index.js#L1) from `commons-ui/next` which always requires a `href` to be passed to it. In our case, we're not passing an `href` which causes it to break. Since we're using the `onClick` event we instead use a button.

kilemensi: What am asking/saying @kelvinkipruto is:

Is it **us** or `mui` that requires `href` on `Link`? If it's us, that should be an easy fix to require either `href` or `onClick`. There is no need to say it's legacy `commons-ui`, etc. If it's not us, which is what I'm suspecting, then yes, lets document that `@mui/material/Link` _now_ requires `href`.

---

Even if it is indeed `mui`, can't we pass _fragment_ as `href` e.g. `/currentURL#new-code` since fragments don't cause server-side navigation? Then `onClick` should work fine.

koechkevin (approved):
@kelvinkipruto this looks good now but I think in the long run we need to find a way of adding a loader for better experience. For instance Accessing this [link](http://localhost:3000/explore/ke.nb), it initially shows a blank page before populating data leading to a poor experience. Maybe add a separate ticket for this?

m453h (approved):
Got it running locally 🚀 

---

Since we're reverting the chart tooltip to its original implementation, should we open a separate issue to track the tooltip flickering?

Just one minor issue I observed that could be nice to fix in this PR,  I noticed that the text on the `DropdownSearch` is not being capitalized when we select an item we are searching for,  perhaps we could look into how we are targeting the `&.MuiInputBase-input` and applying the styles ?

Screenshot 2025-02-25 at 12 07 12

kelvinkipruto:
created #1053 to track the tooltip issue

apps/climatemappedafrica/src/components/DropdownSearch/DownloadSearch.js
```diff
@@ -175,6 +175,7 @@ function DropdownSearch({
             {suggestions.map(({ name, code }) => (
               <ListItem
                 component={Link}
+                href={`${hrefProp}/${code}`}
```
kilemensi, line 178: If we're using `href=${hrefProp}/${code` then we don't need `onClick` anymore because the page will navigate and we already have server-side rendering.

If we're not doing the necessary clean-up in this PR, lets and an issue to track it

kilemensi (approved):
LGTM!

Issue #1055, opened 2025-02-25
Clean up navigation to ensure full server side navigation


kelvinkipruto:
In some sections of the code, we use `router.push` to navigate to other pages. Since the site will now be fuller server-rendered, we need to change them to `href`

PR #1056, opened 2025-02-27
ClimateMap SSG

### LINKED ISSUE #1054: ClimateMap - Render ClimateMap Pages from the Server side

kelvinkipruto:
## Description

This PR improves how we handle server-side generation.

1. We now pre-render all top-level geographies (Continent and Country).
2. Introduces an error page to handle 404 and 500 errors. It also handles errors on the `/explore` page.
3. Improve the fallback page by showing a progress indicator
4. Cache all responses from the server to save load speed. By default, we cache for 10 mins.
5. Adds a progress bar to show route transitions

Fixes #1054
#1055 

## Type of change

## Screenshots

image

[screen-capture.webm](https://github.com/user-attachments/assets/0dce2895-55d0-45e6-8b1c-9a2fa1e85f4b)

image

## Checklist:


apps/climatemappedafrica/src/pages/[[...slugs]].js
```diff
@@ -14,10 +14,7 @@ import Tutorial from "@/climatemappedafrica/components/HURUmap/Tutorial";
 import Navigation from "@/climatemappedafrica/components/Navigation";
 import PageHero from "@/climatemappedafrica/components/PageHero";
 import Summary from "@/climatemappedafrica/components/Summary";
-import {
-  getPageStaticPaths,
-  getPageStaticProps,
-} from "@/climatemappedafrica/lib/data";
+import { getPageStaticProps } from "@/climatemappedafrica/lib/data";
```
Comment (koechkevin) [inline apps/climatemappedafrica/src/pages/[[...slugs]].js:17]: We can as well rename this function🤔 @kelvinkipruto

apps/climatemappedafrica/src/utils/fetchJson.js
```diff
@@ -1,5 +1,28 @@
+const cache = new Map();
```
koechkevin, line 1: We may need to rethink a better caching strategy on the UI. I don't think this strategy is best. Maybe track in an issue?

kelvinkipruto: #1066

kilemensi:
Any updates on this @kelvinkipruto ?

kelvinkipruto:
@kilemensi @koechkevin Your reviews here please.

apps/climatemappedafrica/src/lib/data/blockify/hero.js
```diff
@@ -18,7 +15,7 @@ export default async function hero(block, _api, _context, { hurumap }) {
     rootGeography: { center, code, hasData: pinRootGeography, zoom },
     profile: hurumapProfile,
   } = hurumap ?? { rootGeography: {} };
-  const { geometries } = await fetchProfileGeography(code.toLowerCase(), {
+  const { geometries } = await fetchCachedProfileGeography(code.toLowerCase(), {
```
kilemensi, line 18: Looks like this `fetchCachedProfileGeography` call has to be made for each `code`, twice if/when we're doing comparison (unlike the profile call which is the same for the whole app).
1. Doesn't this mess up the caching plan?
2. Isn't there a way of fetching for many codes in one call?

kelvinkipruto: @kilemensi There are around 815 geographies, but this looks to increase as some countries currently do not have regions. The only page we're building is `explore/af`. I tried building for all 55 countries, but the API became unresponsive due to the many requests.

kelvinkipruto: @kilemensi We may also have to rethink how to optimize it further, for example fetching data for [`ke`](https://climatemapped-africa.hurumap.org/api/v1/all_details/profile/1/geography/KE/?version=Climate) returns a response that is 2.6MB. In most countries I have tested, the response size is more than 2MB. This also extends to regions which also have ~2MB of response data each.

kilemensi: Lets create /explore, /explore/af and 5 - 10 core HURUmap countries @kelvinkipruto; the rest will be created on demand.

If the initial page rendered for a non-built page is good enough all good, otherwise, we may have to change our [`fallback`](https://nextjs.org/docs/pages/building-your-application/data-fetching/get-static-props#when-does-getstaticprops-run) policy to `blocking`.

>We may also have to rethink how to optimize it further, for example fetching data for [ke](https://climatemapped-africa.hurumap.org/api/v1/all_details/profile/1/geography/KE/?version=Climate) returns a response that is 2.6MB...
>
BE optimisation will be done down the line. On FE, we don't need to pass on the whole 2.6MB, since it's a JSON, you can check and return only the props/attributes we need.

kilemensi:
How does an un-built page look like on an initial load @kelvinkipruto ? Hopeful not like 👇🏽 

Screenshot 2025-03-14 at 10 22 49

kelvinkipruto:
> How does an un-built page look like on an initial load

@kilemensi, I believe the URL is wrong. That is the wrong code for Swaziland (Eswatini). We updated it to [`SZ`](https://climatemapped-africa.dev.codeforafrica.org/explore/sz). 
This, however, brings up an edge case we've not addressed. Currently, if the geography does not exist, it just shows a blank screen instead of the 404 page.
I have also encountered a similar page, but only when reloading an `/explore/xyz` page during development.
Let me handle the two before merging this.

kilemensi:
>@kilemensi, I believe the URL is wrong. ...
>

That's not the (only) reason @kelvinkipruto; when using SSG, if there isn't an existing page (whether it should be there or not) and `fallback: true`, Next will return an initial ["fallback"](https://nextjs.org/docs/pages/api-reference/functions/get-static-paths#fallback-true) version of a page. That's what this is. e.g. 👇🏽  is for the first time I load `ke.nb` _after_ a local build i.e. only when countries and continent pages are pre-built.

ke nb

kelvinkipruto:
> when using SSG, if there isn't an existing page (whether it should be there or not) and `fallback: true`, Next will return an initial ["fallback"](https://nextjs.org/docs/pages/api-reference/functions/get-static-paths#fallback-true) version of a page.

Thanks for this explanation 🤯. I have updated the code to improve the pre-rendered page to show a progress indicator.

kilemensi:
Why are we doing an 404 error page just for the HURUmap pages instead of the whole site @kelvinkipruto ?

apps/climatemappedafrica/src/lib/hurumap/index.js
```diff
@@ -160,3 +163,84 @@ export async function fetchProfileGeography(
...
+  const cacheDir = path.resolve(process.cwd(), "public", "cache");
+  const cacheFile = path.join(
+    cacheDir,
+    `geography-${profileId}-${geoCode.toUpperCase()}-${version}.json`,
+  );
+
+  try {
+    await fs.mkdir(cacheDir, { recursive: true });
+
+    try {
```
koechkevin, line 218: I think separating this to a different function would be ideal @kelvinkipruto nesting try catch blocks is a symptom of code smells .

apps/climatemappedafrica/src/lib/hurumap/index.js
```diff
@@ -160,3 +163,96 @@ export async function fetchProfileGeography(
...
+      }
+    }
+
+    const profileData = await fetchProfile(params);
+    const cacheContent = JSON.stringify({
+      timestamp: Date.now(),
+      data: profileData,
+    });
+
+    await fs.writeFile(cacheFile, cacheContent, "utf-8");
```
m453h, line 202: Minor, Issue but since I/O is expensive could we check if the newly fetched data has really changed before we write to the file ? (Maybe compare hashes) 🤔 , maybe something that could be worth considering is how to handle race conditions, It could be I'm worried about something very improbable, but looking at the code there is a chance of the cache being corrupted

m453h (commented):
@kelvinkipruto  this looks good  🚀 there are a few things that I observed after running it locally:

1. For non HURUMap pages the 404 page doesn't render the footer or page header correctly
Screenshot 2025-03-19 at 12 22 35

I think we should try to go for something similar to what we have on `codeforafrica`:

Screenshot 2025-03-19 at 12 26 30

2. Is it possible for the fallback page to include the navigation bar as we show the progress indicator ?

https://github.com/user-attachments/assets/480921aa-dd57-46f6-9535-8b0c0ee853e6

3. Since we have added a progress indicator, we should find a way to show the user something is loading when they search for something that requires fetching the profile data, perhaps we can track this in a separate issue.

kelvinkipruto:
Thanks @m453h 
> For non HURUMap pages the 404 page doesn't render the footer or page header correctly
I can see the on my side. You need to set them up in Payload.
> Is it possible for the fallback page to include the navigation bar as we show the progress indicator ?
No. Because we're fetching the navigation data during run time, they resolve with the other data required to render a page. The fallback page is shown when we visit a page which was not generated by `getStaticPaths`.
>Since we have added a progress indicator, we should find a way to show the user something is loading when they search for something that requires fetching the profile data, perhaps we can track this in a separate issue.
I think this is a good idea. Let me implement it.

apps/climatemappedafrica/package.json
```diff
@@ -34,6 +34,7 @@
   },
   "dependencies": {
     "@apollo/client": "catalog:",
+    "@badrap/bar-of-progress": "catalog:",
```
m453h, line 37: Nice idea keeping the progress indicator at the top, just wondering if the [Linear Progress indicator](https://mui.com/material-ui/react-progress/?srsltid=AfmBOooQPLjRHLAtTxbq6uZfVQTSvixPN-dbS6SxEIHfC3i3_QVI_pFu#linear) can't do the same this library does ?

m453h (approved):

PR #1070, opened 2025-03-10
Built-time Image src Resolution

### LINKED ISSUE #1068: Built-time Image `src` Resolution
**Description**

Images on custom error pages (e.g., 404.tsx, 500.tsx) are not loading in production because Next.js encodes image URLs with localhost:3000 at build time. This results in broken images when deployed to a different environment.

**Steps to Reproduce**

1. Navigate to non existent page on [Charter Africa](https://charter.africa/sd) or [Code for Africa](https://cfa.dev.codeforafrica.org/c).
2. Inspect the image source in DevTools; it appears as `http://localhost:3000/media/cfa-logo.svg` instead of `/media/cfa-logo.svg`

**Expected Behavior**
Image URLs should remain relative (e.g., /images/error.png) so they work across different environments.
**Actual Behavior**
Image URLs are hardcoded with localhost:3000, causing them to break outside of a local environment.

koechkevin:
## Description

Images on custom error pages (e.g., 404.tsx, 500.tsx) are not loading in production because Next.js encodes image URLs with localhost:3000 at build time. This results in broken images when deployed to a different environment.

Fixes #1068

## Type of change

## Checklist:


apps/charterafrica/src/payload/collections/Media.js
```diff
@@ -24,6 +24,12 @@ const Media = {
       ({ doc }) => ({
         ...doc,
         alt: doc.alt ?? null,
+        /*
+        Since we are setting ENV PAYLOAD_PUBLIC_APP_URL=http://localhost:3000 at build time, generating static pages(404,500) ends up with image urls with `http://localhost:3000/media/cfalogobw.svg`.
+        In this proposal, we are just returning pathnames without hostname since our next js apps and payload are served from the same instance.
+        */
+        src: new URL(doc.url).pathname,
+        url: new URL(doc.url).pathname,
       }),
```
kilemensi, line 33: No need for two `new URL` calls; just create a proper function with `return`.
```suggestion
      ({ doc }) => {
        // Can't use absolute URLs in src because we set PAYLOAD_PUBLIC_APP_URL to localhost at build-time & will lead to 404 at runtime
        // Solution: use relative URLs
        const { pathname } = new URL(doc.url);
        return {
          ...doc,
          alt: doc.alt ?? null,
          src: pathname,
          url: pathname,
        }
      },
```

PS: This could even move to somewhere reusable since I assume it's something we're going to apply to all apps?

docker-compose.yml
```diff
@@ -11,6 +11,7 @@ services:
         - SENTRY_ENVIRONMENT
         - SENTRY_ORG
         - SENTRY_PROJECT
+        - PAYLOAD_PUBLIC_APP_URL
```
kilemensi, line 14: `docker-compose` is local dev only & I don't think we use 2 separate domain when doing local development

kilemensi (approved):
🚀 

--

Haven't tried it locally but the solution looks simple and clear to me!

apps/charterafrica/src/payload/collections/Media.js
koechkevin: > PS: This could even move to somewhere reusable since I assume it's something we're going to apply to all apps?
@kilemensi 
We can't somehow move this to packages, it tries to resolve other files with jsx syntax when you import in payload related files.
image

m453h (approved):
LGTM  🚀

packages/commons-ui-core/src/utils/transformDocMediaUrl.js
```diff
@@ -0,0 +1,13 @@
+const transformDocMediaUrl = (doc) => {
```
kilemensi, line 1: `payload`-related functionality should **not** be in `@commons-ui/core` or `@commons-ui/next` packages @koechkevin .

kilemensi:
If this works @koechkevin, then lets merge it as is. Create an issue (if it doesn't exist) to consolidate all payload-related functionality into one package that will be tackled in a different PR.

PR #1075, opened 2025-03-24
Enhance Build Reliability (ClimateMap)

kelvinkipruto:
## Description

During the static build process of ClimateMap, approximately 57 pages need to be pre-rendered, each requiring data from the HURUMap API. To address intermittent build failures caused by API response delays, this PR implements two key improvements:

1. Increases the static page generation timeout from 60s to 180s to accommodate longer API response times
2. Implements a retry mechanism with exponential backoff for failed API requests, making the build process more resilient. If all retries are exhausted, we provide fallback content.

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/apps/climatemappedafrica/src/pages/[[...slugs]].js
+++ b/apps/climatemappedafrica/src/pages/[[...slugs]].js
@@ -79,12 +79,34 @@ function Page({ blocks = [], fallback }) {
   );
 }
 
+async function fetchWithRetry(context, attempt = 1, maxAttempts = 5) {
+  try {
+    const result = await getPageStaticProps(context);
+    return result;
+  } catch (error) {
+    if (attempt === maxAttempts) {
+      return {
+        props: {
+          blocks: [],
+          fallback: {},
+        },
+        revalidate: 60,
+      };
+    }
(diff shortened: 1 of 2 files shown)


apps/climatemappedafrica/src/pages/[[...slugs]].js
```diff
@@ -79,12 +79,34 @@ function Page({ blocks = [], fallback }) {
   );
 }
 
+async function fetchWithRetry(context, attempt = 1, maxAttempts = 5) {
+  try {
+    const result = await getPageStaticProps(context);
+    return result;
+  } catch (error) {
```
Comment (kilemensi) [inline apps/climatemappedafrica/src/pages/[[...slugs]].js:86]: 🤔 ... isn't it better to retry a failing fetch call (I'm assuming probably the one fetching a geometry) rather than retrying the whole page build?

apps/climatemappedafrica/src/pages/[[...slugs]].js
```diff
@@ -79,12 +79,34 @@ function Page({ blocks = [], fallback }) {
...
+    if (attempt === maxAttempts) {
```
Comment (koechkevin) [inline apps/climatemappedafrica/src/pages/[[...slugs]].js:87]: I'd use `>==` here to be on the safer side.

kilemensi (approved):

PR #1122, opened 2025-04-25
@/RoboShield: Migrate rich text editor from Slate to Lexical

### LINKED ISSUE #1120: Migrate RoboShield Rich Text Editor from Slate to Lexical Editor
After upgrading `RoboShield` to v3, we should migrate the CMS Rich Text Editor from the legacy Slate Editor to the Lexical Editor.

m453h:
## Description

This PR migrates the Rich Text editor from Slate to Lexical used in Payload. While both Slate and Lexical save the editor state in JSON, the structure of the JSON is different, thus requiring a migration to convert all previously saved Rich Text content to the new format.

Fixes #1120

## Type of change

## Screenshots
Screenshot 2025-04-25 at 12 05 38
Screenshot 2025-04-25 at 12 06 08

## Checklist:


apps/roboshield/migrations/20250425_082829_migration.ts
kilemensi, line 1: Lets make the file names a bit more descriptive e.g. `20250425_082829_slate_to_lexical.ts`

apps/roboshield/payload.config.ts
```diff
@@ -42,7 +43,12 @@ const smtpHost = process.env.SMTP_HOST || "smtp.sendgrid.net";
 
 export default buildConfig({
   serverURL: appURL,
-  editor: slateEditor({}),
+  editor: lexicalEditor({
+    features: ({ defaultFeatures }) => [
+      ...defaultFeatures,
+      SlateToLexicalFeature({ disableHooks: true }),
```
kilemensi, line 49: Why do we need this?

apps/roboshield/src/components/RichText/index.tsx
kilemensi, line 1: Aren't we sticking with the standard of `components/Component.{js|tsx},index.{js|ts}`?

apps/roboshield/src/components/RichText/styleConverter.tsx
```diff
@@ -0,0 +1,48 @@
+import { JSXConverters } from "@payloadcms/richtext-lexical/react";
+import { RichTypography, Link } from "@commons-ui/next";
+
+const DEFAULT_PROPS = {
+  html: false,
+};
+
+export const styleConverter = (converterProps: any): JSXConverters<any> => ({
+  heading: ({ node, nodesToJSX }) => {
+    const Tag = node.tag || "h1";
```
kilemensi, line 10: Why are we defaulting to `h1`? Shouldn't we just leave it up to the default component in `RichTypography`?

apps/roboshield/src/components/RichText/styleConverter.tsx
```diff
@@ -0,0 +1,48 @@
...
+    );
+  },
+  quote: ({ node, nodesToJSX }) => (
+    <blockquote>{nodesToJSX({ nodes: node.children })}</blockquote>
+  ),
+  link: ({ node, nodesToJSX }) => {
+    return (
+      <RichTypography
+        component={Link}
+        href={node.fields.url || node.href || ""}
```
kilemensi, line 28: There is no standard prop for href?

apps/roboshield/src/components/RichText/styleConverter.tsx
```diff
@@ -0,0 +1,48 @@
...
+        {nodesToJSX({ nodes: node.children })}
+      </RichTypography>
+    );
+  },
+  paragraph: ({ node, nodesToJSX }) => {
+    return (
+      <RichTypography
+        {...node?.typographyProps}
+        {...DEFAULT_PROPS}
+        {...converterProps.typographyProps}
```
kilemensi, line 42: We can create:
```jsx
const props = { ...node?.typographyProps, ...DEFAULT_PROPS, ...converterProps.typographyProps };
```
Somewhere at the top of the function instead of repeating them on each match.

PS: Is it `node`, `DEFAULT` and then `converterProps` **or** `node`, `converterProps` and then `DEFAULT`?

kilemensi (commented):
👍🏽

apps/roboshield/payload.config.ts
m453h: We need to add `SlateToLexicalFeature` to enable automatic preview of slate content in the admin panel. As per the [migration instructions](https://payloadcms.com/docs/rich-text/migration#migration-via-migration-script-recommended) we are supposed to disable hooks when we run the migration

m453h:
> @m453h this looks good, I have tested locally. Something small I noticed on Hero description, every text is bold meaning any bold styling doesn't get applied to the output. image

@koechkevin , this was deliberately implemented to use a fixed font variant i.e. `h6`. Made an adjustment to remove this since we can change the font variant on the admin panel. It should be good now

apps/roboshield/payload.config.ts
kilemensi: Thanks @m453h . Is there a way of us to know whether we're supposed to remove that code after the migration or not?

apps/roboshield/src/components/RichText/styleConverter.tsx
m453h: @kilemensi I did a couple of tests and realized we actually don't need this.  I struggled to parse internal links and its best we don't add a case to intercept how we format them i.e. the default behaviour of the default Convertor works consistently with what we have on prod

kilemensi: Then move it to an issue we can track separately @m453h (as we may need to be able to style links separately but it's not needed _today_).

apps/roboshield/payload.config.ts
kilemensi: ... cool, then lets remove it @m453h

apps/roboshield/src/components/RichText/styleConverter.tsx
m453h: I think as we increase the order of specificity it should be `DEFAULT`, `converterProps`, `node`

apps/roboshield/payload.config.ts
m453h: I'll keep a TODO note and we'll remove after we migrate PROD db

koechkevin (approved):

kelvinkipruto (approved):
LGTM

PR #1152, opened 2025-05-14
Improve `@/trustlaba` Project Setup

kilemensi:
## Description

- [x] Ensure there is valid Email config. Fixes `WARN: No email adapter provided. Email will be written to console. More info at https://payloadcms.com/docs/email/overview.`
- [x] Fix env var used for default locale
- [x] Ensures working `pageExtensions` in Next.js configuration
- [x] Small fixes around dependencies, turbo, etc.

## Type of change

- [x] Chore

## Screenshots

N/A

## Checklist:


koechkevin (approved):
LGTM, this fixes an issue I was facing here locally 🚀

apps/trustlab/package.json
```diff
@@ -19,16 +19,17 @@
...
     "@payloadcms/plugin-cloud-storage": "catalog:payload-v3",
-    "@payloadcms/richtext-lexical": "catalog:payload-v3",
-    "@payloadcms/storage-s3": "catalog:payload-v3",
     "@payloadcms/plugin-seo": "catalog:payload-v3",
     "@payloadcms/plugin-nested-docs": "catalog:payload-v3",
+    "@payloadcms/richtext-lexical": "catalog:payload-v3",
+    "@payloadcms/storage-s3": "catalog:payload-v3",
     "@payloadcms/ui": "catalog:payload-v3",
     "cross-env": "catalog:",
+    "graphql": "catalog:payload-v3",
```
kelvinkipruto, line 32: @kilemensi The reason for adding this was because we do not need graphql. If we do need the package, we should then disbale it in the config 
```
  graphQL:{
    disable: true,
  },
```

kelvinkipruto (approved):

apps/trustlab/package.json
kilemensi: Whether we use it or not @kelvinkipruto , Payload [**_expects_**](https://github.com/payloadcms/payload/blob/98283ca18c1e68c44249ee7c64367b61d98d49db/packages/next/package.json#L126-L127) that you'll install it.

Yes, disabling it (via a build time arg or env var) is probably the right approach. Lets add it to lower priority issues/tasks we need to do when we're about to go live.

PR #1159, opened 2025-05-20
Fix failing `@/trustlab` build

kilemensi:
## Description

`@svgr/webpack` dev dependency is missing causing pnpm build to fail. This PR fixes that as well as:

 - [x] Add `image` name to `trustlab` service in compose file
 - [x] Pass `@/trustlab`'s `.env.local` file to docker compose command so that env vars are available to compose.

## Type of change

- [x] Chore

## Screenshots

N/A

## Checklist:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -201,6 +201,7 @@ services:
       - path: ./apps/trustlab/.env
       - path: ./apps/trustlab/.env.local
         required: false
+    image: codeforafrica/trustlab-ui:latest
     ports:
       - 3000:3000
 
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -201,6 +201,7 @@ services:
       - path: ./apps/trustlab/.env
       - path: ./apps/trustlab/.env.local
         required: false
+    image: codeforafrica/trustlab-ui:latest
```
koechkevin, line 204: @kilemensi Doesn't this make `make trustlab` command test an already published image when it is supposed to build an image locally? I thought docker-compose is only for dev purposes.

kelvinkipruto (approved):
LGTM, apart from the issue raised by @koechkevin

docker-compose.yml
kilemensi: [Not necessary](https://docs.docker.com/reference/compose-file/build/#using-build-and-image) @koechkevin ... we can also include [`pull_policy`](https://docs.docker.com/reference/compose-file/services/#pull_policy) to build every time.

Issue #1136, opened 2025-05-07
Roles & Access Control Creation


RalphOwah:
Roles & Access Control Creation

- [ ] Admin – Full access to all collections and settings
- [ ] Editor – Can create/edit content and publish settings
- [ ] Reviewer - Can create/edit content BUT CANNOT PUBLISH.
- [ ] Viewer – Read-only access (e.g. for external reviewers or limited team members)

PR #1161, opened 2025-05-20
feat(access): add role-based access control and permissions

kelvinkipruto:
## Description

Fixes # (issue)

## Type of change

## Checklist:


kilemensi:
👍🏽 @kelvinkipruto ... will take a deeper look but a few things that I'm not too sure:

1. Lets start with about 3 roles only i.e. those that can do something in the dashboard. Looking at [WordPress](https://wordpress.com/support/invite-people/user-roles/#list-of-user-roles), [Medium](https://help.medium.com/hc/en-us/articles/115004681647-Manage-publication-editors-and-writers), [Ghost](https://ghost.org/docs/staff/#roles--permissions) (and [others](https://apostrophecms.medium.com/user-roles-for-small-teams-5f03f2f0be3f)), these should probably be just `Administrator` (or `Admin` but since the dashboard is also admin, may be we should stick to `Administrator`), `Editor` and `Author` (or `Writer` but I think more CMSes use `Author`). For now, we don't need anything higher than `Administrator` (e.g. `Owner`) or lower than `Author` (e.g. `Contributor`, `Viewer`, `Guest`, etc.). We can always add when needed but removing may be tricky (migrations, etc.)
2. Users should have exactly **one** role e.g. Everything `Author` can do, `Editor` can do so there is no need to give a user both `Editor` and `Author` roles i.e. roles are hierarchical.

kelvinkipruto:
Thanks, @kilemensi. @RalphOwah, are the roles listed in the [issue](https://github.com/CodeForAfrica/ui/issues/1136#issue-3045902641) a "hard requirement", or can we go with Clemence's [suggestion](https://github.com/CodeForAfrica/ui/pull/1161#issuecomment-2894575841) above?

RalphOwah:
> Thanks, @kilemensi. @RalphOwah, are the roles listed in the [issue](https://github.com/CodeForAfrica/ui/issues/1136#issue-3045902641) a "hard requirement", or can we go with Clemence's [suggestion](https://github.com/CodeForAfrica/ui/pull/1161#issuecomment-2894575841) above?

...................................................................................................................................................................................................
Alright @KiprotichKoros, We can work with @kilemensi suggested approach of 3 Roles : 

1. Administrator
Full access to settings, collections, and user management

2. Editor
Can manage, create, edit site blocks and publish content

3. Author
Can create and edit content, but cannot publish.

kilemensi:
> 2. Editor
> Can manage, create, edit site blocks and publish content
>
And add/remove Authors

> 3. Author
> Can create and edit content, but cannot publish.
>

Authors **should** be able to create, edit and publish **_own_** posts and tags. If we ever need users that can't publish, we'll add `Contributor` role but that's down the line if/when we need it.

The only thing we need to decide @kelvinkipruto / @RalphOwah is can Authors or Editors edit `pages` and other sections of the site e.g. site settings, etc., or limit them to just posts/articles.

kelvinkipruto:
> The only thing we need to decide @kelvinkipruto / @RalphOwah is can Authors or Editors edit `pages` and other sections of the site e.g. site settings, etc., or limit them to just posts/articles.

@kilemensi, we can follow the WordPress model, where editors can create/edit/delete pages while authors are restricted to just posts. Neither should be able to change the site settings.

kilemensi:
>@kilemensi, we can follow the WordPress model, where editors can create/edit/delete pages while authors are restricted to just posts. Neither should be able to change the site settings.
>

Works for me @kelvinkipruto

kilemensi:
Is this ready for a review? I still see `roles: ("admin" | "editor" | "reviewer" | "viewer")` in code.

kelvinkipruto:
@kilemensi Now ready.

kilemensi:
Just tested this @kelvinkipruto :

1. If there are existing users, all users become non-admin and hence can't add new users or even edit.
2. If I delete the database and start from scratch, I can't create a user at all.
Screenshot 2025-05-22 at 14 55 31

Could it be related to the fact only admins can create users and the first user is obviously not an admin (yet)?

apps/trustlab/src/app/preview/route.ts
```diff
@@ -50,14 +51,22 @@ export async function GET(
...
     draft.disable();
     return new Response("You are not allowed to preview this page", {
       status: 403,
     });
   }
 
-  //TODO: Add further checks to ensure the user is allowed to preview this page
+  const user = authResult.user || authResult;
+
+  const isEditor = canManageContent(user);
```
kilemensi, line 63:
```suggestion
  const user = authResult.user || authResult;
  const isEditor = canManageContent(user);
```
Why empty lines between related code?

apps/trustlab/src/app/preview/route.ts
```diff
@@ -50,14 +51,22 @@ export async function GET(
...
+  if (!isEditor) {
```
kilemensi, line 64:
```suggestion
  if (canManageContent(user)) {
```

Authors can manage their own content; doesn't need to be an editor.

apps/trustlab/src/payload/access/abilities.js
```diff
@@ -0,0 +1,67 @@
...
+            equals: user.id,
+          },
+        },
+        {
+          "createdBy.role": {
+            in: [ROLE_EDITOR, ROLE_AUTHOR],
+          },
+        },
+      ],
+    };
```
kilemensi, line 24: Editors should be able to edit any content on the site. Whether the content was created by themselves, authors, other editors or admins, it shouldn't matter.

apps/trustlab/src/payload/access/abilities.js
```diff
@@ -0,0 +1,67 @@
+import { checkRole } from "./checkRole";
+import { ROLE_ADMIN, ROLE_AUTHOR, ROLE_DEFAULT, ROLE_EDITOR } from "./roles";
+
+export const canManageContent = (user) => {
+  if (!user) return false;
+  const isAdmin = checkRole([ROLE_ADMIN], user);
+  if (isAdmin) return true;
```
kilemensi, line 7:
```suggestion
  if (isAdmin) {
    return true;
  }
```

Write full `if`'s with `{ }`. Aren't `prettier`/`eslint` configured for this?

apps/trustlab/src/payload/access/abilities.js
```diff
@@ -0,0 +1,67 @@
...
+  checkRole([ROLE_ADMIN, ROLE_EDITOR], user);
+
+export const canManageSiteSettings = (user) => checkRole([ROLE_ADMIN], user);
+
+export const canManageUsers = (user) => {
+  if (!user) return false;
+
+  if (user.role === ROLE_ADMIN) return true;
+
+  if (user.role === ROLE_DEFAULT)
```
kilemensi, line 43: If manage = edit/remove, then:

1. Admins can `create` and `manage` all users
2. Editors can `manage` their own account and `create`/`manage` authors (an editor can not create or delete other editors or admins)
3. Authors can only `manage` their own account.

---

One thing to think about is what happens to the content when a user is removed.

apps/trustlab/src/payload/access/anyone.js
```diff
@@ -0,0 +1,3 @@
+/* eslint-disable import/prefer-default-export */
+
+export const anyone = () => true;
```
kilemensi, line 3: An alternative to disabling this is:
```suggestion
export const anyone = () => true;

export default undefined;
```

apps/trustlab/src/payload/access/roles.js
```diff
@@ -0,0 +1,9 @@
+export const ROLE_ADMIN = "administrator";
+export const ROLE_EDITOR = "editor";
+export const ROLE_AUTHOR = "author";
+export const ROLE_DEFAULT = ROLE_AUTHOR;
```
kilemensi, line 4: I'm not sure we need this... why not make sure a role is specified wherever/whenever it's needed?

apps/trustlab/src/payload/collections/Media.js
```diff
@@ -1,13 +1,21 @@
...
+
 const filename = fileURLToPath(import.meta.url);
 const dirname = path.dirname(filename);
 
 const Media = {
   slug: "media",
   access: {
-    read: () => true,
+    read: anyone,
+    create: anyone,
```
kilemensi, line 16: If `anyone` means a **logged in** user with any role then cool; otherwise we need to make sure only staff i.e. dashboard users can create `Media`, `Posts`, etc.

apps/trustlab/src/payload/collections/Users/hooks/protectRoleField.js
```diff
@@ -0,0 +1,33 @@
+/* eslint-disable import/prefer-default-export */
+
+import {
+  ROLE_ADMIN,
+  ROLE_AUTHOR,
+  ROLE_DEFAULT,
+  ROLE_EDITOR,
+} from "@/trustlab/payload/access/roles";
+
+export const protectRoleField = ({ req, data, originalDoc, operation }) => {
```
kilemensi, line 10: Can you add a comment what does this actually do?

kilemensi (commented):
👍🏽

kelvinkipruto:
> If I delete the database and start from scratch, I can't create a user at all.

The best way forward is to ensure the first user is always an admin using the `beforeChange` or `beforeValidate` hook.

> If there are existing users, all users become non-admin and hence can't add new users or even edit.

I think the best way forward for this is to start with a new db for this pr.

kilemensi:
> @kilemensi, just a quick one, can a user demote themself?
>

Weird but yes @kelvinkipruto, I think that should be allowed i.e. an admin can change their own role to an editor, and an editor can change their own role to author.

apps/trustlab/src/app/preview/route.ts
kilemensi, line 1: Still so many empty lines/repeated code blocks ... not sure why. Instead of:
```js
  let authResult;

  try {
    authResult = await payload.auth({
      req: req as unknown as PayloadRequest,
      headers: req.headers,
    });
  } catch (error) {
    payload.logger.error(
      { err: error },
      "Error verifying token for live preview",
    );
    return new Response("You are not allowed to preview this page", {
      status: 403,
    });
  }

  const draft = await draftMode();

  if (!authResult) {
    draft.disable();
    return new Response("You are not allowed to preview this page", {
      status: 403,
    });
  }

  const user = authResult.user || authResult;

  if (canManageContent(user)) {
    draft.enable();
  } else {
    draft.disable();
    return new Response("You are not allowed to preview this page", {
      status: 403,
    });
  }

  redirect(path);
```

why not just:
```js
  let authResult;
  try {
    authResult = await payload.auth({
      req: req as unknown as PayloadRequest,
      headers: req.headers,
    });
  } catch (error) {
    payload.logger.error(
      { err: error },
      "Error verifying token for live preview",
    );
  }
  if (!(authResult && canManageContent(authResult?.user))) {
    return new Response("You are not allowed to preview this page", {
      status: 403,
    });
  }

  const draft = await draftMode();
  draft.enable()
  redirect(path);
``

File: apps/trustlab/src/payload/access/abilities.js [hunk 10]
```diff
@@ -0,0 +1,60 @@
...
+    },
+  };
+};
+
+export const canManagePages = (user) =>
+  checkRole([ROLE_ADMIN, ROLE_EDITOR], user);
+
+export const canManageSiteSettings = (user) => checkRole([ROLE_ADMIN], user);
+
+// TODO: what happens on delete? cascade or not?
```
kilemensi, line 26:
```suggestion
// TODO(kelvinkipruto): what happens on delete? cascade or not?
```

Add a name so we know who to talk about about this or who is going to do it.

apps/trustlab/src/payload/access/abilities.js
```diff
@@ -0,0 +1,60 @@
...
+export const canManageContent = (user) => {
+  // Admin and editors can manage any content
+  if (checkRole([ROLE_ADMIN, ROLE_EDITOR], user)) {
+    return true;
+  }
+
+  // Everyone else can only manage their own content
+  return {
+    createdBy: {
+      equals: user.id,
```
koechkevin, line 13: @kelvinkipruto this throws an error when run on an initially empty database
image

apps/trustlab/src/payload/access/abilities.js
```diff
@@ -0,0 +1,60 @@
...
+      ],
+    };
+  }
+
+  return {
+    id: {
+      equals: user.id,
+    },
+  };
+};
```
kilemensi, line 60: Again, simpler, tighter code is better I believe.
```suggestion
  if (!user) {
    return false;
  }

  // Admins can manage all users
  if (user.role === ROLE_ADMIN) {
    return true;
  }

  // All other users can manage their own accounts
  orQuery = [{
    id: {
      equals: user.id,
    }
  }];
  // Editors can manage authors (in addition to their accounts)
  if (user.role === ROLE_EDITOR) {
    orQuery.push({ role: { equals: ROLE_AUTHOR } });
  }
  return {
    or: orQuery,
  };
};
```

apps/trustlab/src/payload/access/checkRole.js
```diff
@@ -0,0 +1,8 @@
+export const checkRole = (roles, user) => {
+  if (!user) {
+    return false;
+  }
+  return roles.includes(user.role);
```
kilemensi, line 5: Or just:
```suggestion
  return roles.includes(user?.role);
```

`roles` doesn't include `undefined`.

apps/trustlab/src/payload/collections/Media.js
```diff
@@ -1,13 +1,22 @@
...
+
 const filename = fileURLToPath(import.meta.url);
 const dirname = path.dirname(filename);
 
 const Media = {
   slug: "media",
   access: {
-    read: () => true,
+    read: anyone,
+    create: ({ req: { user } }) => loggedIn(user),
```
kilemensi, line 17:
```suggestion
    create: ({ req: { user } }) => canManageContent(user),
```

koechkevin (commented):
This looks good except for the issue highlighted on initial database

apps/trustlab/src/payload/collections/Pages.js
```diff
@@ -1,23 +1,20 @@
...
-        },
-      };
-    },
-    create: () => true,
-    update: () => true,
+    read: anyone,
+    create: ({ req: { user } }) => canManagePages(user),
+    update: ({ req: { user } }) => canManagePages(user),
+    delete: ({ req: { user } }) => canManagePages(user),
+    admin: ({ req: { user } }) => canManagePages(user),
```
kilemensi, line 14: Can you add a comment why we might need this here? It's been a while since I read the docs but my old understanding was that `admin` access can **_only_** be used on collection that supports authentication. In our case that would be `Users` collection and not `Pages`.

apps/trustlab/src/payload/collections/Pages.js
```diff
@@ -29,6 +26,14 @@ const Pages = {
...
       return `${process.env.NEXT_PUBLIC_APP_URL}/preview?${encodedParams.toString()}`;
     },
+    components: {
+      edit: {
+        PublishButton:
+          "@/trustlab/payload/components/PagesPublishButton#PagesPublishButton",
+        PreviewButton:
+          "@/trustlab/payload/components/PagesPreviewButton#PagesPreviewButton",
+      },
+    },
```
kilemensi, line 36: Do we need this? If we've already set access to edit pages to editor/admins only, I assume authors won't be able to see pages altogether or they'll see them in readonly mode.

apps/trustlab/src/payload/collections/Post.js
kilemensi, line 1: File is `Post` but from `slug`, collection name seem to be `Posts`. Which is which?

apps/trustlab/src/payload/access/abilities.js
```diff
@@ -0,0 +1,60 @@
...
+  if (checkRole([ROLE_ADMIN, ROLE_EDITOR], user)) {
+    return true;
+  }
+
+  // Everyone else can only manage their own content
+  return {
+    createdBy: {
+      equals: user.id,
+    },
+  };
```
kilemensi, line 18: If we're following [WordPress](https://wordpress.com/support/invite-people/user-roles/#list-of-user-roles) or [Ghost](https://ghost.org/docs/staff/#roles--permissions) model, authors should only be able to manage posts. If I'm reading this correctly, authors will be able to manage pages if/when they create one.

apps/trustlab/src/payload/collections/Post.js
```diff
@@ -0,0 +1,48 @@
...
+    slug(),
+    createdBy({
+      overrides: {
+        hidden: false,
+      },
+    }),
+  ],
+  versions: {
+    drafts: true,
+  },
```
kilemensi, line 45:
```suggestion
  versions: {
    drafts: {
      autosave: true,
    },
  },
```

We don't want people to lose their work... should be on `Pages` as well.

apps/trustlab/src/payload/collections/Users/hooks/protectRoleField.js
```diff
@@ -0,0 +1,64 @@
...
+  // ensure first user is admin
+  if (operation === "create") {
+    const { docs } = await req.payload.find({
+      collection: "users",
+      limit: 1,
+    });
+    if (!docs.length) {
+      return ROLE_ADMIN;
+    }
+  }
```
kilemensi, line 23: I think another option would be to customise the `create-first-user` [route](https://payloadcms.com/docs/admin/overview#admin-level-routes) to use a custom view that doesn't ask for role but set it to admin.

apps/trustlab/src/payload/access/abilities.js
kelvinkipruto: @kilemensi We have a different function for pages [`canManagePages`](https://github.com/CodeForAfrica/ui/blob/e27bdf993f740acc66b8767bc4308d15f1657391/apps/trustlab/src/payload/access/abilities.js#L21-L22)

apps/trustlab/src/payload/collections/Pages.js
kelvinkipruto: You are very right on this @kilemensi. It works only for auth collections

apps/trustlab/src/payload/collections/Users/hooks/protectRoleField.js
kelvinkipruto: @kilemensi This does not allow us to customise the `create-first-user` page, just the route. So if we use `/first-user`, the first user registration route would be `/admin/first-user`. Looking through their Discord, I came across this [message](https://discord.com/channels/967097582721572934/1084195567946702908/1087797496849367091) on customising the first user page:

> ... currently we don't have a component override in place for the create new user view. This would be a great addition if you want to open a feature request on our GitHub OR if you fancy building it, we would also happily accept a PR.
You can still use custom css on this page and possibly use a workaround.

apps/trustlab/src/payload/access/abilities.js
kilemensi: ... cool, content == posts + media.

apps/trustlab/src/payload/collections/Users/hooks/protectRoleField.js
```diff
@@ -0,0 +1,52 @@
...
+  switch (user.role) {
+    case ROLE_ADMIN:
+      return data?.role;
+    case ROLE_EDITOR:
+      // Editor can only set role to EDITOR or AUTHOR
+      if ([ROLE_EDITOR, ROLE_AUTHOR].includes(data?.role)) {
+        return data?.role;
+      }
+      // Ensure editors cannot create admin users
+      return operation === "create" ? ROLE_AUTHOR : originalDoc?.role;
```
kilemensi, line 44: Add a issue to track this.. I assume the system will silently convert say ROLE_EDITOR to ROLE_AUTHOR without warning editors that they can't create editors.

kilemensi (approved):
👍🏽 ... haven't run it locally but code looks good.

apps/trustlab/src/payload/collections/Users/hooks/protectRoleField.js
kilemensi: That is messed up then... but since we're probably the ones that will create the first user, no need to stress too much on this.

PR #1162, opened 2025-05-20
Fix `@/trustlab` build

kilemensi:
## Description

More fixes as we deploy Docker image:

  - [x] Set `PROJECT_ROOT` to [help](https://nextjs.org/docs/pages/api-reference/config/next-config-js/output#caveats) build process since we're using `output: 'standalone'`
  - [x] Include `require-in-the-middle` as a dependency as it was missing at runtime causing application start error (Seems like PayloadCMS is using it so we may need to track it in their repo as to why it's not included when building using `output: 'standalone'`)
  - [x] Configure nodeMailerAdaptorArgs first before initialising email. This should stop mock email from being created even when SMTP_ settings are provided.
  - [x] Disable telemetry
  - [x] Fix pnpm start command since next start doesn't work when using `output: 'standalone'`.
  
## Type of change

- [x] Chore

### This PR (`make trustlab`)

Screenshot 2025-05-20 at 16 25 14

### `require-in-the-middle` bug

Screenshot 2025-05-20 at 13 59 45

### SMTP bug

Screenshot 2025-05-20 at 13 00 24

## Checklist:


kilemensi:
> @kilemensi Looks good. I have, however, noticed that it will still provide a mock email address if we do not provide `SMTP_HOST` and `SMTP_PASS`...
>

Yes, that's the design @kelvinkipruto: if `SMTP_HOST` and `SMTP_PASS` are not provided, use mock email service. The problem was even when `SMTP_HOST` and `SMTP_PASS` were provided, **it still create**d mock email service.

koechkevin (approved):

PR #1172, opened 2025-05-28
Fix TrustLab Docker build failure due to `--link` and `chown` conflict

kelvinkipruto:
## Description

The TrustLab docker build currently [fails](https://github.com/CodeForAfrica/ui/actions/runs/15270064450/job/42947211368) with an error:

```bash
ERROR: failed to solve: invalid user index: -1
```
This is caused by using `--link` together with `--chown`, which is a known issue with Docker BuildKit. Other apps are unaffected by this, but TrustLab hits this specific conflict.

Related issues: 
 - https://github.com/docker/buildx/issues/1526
 - https://github.com/moby/buildkit/issues/2987
 - https://github.com/docker/buildx/issues/1408
 - https://github.com/docker/buildx/issues/1855

~~Two proposed [solutions](https://github.com/docker/buildx/issues/1526#issuecomment-1396768545) have been discussed in the community:~~
~~- Introduce an intermediate build stage~~
~~- Use numeric UID/GID instead of user/group names~~

~~This PR implements the second approach by using the numeric UID/GID (1001:1002) instead of `nextjs:nodejs`, which resolves the build failure~~

This PR solves the issue by removing the `--link` that we do not need.

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -969,7 +969,7 @@ COPY --from=trustlab-builder --chown=nextjs:nodejs /workspace/node_modules ./nod
 
 # Next.js
 # Public assets
-COPY --from=trustlab-builder --chown=nextjs:nodejs /workspace/apps/trustlab/public ./apps/trustlab/public
+COPY --from=trustlab-builder --chown=nextjs:nodejs /workspace/apps/trustlab/publi[c] ./apps/trustlab/public
 
 # Automatically leverage output traces to reduce image size
 # https://nextjs.org/docs/advanced-features/output-file-tracing


kilemensi:
Thanks @kelvinkipruto 

Still not sure about all of this but:

(Option 1) Fix in this PR: Why are we changing GID value? Both should be `1001`
(Option 2) What is `--link` actually giving us? Why not just remove it if other apps are not using it and we're fine with it?

kelvinkipruto:
@kilemensi Thanks. The simplest option is removing the `--link` that we do not need.

Dockerfile
```diff
@@ -965,11 +965,11 @@ FROM base-runner AS trustlab-runner
 
 # PNPM
 # symlink some dependencies
-COPY --from=trustlab-builder --link --chown=nextjs:nodejs /workspace/node_modules ./node_modules
+COPY --from=trustlab-builder --chown=nextjs:nodejs /workspace/node_modules ./node_modules
 
 # Next.js
 # Public assets
-COPY --from=trustlab-builder --chown=nextjs:nodejs /workspace/apps/trustlab/publi[c] ./apps/trustlab/public
+COPY --from=trustlab-builder --chown=nextjs:nodejs /workspace/apps/trustlab/public ./apps/trustlab/public
```
kilemensi, line 972:
```suggestion
COPY --from=trustlab-builder --chown=nextjs:nodejs /workspace/apps/trustlab/publi[c] ./apps/trustlab/public
```

The reason for `publi[c]` is so that if the folder public doesn't exist, the `COPY` instruction should not throw an error.

kilemensi (approved):

PR #938, opened 2024-10-08
Switch to linux/arm64 images

VinneyJ:
## Description

Switch to building `linux/arm64` images (except where we don't have control of the infrastructure e.g. charterAFRICA).

## Type of change

- [x] Chore

## Screenshots

N/A

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation

PR #1175, opened 2025-05-29
@/trustlab payload improvements

kilemensi:
## Description

 - [x] Enable disable APIs in `/admin` based on `site.environment` (using `SENTRY_ENVIRONMENT`).
 - [x] Simplify localized: true in Group fields i.e. no need to enable in nested fields since enabling it at group level ensures everything inside the group is localised.
 - [x] Ensure only loggedIn users can read settings as it may contain secrets that shouldn't be publicly available. Since we're using local APIs to build the site, this shouldn't have any impact on generating pages.

## Type of change

- [x] Chore

## Screenshots

N/A

## Checklist:

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -32,7 +32,7 @@ jobs:
 
       # Add support for more platforms with QEMU (optional)
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v
+        uses: docker/setup-qemu-action@v3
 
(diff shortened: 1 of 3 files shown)


kilemensi:
FYI @CodeForAfrica/tech, the linting issues are project-level issues and not related to the changes in this PR. They're fixed as part of [this](https://github.com/CodeForAfrica/ui/pull/938) PR. May be we should start by reviewing and merging that one.

PR #1180, opened 2025-06-04
TrustLab: Reusable Banner

kelvinkipruto:
## Description
This PR adds a reusable banner.
It can be used for the Page Header and other sections, like a simple showcase section

Fixes # (issue)

## Type of change

## Screenshots
image
image
image
image

## Checklist:


apps/trustlab/src/payload/blocks/Banner.js
```diff
@@ -0,0 +1,50 @@
...
+      name: "image",
+      relationTo: "media",
+    },
+    {
+      type: "row",
+      fields: [
+        {
+          type: "text",
+          name: "backgroundColor",
+          defaultValue: "#02041C",
```
kilemensi, line 27: Lets use/create a proper color field. Quick googlin' shows:

- https://www.npmjs.com/package/@nouance/payload-better-fields-plugin
- https://www.npmjs.com/package/@modernaut/payload-color-swatch-field

Lets review the packages (popularity, history, etc.) and pick one. Otherwise, we can also [implement](https://oleksii-s.dev/blog/how-to-build-color-input-in-payload-cms-3) a custom one.

kilemensi (commented):
👍🏽 

--

To avoid confusing content editors in the CMS, we should create the actual expected blocks with proper defaults that reuses this Banner block code. Front-end can remain just one Banner component that knows how to render `page-header`, `about-project`, `what-we-do`, etc.

apps/trustlab/src/payload/blocks/PageHeader.js
```diff
@@ -0,0 +1,61 @@
...
+    image({
+      overrides: {
+        name: "image",
+      },
+    }),
+    {
+      type: "checkbox",
+      name: "isPageHeader",
+      defaultValue: true,
+      label: "Is Page Header",
```
kilemensi, line 27: 1. Why do we need `label` if it's `hidden`?
2. Since even `readOnly` fields can be edited via APIs, why not add `isPageHeader` using `afterRead` hook?

apps/trustlab/src/payload/blocks/WhatWeDo.js
```diff
@@ -0,0 +1,63 @@
...
+    {
+      type: "checkbox",
+      name: "isPageHeader",
+      defaultValue: false,
+      label: "Is Page Header",
+      admin: {
+        hidden: true,
+        readOnly: true,
+      },
+    },
```
kilemensi, line 34: Why do we need this if it's supposed to always be `false`?

apps/trustlab/src/components/Banner/Banner.js
```diff
@@ -0,0 +1,85 @@
...
+
+  return (
+    <Section
+      sx={{
+        maxWidth: { md: "100%", xs: "100%" },
+        backgroundColor,
+      }}
+      fixed={false}
+    >
+      <Box
```
kilemensi, line 24: Aren't we using this in reverse? Section is supposed to handle maxWidth uniformly for the whole site. If we want something that is always 100% wide, why not:

1. `<Box><Section...`
2. Just `Section` with `before` [pseudo-element](https://developer.mozilla.org/en-US/docs/Web/CSS/Pseudo-elements) to sent background at 100% width, etc.

kilemensi (commented):
👍🏽

kilemensi (approved):
👍🏽 

---

Aim for structured code in addition to working code.

PR #1186, opened 2025-06-10
Fix isPageHeader bug

kelvinkipruto:
## Description

[`Blocks`](https://payloadcms.com/docs/fields/blocks#block-configs) do not support hooks. Only `Fields`, `Globals` and `Collections` do.
This PR fixes a bug where we used `afterRead` to update the `isPageHeader` field by using a [virtual field](https://payloadcms.com/docs/fields/checkbox#:~:text=a%20JSON%20schema-,virtual,-Provide%20true%20to) to ensure that the value is not saved in the db, but is computed dynamically.

## Type of change

## Checklist:

### DIFF
--- a/apps/trustlab/src/payload/blocks/PageHeader.js
+++ b/apps/trustlab/src/payload/blocks/PageHeader.js
@@ -4,9 +4,17 @@ const PageHeader = BannerBlock(
   "page-header",
   "/images/cms/blocks/page-header.png",
   {
-    hooks: {
-      afterRead: [({ doc }) => ({ ...doc, isPageHeader: true })],
-    },
+    fields: [
+      {
+        name: "isPageHeader",
+        type: "checkbox",
+        defaultValue: true,
+        virtual: true,
+        admin: {
+          hidden: true,
+        },
+      },
+    ],
   },
 );
 


apps/trustlab/src/payload/blocks/PageHeader.js
```diff
@@ -4,9 +4,17 @@ const PageHeader = BannerBlock(
...
-      afterRead: [({ doc }) => ({ ...doc, isPageHeader: true })],
-    },
+    fields: [
+      {
+        name: "isPageHeader",
+        type: "checkbox",
+        defaultValue: true,
+        virtual: true,
+        admin: {
+          hidden: true,
```
koechkevin, line 14: Not exactly sure why we are introducing a field but have it hidden again.

kilemensi: 🤔 ... yeah, it's bummer if we can't have hooks on blocks (I assume we can still add a field hook on the collection using these blocks that will mean repeating the hooks on each collection).

Stepping back for a bit, our implementation of `BannerBlock` mandates that each custom banner passes a `slug` i.e. `blockType`. If we already have this on the frontend, that is we can tell these banners apart, do we really need anything extra?

kelvinkipruto: @kilemensi That's the simplest option. Let me use the `slug` to differentiate the different banners.

koechkevin (approved):

PR #1190, opened 2025-06-13
Fix failing app start

kelvinkipruto:
## Description

This PR fixes an issue encountered when starting any app after merging #1183.

The main issue I encountered appeared to be an issue with multiple versions of the same package being installed.

This was resolved by removing `nodelinker: hoisted` from `pnpm-workspace.yaml`.

From the `nodelinker` [docs](https://pnpm.io/settings#nodelinker), it can be used only for a number of use cases, which I do not think satisfy our usage.

By default, [`autoInstallPeers`](https://pnpm.io/settings#autoinstallpeers) is `true`; hence, there is no need to include it.

## Type of change

## Checklist:


kilemensi:
Happy to see you quick on this but lets discuss this. `nodeLinker` change is so that `pnpm` behaves similarly to `npm` (and `yarn`, etc.). One of the issue with default `pnpm` is it uses simlinks which can cause errors. For example, `eslint` [recommends](https://eslint.org/docs/latest/use/getting-started#manual-set-up) this change.

Haven't tested Docker builds but the real issue is likely on `Dockerfile` where we're [copying](https://github.com/CodeForAfrica/ui/blob/b9402ec6273e4a4552202fe00a37169dd576839b/Dockerfile#L957-L958) workspace `node_modules`. By changing `nodeLinker` may be we no longer need this line

kilemensi:
Finally getting back to this @kelvinkipruto, is it still an issue?

koechkevin (approved):

PR #1248, opened 2025-08-11
Trustlab/Fixes

koechkevin:
## Description

Enhance Footer, header and selection banner

Fixes # (issue)

## Type of change

## Checklist:

### DIFF
--- a/apps/trustlab/src/components/NavBar/NavBar.js
+++ b/apps/trustlab/src/components/NavBar/NavBar.js
@@ -4,7 +4,6 @@ import {
   DesktopNavBar,
 } from "@commons-ui/core";
 import { Link, NextImageButton } from "@commons-ui/next";
-import { Box } from "@mui/material";
 import React from "react";
 
 import CloseIcon from "@/trustlab/assets/icons/Type=x, Size=24, Color=CurrentColor.svg";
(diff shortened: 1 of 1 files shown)


kilemensi (commented):
👍🏽 ... QQ from me:

1. I see we're changing code to remove the social media link(s). Can't they just be optional in the CMS such that if not set, we show nothing (this could be the way it is already)? I don't think every UI tweak should be code change.
2. In the footer, there is the heading "Quick Links". Where does it come from (I'm talking about CMS)? And does it show just primary navigation or just secondary navigation or both?

koechkevin:
> 2. In the footer, there is the heading "Quick Links". Where does it come from (I'm talking about CMS)? And does it show just primary navigation or just secondary navigation or both?

This section only shows primary links  and the label comes from CMS, General Tab

apps/trustlab/src/lib/data/common/index.js
```diff
@@ -29,14 +29,16 @@ function getFooter(settings) {
     secondaryLogo,
     secondaryNavigation,
     title,
+    quickLinksLabel,
```
kilemensi, line 32: Do we need this? I thought `label`/`title` will be `primaryNavigation.title || secondaryNavigation.title`.

apps/trustlab/src/payload/globals/tabs/GeneralTab.js
```diff
@@ -18,6 +18,13 @@ const GeneralTab = {
           required: true,
           localized: true,
         }),
+        {
+          name: "quickLinksTitle",
+          type: "text",
+          required: true,
+          localized: true,
+          defaultValue: "Quick Links",
+        },
```
kilemensi, line 27: We shouldn't have this, no?

apps/trustlab/src/components/FooterLinks/FooterLinks.js
```diff
@@ -4,39 +4,73 @@ import React from "react";
...
+            display: "flex",
           }}
           component="div"
+          variant="h1"
         >
-          {title}
-        </Box>
-      </Typography>
-      <FooterNavList menus={primaryMenus} />
+          <Box
```
kilemensi, line 29: Do we really need `Box` inside `Typography` or am I reading the code wrong?

kilemensi (commented):
👍🏽

kilemensi:
Most likely there won't be a connect action on NavBar but **IF** there is one, does this arrangment looks 💯 to you @koechkevin ? Screenshot 2025-08-12 at 10 50 01

PR #1459, opened 2026-04-16
feat: add dynamic sitemap generation and API endpoint

kelvinkipruto:
## Description

Implement sitemap generation functionality with a new API endpoint at `/api/v1/sitemap`. The sitemap includes all published pages (excluding error pages).

- Add sitemap.xml rewrite rule in Next.js config
- Create sitemap API endpoint handler
- Implement sitemap generation utilities including URL normalization and deduplication

Fixes # (issue)

## Type of change

## Checklist:


kelvinkipruto:
@claude @codex Review

kelvinkipruto:
@claude Review

kelvinkipruto:
@claude Review

kelvinkipruto:
@claude  Review

apps/trustlab/src/lib/data/common/sitemap.js
```diff
@@ -0,0 +1,134 @@
...
+    return null;
+  }
+
+  return {
+    url,
+    lastModified: getLastModified(doc),
+  };
+}
+
+function getPagePathname(doc) {
```
kilemensi: We don't need these utilities anymore I believe. Every collection item that ends up as a page on the frontend, should have a `pathname` attribute created in the `afterRead` hook.

apps/trustlab/src/lib/data/common/sitemap.js
```diff
@@ -0,0 +1,134 @@
...
+  if (!doc) {
+    return null;
+  }
+
+  const pathname = formatPagePath("pages", doc);
+  return normalizePathname(pathname);
+}
+
+async function getPagesEntries(api) {
+  const { docs } = await api.getCollection("pages", {
```
kilemensi, line 60: Only `published` items should be returned.

apps/trustlab/src/lib/data/common/sitemap.js
```diff
@@ -0,0 +1,134 @@
...
+      },
+    },
+  });
+
+  return docs
+    .map((doc) => toSitemapEntry(doc, getPagePathname(doc)))
+    .filter(Boolean);
+}
+
+function dedupeEntries(entries) {
```
kilemensi: `slug`s are unique so why would we have duplicates? If we do end up with duplicates, that should be an error that's logged in Sentry.

apps/trustlab/src/lib/data/common/sitemap.js
```diff
@@ -0,0 +1,134 @@
...
+    .map(({ url, lastModified }) => {
+      const lastModifiedNode = lastModified
+        ? `\n    <lastmod>${lastModified}</lastmod>`
+        : "";
+
+      return `  <url>\n    <loc>${escapeXml(url)}</loc>${lastModifiedNode}\n  </url>`;
+    })
+    .join("\n");
+
+  return `<?xml version="1.0" encoding="UTF-8"?>\n<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">\n${xmlEntries}\n</urlset>\n`;
```
kilemensi, line 117: 1. Do we need to escape XML? e.g. https://nextjs.org/learn/seo/xml-sitemaps doesn't (and do we need this to be in `pages/api` instead of just `pages`?
2. If it's going to be too big, we may need to look into streaming instead of creating the whole file in memory.

kilemensi (commented):
👍🏽 

---

1. We have more pages than just `pages` collection e.g. most of `opportunities` collection items also become pages.
2. If you've addressed all comments, please hide them/mark them as resolved. Otherwise, it's hard to review not knowing if you should comment on something that already has a comment or not.

kelvinkipruto:
@claude @codex Review

kilemensi:
Haven't looked at the new code yet @kelvinkipruto but my previous question still remains, why do we need we need to use `pages/api` and a `redirect` instead of just in `pages`?

> ~Do we need to escape XML? e.g. https://nextjs.org/learn/seo/xml-sitemaps doesn't~ (and do we need this to be in pages/api instead of just pages?
>

apps/trustlab/src/lib/data/common/sitemap.js
```diff
@@ -0,0 +1,156 @@
...
+  if (pathname === "/") {
+    return pathname;
+  }
+
+  const trimmed = pathname.trim();
+  if (!trimmed) {
+    return null;
+  }
+
+  const withLeadingSlash = trimmed.startsWith("/") ? trimmed : `/${trimmed}`;
```
kilemensi: 1. I don't think we need path normalisation since all collection items that appear as pages on the front-end, already have `pathname` attribute, and it should be normalised already (please check the code that generates them and lets fix that if broken)
2. We seem to be removing trailing `/` from `site.url` but here we're prepending `/` to pathname. It seems counter-productive.

apps/trustlab/src/lib/data/common/sitemap.js
```diff
@@ -0,0 +1,156 @@
...
+  if (!rawDate) {
+    return null;
+  }
+
+  const parsedDate = new Date(rawDate);
+  if (Number.isNaN(parsedDate.getTime())) {
+    return null;
+  }
+
+  return parsedDate.toISOString();
```
kilemensi: Lets also remove these unnecessary empty lines; if it's from the agent, please review and fix it before creating PRs.

apps/trustlab/src/lib/data/common/sitemap.js
```diff
@@ -0,0 +1,156 @@
...
+    .filter(Boolean);
+}
+
+async function getSitemapEntries(api) {
+  const [pages, opportunities] = await Promise.all([
+    getPagesEntries(api),
+    getOpportunitiesEntries(api),
+  ]);
+
+  return [...pages, ...opportunities].sort((left, right) =>
```
kilemensi: 1. [I don't think we need](https://www.sitemaps.org/faq.html#faq_url_position) this expensive sorting.
2. If we _want_ to sort, may be sorting by last modified, is more meaningful to see what has changed on the site.

apps/trustlab/src/pages/sitemap.xml.page.js
```diff
@@ -0,0 +1,30 @@
...
+      props: {},
+    };
+  } catch (error) {
+    Sentry.captureException(error);
+    res.statusCode = 500;
+    res.end();
+    return {
+      props: {},
+    };
+  }
```
kilemensi: Unless there is a reason, lets try reducing code duplication

```suggestion
  } catch (error) {
    Sentry.captureException(error);
    res.statusCode = 500;
  }
  res.end();
  // needed by getServerSideProps
  return {
    props: {},
  };
```

apps/trustlab/src/pages/sitemap.xml.page.js
```diff
@@ -0,0 +1,30 @@
...
+}
+
+export default function SitemapXml() {
+  return null;
```
kilemensi, line 26:
```suggestion
  // All work is done and returned in getServerSideProps.
  return null;
```

kilemensi (commented):
👍🏽

apps/trustlab/src/lib/data/common/sitemap.js
```diff
@@ -0,0 +1,156 @@
...
+    left.url.localeCompare(right.url),
```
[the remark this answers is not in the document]
kilemensi: Yes, `slug`s _could_ be duplicate but `pathname`, which is what we're interested in, will always be different i.e it's the URL the items appear on the frontend and two items can't have the same URL regardless of collection.

koechkevin (approved):

apps/trustlab/src/lib/data/common/sitemap.js
```diff
@@ -0,0 +1,127 @@
...
+    .map((doc) => {
+      if (!doc?.pathname) {
+        warnOnMissingPathname("opportunities", doc);
+        return null;
+      }
+
+      return toSitemapEntry(doc, doc.pathname);
+    })
+    .filter(Boolean);
+}
```
kilemensi, line 98: Seems like the mapping of collection items to `sitemapEntry` is repeated for `pages` and `opportunities`

```suggestion
}

function docsToSitemapEntries(docs, collection) {
  return docs
    .map((doc) => {
      if (!doc?.pathname) {
        warnOnMissingPathname(collection, doc);
        return null;
      }
      return toSitemapEntry(doc, doc.pathname);
    })
    .filter(Boolean);
}

async function getOpportunitiesEntries(api) {
  const { docs } = await api.getCollection("opportunities", {
    pagination: false,
    select: {
      pathname: true,
      slug: true,
      type: true,
      updatedAt: true,
      createdAt: true,
      date: true,
    },
  });
  return docsToSitemapEntries(docs, "opportunities");
}
```

We can then reuse `docsToSitemapEntries` in `getPagesEntries` as well.

kilemensi (approved):

kilemensi:
Can we also get this updated and merged @kelvinkipruto so that Claude/Codex can do another full site review?

PR #1460, opened 2026-04-17
feat: add default sort option to OpportunityList and ReportsList components; update related logic in data fetching and API handlers

koechkevin:
## Description
This PR fixes several correctness bugs in the OpportunityList and ReportsList filter/search/sort feature, and adds editor-configurable default sort to both blocks.                                                    
   
  ---                                                                                                                                                                                                                    
  Changes                                                                                                                                                                                                                
                                                                                                                                                                                                                         
  getOpportunities.js — clean up Payload where clause                                                                                                                                                                    
                                                                                                                                                                                                                         
  Previously, the function mixed flat field conditions (where.type, where.title, etc.) with an explicit where.and array when date filters were also active. Payload CMS's behaviour when both are present simultaneously 
  is implementation-defined and may break across versions. All conditions are now collected into a single andConditions array and composed as { and: andConditions }, consistent with the reports API.                   
                                                                                                                                                                                                                         
  opportunities.page.js — sort allowlist + safe default                                                                                                                                                                  
   
  Added an ALLOWED_SORT allowlist mirroring the reports API. Raw ?sort= values that don't match are dropped; the default is "-date". This prevents arbitrary strings (e.g. ?sort=__proto__) from reaching the ORM.       
                                                                                                                                                                                                                       
  OpportunityList.js (Payload block) — clearFiltersLabel condition widened                                                                                                                                               
                                                                                                                                                                                                                       
  clearFiltersLabel was previously gated only on hasFilters. If a block enabled hasSearch or hasSortBy without hasFilters, the clear button rendered in the UI but editors had no way to localise its label. The         
  condition now checks hasFilters || hasSearch || hasSortBy, matching ReportsList.                                                                                                                                     
                                                                                                                                                                                                                         
  OpportunityList.js (component) — router.isReady guard                                                                                                                                                                  
   
  The useEffect that restores filter/search/sort state from the URL query had an empty [] deps array. On a hard page load with ?search=foo&sort=-date, Next.js fires the effect before router.query is populated, so the 
  restored state was silently discarded. The effect now guards with if (!router.isReady) return and uses [router.isReady] as its dependency — the same pattern already used in ReportsList.                            
                                                                                                                                                                                                                         
  ReportsList.js (component) — skip condition accounts for search/sort

  useReports was called with skip=!hasPagination. When hasPagination=false and hasSearch=true, typing into the search box updated params.search correctly but the SWR fetch was never fired — search silently did        
  nothing. The skip flag is now !hasFilters && !hasPagination && !hasSearch && !hasSortBy, matching OpportunityList.
                                                                                                                                                                                                                         
  Default sort — both blocks and components                                                                                                                                                                              
   
  Both OpportunityList and ReportsList Payload blocks gain a defaultSort text field (visible when hasSortBy is enabled). Editors enter a value like -date or -createdAt that must match one of the configured sort option
   values. This sort is:
  - Used for the SSR initial fetch in getOpportunityListBlock and reportsList blockify functions                                                                                                                         
  - Seeded into the component's initial params state                                                                                                                                                                     
  - Restored by handleClearAll so clearing filters snaps back to the configured default rather than the API's implicit fallback
  - null-defaulted in blockify to avoid undefined serialisation errors from getStaticProps                                                                                                                               
                                                                                                                                                                                                                         
  ---                                                                                                                                                                                                                    
  Test plan                                                                                                                                                                                                              
                                                                                                                                                                                                                         
  - Hard-reload a page with ?search=foo&sort=-date — results should match the URL state immediately without an extra interaction
  - With hasSearch=true, hasPagination=false: type in the search box — results should update                                                                                                                             
  - With hasSortBy=true, hasPagination=false: change sort — results should update                                                                                                                                        
  - Apply date + location + search filters simultaneously — confirm all conditions are honoured (no silent drops)                                                                                                        
  - Set defaultSort=-date in CMS, reload page — initial order should reflect -date                                                                                                                                       
  - Click "Clear all" — list should revert to defaultSort order, not API implicit default                                                                                                                                
  - Send ?sort=__proto__ to /api/v1/opportunities — should 200 with default sort, not forward the raw string                                                                                                             
  - clearFiltersLabel field appears in the CMS admin when only hasSearch or hasSortBy is enabled (without hasFilters) 

Fixes # (issue)

## Type of change

## Screenshots

## Checklist:


koechkevin:
@claude review

koechkevin:
@claude review

koechkevin:
@claude rereview

apps/trustlab/src/payload/blocks/ReportsList.js
```diff
@@ -135,6 +135,16 @@ const ReportsList = {
         },
       ],
     },
+    {
+      name: "defaultSort",
+      type: "text",
+      label: { en: "Default Sort (e.g. -date, -createdAt)" },
```
kilemensi: 1. How do content editors know about `-date`, `-createdAt`, etc? Shouldn't this be a select/dropdown option?
2. It should have a `defaultValue` of `-date`

apps/trustlab/src/payload/blocks/OpportunitiesList.js
```diff
@@ -150,6 +150,16 @@ const OpportunityList = {
         },
       ],
     },
+    {
```
kilemensi, line 163: Similar comments to the ReportsList but one question, why is this block named `OpportunityList` (singular) but for reports, we have `ReportsList` (plural)?

apps/trustlab/src/pages/api/v1/opportunities.page.js
```diff
@@ -16,20 +27,10 @@ export default async function handler(req, res) {
       location,
       opportunity,
       search,
-      sort,
+      sort = "-date",
```
kilemensi: This should match what is default in the CMS.

apps/trustlab/src/pages/api/v1/reports/index.page.js
```diff
@@ -7,7 +7,7 @@ export default async function handler(req, res) {
   if (method === "GET") {
     const {
       page,
-      sort,
+      sort = "-date",
```
kilemensi: Same here, this should default to what is in the CMS... and why do we have `v1/reports/index.page.js` but `v1/opportunities.page.js`? I know they both work but why this little inconsistencies everywhere?

apps/trustlab/src/lib/data/getOpportunities.js
```diff
@@ -25,74 +25,51 @@ async function getOpportunities(api, options = {}) {
...
-      });
-      andConditions.push({
-        date: {
-          less_than_equal: endOfMonth,
-        },
-      });
-    }
+  if (year) {
+    const startOfYear = new Date(year, 0, 1).toISOString();
+    const endOfYear = new Date(year, 11, 31, 23, 59, 59, 999).toISOString();
```
kilemensi: Why not just use next year/month for end of year/month _and_ then use `less_than` instead of `less_than_equal` in query?

apps/trustlab/src/lib/data/blockify/reportsList.js
```diff
@@ -1,10 +1,15 @@
 import { getReports } from "@/trustlab/utils/reports";
 
 async function reportsList(block, api) {
-  const { blockType: slug, reportsType, hasPagination } = block;
+  const {
+    blockType: slug,
+    reportsType,
+    hasPagination,
+    defaultSort = null,
```
kilemensi: Add `defaultValue` in the CMS so that we are sure there will always be a value.

kilemensi:
The issue has been outstanding for a while now @koechkevin so if you've reviewed and tested, lets deploy it.

kelvinkipruto (approved):

PR #1464, opened 2026-04-23
fix(eslint): stabilise eslint flat config and restore monorepo lint workflows

kilemensi:
## Description

This PR repairs and stabilises ESLint across the monorepo by finishing the flat-config migration, removing legacy lint plumbing, and restoring consistent workspace-aware lint execution.

### What changed

  - replaced the old compatibility-based shared ESLint setup with native flat config in `eslint-config-commons-ui`
  - split shared config into base, Next.js, and TypeScript variants
  - restored Turbo env-var linting via `eslint-config-turbo`
  - restored import/order consistently across both JS and TS configs
  - extracted shared rule fragments instead of duplicating rule definitions across presets
  - removed `eslint-plugin-prettier` from linting so ESLint is only responsible for linting
  - replaced next lint and unstable config lookup flags with direct eslint CLI usage
  - standardised on `lint:check` as the read-only lint task across root, Turbo, apps, and packages
  - added `apps/eslint.config.js` so JS apps under `apps/` inherit the shared Next.js config
  - centralised `lint-staged` behaviour in the shared ESLint package and switched workspace wrappers to a factory pattern using `__dirname`
  - removed legacy `/* eslint-env */` comments now covered by flat-config globals
  - removed stale lint dependencies that were no longer used, including Airbnb/Prettier integration leftovers
  - removed dead `eslint-import-resolver-jsconfig` config/dependency wiring
  - removed `TIMING=1` from normal lint scripts as [`--stats`](https://eslint.org/docs/latest/extend/stats) command can be used to give more detail if/when needed e.g. `pnpm lint:check -- --stats` at root dir.
  - added a conservative `turbo.json` env declaration set so Turbo cache correctness checks work again

### Why

The previous setup had become brittle in a few ways:

  - the ESLint 9 flat-config migration was incomplete
  - linting and formatting concerns were coupled together
  - next lint and old resolver assumptions were hiding real monorepo issues
  - staged-file linting was not reliably running in the owning workspace
  - Turbo env-var enforcement had been dropped during the migration

This PR restores a clean lint baseline without changing formatter strategy yet.

### Verification

  - `pnpm` run `lint:check` passes at repo root
  - representative JS and TS workspaces resolve the expected shared config
  - Next.js rules are active again for apps inheriting `apps/eslint.config.js`
  - `lint-staged` now runs ESLint from the correct workspace context

### Notes

This PR intentionally leaves warning-only debt in place, especially:

  - unused `eslint-disable` comments now surfaced by `reportUnusedDisableDirectives: "warn"`
  - a small number of existing non-blocking app warnings such as `@next/next/no-img-element`

Formatter migration is intentionally out of scope. That will follow in a separate PR focused on replacing `prettier` with `oxfmt`.

## Type of change

- [x] Chore

## Screenshots

N/A

## Checklist:


kilemensi:
@codex review

kilemensi:
@claude review

kilemensi:
@claude 

### What I agree with:

  - `AGENTS.md` stale description: yes, fixed.
  - `typescript.js` disabling both unused-vars rules: real gap, but I would treat it as a separate policy change, not a last-minute tweak in this stabilisation PR.
  - `react-hooks/exhaustive-deps`: same answer. Useful signal, but changing it to `warn` now will add repo-wide warning churn and distract from the goal of restoring a stable lint baseline.

### What I don't want to change in this PR:

  - `next.rootDir`: `"./"` in `packages/eslint-config-commons-ui/next.js:1:` low risk in the current execution model. All important paths (`turbo`, `lint-staged`, workspace scripts) run ESLint from the owning workspace. Per-app `__dirname` overrides would be more robust, but that is not worth widening this PR.
  - `lintstaged.js` quoting via `JSON.stringify`: acceptable for this repo’s actual contributor/CI path today; no one is using Windows.
  - explicit `import/resolver` settings: I would avoid reintroducing them casually. The old resolver setup was part of what made lint brittle in the first place.

  On the Turbo env feedback:

  - `NEXT_PUBLIC_GOOGLE_ANALYTICS_ID` is used but at app level and not global. It has been moved into respective apps' `turbo.json` instead of polluting global envs.
  
@claude review again with the above context.

kilemensi:
`NEXT_PUBLIC_GOOGLE_ANALYTICS_ID` moved to app-level turbo.json. @claude review and confirm.

koechkevin (approved):

kelvinkipruto (approved):

PR #1473, opened 2026-05-13
fix(trustlab): enforce role-based Payload content access

kilemensi:
## Description

`@/trustlab`'s Payload access layer had two categories of issues.

**Authorization gaps** (P1):
- Reports, Playbooks, and Toolkits used `Boolean(user)`, so any authenticated session, regardless of role, could create, update, or delete these collections.
- Partners, Donors, Opportunities, and Organisations had no explicit `create`/`update`/`delete` access at all, relying on Payload's implicit behaviour.
- `SiteSettings` was world-readable: `loggedIn` was defined as `(user) => Boolean(user)` but wired as a Payload access adapter, so Payload called it as `loggedIn({ req })`, which was always truthy.

**Structural issues** that made auditing difficult:
- User-level predicates and Payload access adapters shared the same function signatures, so call sites had to inline `({ req: { user } }) => helper(user)` everywhere.
- `Boolean(user)` trusts any truthy `req.user` value; a token with an arbitrary payload would pass.

### Changes

**Access layer refactor (`src/payload/access/`)**

Introduces a clear three-tier naming convention:

| Tier | Returns | Used by |
|---|---|---|
| `isXxx(user)` | `boolean` | custom route/middleware code |
| `canXxx(user)` | `boolean \| Payload where-filter` | internal policy logic |
| `hasXxxAccess({ req })` | Payload access adapter | collection/global `access:` config |

- `hasValidRole(user)` added to `roles.js`: validates `user.role` against the known role constants. Replaces `Boolean(user)` everywhere.
- `loggedIn.js` removed. `isLoggedIn`, `isEditor`, `isAuthor`, `isAdmin`, `canAuthor`, `canManageUser` are the new user-level predicates.
- `hasLoggedInAccess`, `hasEditorAccess`, `hasAuthorAccess`, `hasAdminAccess`, `hasManageUserAccess`, `hasCreateUserAccess` are the new Payload adapters, each a thin wrapper over the corresponding predicate.

**Collections and globals**

| Surface | Before | After |
|---|---|---|
| Reports, Playbooks, Toolkits (write) | `Boolean(user)` | `hasEditorAccess` |
| Pages (write) | `canManagePages(user)` inline | `hasEditorAccess` |
| Opportunities, Organisations, Partners, Donors (write) | missing (implicit) | `hasEditorAccess` |
| Posts, Media (write) | `canManageContent(user)` inline | `hasAuthorAccess` |
| Users collection | inline wrappers | named adapters; role-field update tightened to `hasCreateUserAccess` |
| SiteSettings read | `loggedIn` (broken — always `true`) | `hasLoggedInAccess` |
| SiteSettings update | `canManageSiteSettings` inline | `hasAdminAccess` |

Posts and Media intentionally remain on `hasAuthorAccess` to preserve the ownership-aware author workflow.

**Preview route**

`preview/route.ts` now uses `isAuthor(user)`, the correct boolean predicate, instead of the old `canManageContent(user)` whose filter-object return value was only coincidentally truthy in a boolean context.

**Tests (`src/payload/access/index.test.js`)**

Full policy matrix at the helper layer: `isAuthor`, `canAuthor`, `hasAuthorAccess`, `hasEditorAccess`, `hasAdminAccess`, `hasLoggedInAccess`, `hasCreateUserAccess`, `hasManageUserAccess`, and `anyone`. Cases cover admin/editor/author/anonymous and invalid-role shapes (`role: "unknown"`, bare integer, `undefined`). The `canAuthor` suite includes the id-guard edge case (author with no `id` returns `false` rather than a filter with `equals: undefined`).

## Type of change

## Screenshots

N/A

## Checklist:

### DIFF
--- a/apps/trustlab/src/payload/access/abilities.js
+++ b/apps/trustlab/src/payload/access/abilities.js
@@ -41,7 +41,7 @@ export const hasAdminAccess = ({ req } = {}) => isAdmin(req?.user);
 
 // TODO(@kelvinkipruto): what happens on delete? cascade or not?
 export const canManageUser = (user) => {
-  if (!user) {
+  if (!isLoggedIn(user)) {
     return false;
   }
   // Admins can manage all users
@@ -49,6 +49,9 @@ export const canManageUser = (user) => {
     return true;
   }
   // All other users can manage their own accounts
+  if (!user.id) {
+    return false;
+  }
   const orQuery = [
     {
       id: {
--- a/apps/trustlab/src/payload/access/abilities.js
+++ b/apps/trustlab/src/payload/access/abilities.js
@@ -1,16 +1,31 @@
 import { checkRole } from "./checkRole";
-import { ROLE_ADMIN, ROLE_AUTHOR, ROLE_EDITOR } from "./roles";
+import { hasValidRole, ROLE_ADMIN, ROLE_AUTHOR, ROLE_EDITOR } from "./roles";
 
-export const canManageContent = (user) => {
-  if (!user) {
+export const isLoggedIn = (user) => hasValidRole(user);
+
+export const hasLoggedInAccess = ({ req } = {}) => isLoggedIn(req?.user);
+
+// Admins and editors can manage any content
+export const isEditor = (user) => checkRole([ROLE_ADMIN, ROLE_EDITOR], user);
+
+export const hasEditorAccess = ({ req } = {}) => isEditor(req?.user);
+
+export const isAuthor = (user) => {
+  // Any valid CMS user has at least author-level capability.
+  return isLoggedIn(user);
+};
+
+export const canAuthor = (user) => {
+  if (!isLoggedIn(user)) {
     return false;
   }
-  // Admin and editors can manage any content
-  if (checkRole([ROLE_ADMIN, ROLE_EDITOR], user)) {
+  if (isEditor(user)) {
     return true;
   }
-
-  // Everyone else can only manage their own content
+  if (!user.id) {
+    return false;
+  }
+  // Authors can manage own content only
   return {
     createdBy: {
       equals: user.id,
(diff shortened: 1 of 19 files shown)


kilemensi:
@codex review

apps/trustlab/src/payload/globals/index.js
```diff
@@ -14,8 +16,8 @@ const SiteSettings = {
   access: {
     // Since we're using Local APIs, we should still be able to pull data server-side
     // See: note in https://payloadcms.com/docs/local-api/overview#transactions
-    read: loggedIn,
-    update: canManageSiteSettings,
+    read: hasLoggedInAccess,
```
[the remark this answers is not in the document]
kilemensi: Fixed. @codex review

kilemensi:
@claude review

kilemensi:
[P2] canManageUser skips the hasValidRole guard and [P3] Missing edge-case tests for hasManageUserAccess without id are now fixed.

On isAuthor naming vs semantics, this matches isEditor which also returns true for editors and admins.

@claude review

apps/trustlab/src/payload/access/abilities.js
```diff
@@ -1,39 +1,57 @@
...
+
+// Admins and editors can manage any content
+export const isEditor = (user) => checkRole([ROLE_ADMIN, ROLE_EDITOR], user);
+
+export const hasEditorAccess = ({ req } = {}) => isEditor(req?.user);
+
+export const isAuthor = (user) => {
+  // Any valid CMS user has at least author-level capability.
+  return isLoggedIn(user);
+};
```
koechkevin, line 16: I think this `isAuthor` name is misleading/redundant if its functionality is identical to `isLoggedIn`

kilemensi:
Going twice @CodeForAfrica/tech ...

koechkevin:
> Going twice @CodeForAfrica/tech ...

Currently testing @kilemensi

koechkevin (approved):
Looks good to me

apps/trustlab/src/payload/access/abilities.js
kilemensi: My understanding of the code is they're different @koechkevin :

  * `isLoggedIn`: is used to check if the current user is anonymous or authenticated i.e. handles the first A in [AAA](https://en.wikipedia.org/wiki/Authentication,_authorization,_and_accounting).
  * `isAuthor` (and `isEditor`, `isAdmin`, etc.): is used to check authorisation level that should be given to the authenticated user i.e. the second A in AAA. The fact this two functions mean the same right now is just a coincidence; if we introduce say `contributor` role to the app, this would no longer be the case.

PR #1486, opened 2026-06-12
Guard against missing date and location in opportunities

koechkevin:
## Summary

Defensive null-handling in the opportunities data path to avoid rendering errors when a record is missing a date or location.

## Changes

- **`payload/utils/formatDate.js`** — return `null` when no `date` is passed, instead of formatting `undefined` (which produces an "Invalid Date").
- **`lib/data/pagify/opportunities.js`** — fall back to `null` for a missing `doc.location` (`doc?.location ?? null`).

## screenshots
image

## Why

Opportunity overview content could be built from records without a `date`/`location`, surfacing an invalid date or throwing on an undefined field. These guards make both paths resilient.

## Testing

- `eslint` clean on changed files.

### DIFF
--- a/apps/trustlab/src/lib/data/pagify/opportunities.js
+++ b/apps/trustlab/src/lib/data/pagify/opportunities.js
@@ -32,7 +32,7 @@ function pagifyOpportunities(collection) {
         return {
           ...block,
           date,
-          location: doc?.location ?? null,
+          location: doc.location || null,
           title: block.title || "Overview",
         };
       }
--- a/apps/trustlab/src/lib/data/pagify/opportunities.js
+++ b/apps/trustlab/src/lib/data/pagify/opportunities.js
@@ -32,7 +32,7 @@ function pagifyOpportunities(collection) {
         return {
           ...block,
           date,
-          location: doc.location,
+          location: doc?.location ?? null,
           title: block.title || "Overview",
         };
       }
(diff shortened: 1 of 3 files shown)


koechkevin:
@claude review

apps/trustlab/src/lib/data/pagify/opportunities.js
```diff
@@ -32,7 +32,7 @@ function pagifyOpportunities(collection) {
         return {
           ...block,
           date,
-          location: doc.location,
+          location: doc?.location ?? null,
```
kilemensi: We've established `doc` already exists above so we shouldn't need `?` in `doc?.location`. May be:

```suggestion
          location: doc.location || null,
```
?

apps/trustlab/src/payload/utils/formatDate.js
```diff
@@ -1,6 +1,8 @@
 const formatDate = (date, options = {}) => {
   const { locale = "en", includeTime, ...restOptions } = options;
-
+  if (!date) {
+    return null;
+  }
```
kilemensi: I'd put the `if(!date?.length)` above the `const` declaration as we don't need to declare anything if we don't have a date (and if date is string, we need to guard against empty string as well).

apps/trustlab/src/payload/utils/formatDate.test.js
```diff
@@ -0,0 +1,31 @@
...
+      includeTime: true,
+      timeZone: "UTC",
+    });
+
+    expect(formatted).toContain("January 15, 2024");
+    expect(formatted).toMatch(/1:45/);
+  });
+
+  it("returns null when formatting throws", () => {
+    expect(formatDate("2024-01-15", { locale: "!!bad locale" })).toBeNull();
```
kilemensi: Include `undefined`, `null`, bad date and bad locale here.

kilemensi (approved):

kelvinkipruto (approved):

